PR #6, opened 2019-04-24
Add signing to worker, spark binaries and nuget packages

safern:
Here is a example build:

https://dnceng.visualstudio.com/internal/_build/results?buildId=167505

For easier reviewing disable spaces in the diff setting

cc: @danmosemsft @eerhardt @rapoth @imback82

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,135 +3,258 @@
 trigger:
 - master
 
-pool:
-  vmImage: 'VS2017-Win2016'
-
 variables:
   solution: '**/*.sln'
   buildConfiguration: 'Release'
+  _SignType: real
+  _TeamName: DotNetSpark
 
-steps:
-- task: NuGetToolInstaller@0
-  inputs:
-    versionSpec: '4.9.2'
+  # Azure DevOps variables are transformed into environment variables, with these variables we
+  # avoid the first time experience and telemetry to speed up the build.
+  DOTNET_CLI_TELEMETRY_OPTOUT: 1
+  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
 
-- task: DotNetCoreCLI@2
-  displayName: '.NET restore'
-  inputs:
-# Use a custom restore command because the built in restore command uses a temp nuget.config
-# which overwrites the MSBuild restore properties
-    command: 'custom'
-    custom: 'restore'
-    projects: '$(solution)'
+jobs:
+- job: Build
+  displayName: Build and Test Sources
+  pool: Hosted VS2017
 
-- task: DotNetCoreCLI@2
-  displayName: '.NET build'
-  inputs:
-    command: build
-    projects: '$(solution)'
-    arguments: '--configuration $(buildConfiguration)'
+  steps:
+  - task: DotNetCoreCLI@2
+    displayName: '.NET build'
+    inputs:
+      command: build
+      projects: '$(solution)'
+      arguments: '--configuration $(buildConfiguration)'
 
-- task: BatchScript@1
-  displayName: Publish Microsoft.Spark.Worker
-  inputs:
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -3,135 +3,258 @@
...
+        pathtoPublish: '$(Build.ArtifactStagingDirectory)'
+        artifactName:  Microsoft.Spark.Binaries
+
+- ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+  - job: Package
+    dependsOn:
+      - Build
+    displayName: Build packages
+    pool:
+      name: NetCoreInternal-Int-Pool
```
eerhardt: Why not just run the whole build on `NetCoreInternal-Int-Pool` and collapse this into a single job?  That way you don't have to publish the artifacts and then pull them down. It could be simpler to do it all in 1 job.

azure-pipelines.yml
```diff
@@ -3,135 +3,258 @@
...
+        msbuildArguments: /t:SignBinaries
+                          /p:SignWorkerBinaries=true
+                          /p:SignAssetsDir=$(Build.ArtifactStagingDirectory)\Microsoft.Spark.Binaries\Microsoft.Spark.Worker\
+                          /p:SignType=$(_SignType)
+        msbuildVersion: 15.0
+
+    - task: NuGetCommand@2
+      inputs:
+        command: pack
+        packagesToPack: '$(Build.SourcesDirectory)\src\csharp\Microsoft.Spark.nuspec'
```
eerhardt: Why aren't we using `dotnet pack` here?

azure-pipelines.yml
eerhardt: Ah, I see we weren't before. So maybe this is a question for @imback82 and @rapoth

---
In reply to: [278199899](https://github.com/dotnet/spark/pull/6#discussion_r278199899) [](ancestors = 278199899)

azure-pipelines.yml
eerhardt: Ah, I see we weren't before. So maybe this is a question for @imback82 and @rapoth

---
In reply to: [278200318](https://github.com/dotnet/spark/pull/6#discussion_r278200318) [](ancestors = 278200318)

azure-pipelines.yml
safern: That was the initial approach but the machines in `NetCoreInternal-Int-Pool` don't have Maven in it and we need Maven for the scala sources.

azure-pipelines.yml
imback82: `dotnet pack` should be also fine. What will be the benefit over `nuget` in this case? Just curious.

safern: It is cross plat, you don't need to bootstrap nuget.exe when you already have dotnet, you can package a csproj, so no need to create nuspec.

eerhardt: I've opened https://github.com/dotnet/spark/issues/36 for this.

azure-pipelines.yml
eerhardt: Just an FYI - that might be a good comment to have in the .yml file.

Also - can we fix `NetCoreInternal-Int-Pool` to have it?

safern: I talked to @chcosta yesterday and he recommended this approach, but yes I agree, maybe the `NetCoreInternal-Int-Pool` should have it. cc: @mattgal is this possible? 

I'll add a comment in the meantime

MattGal: @safern the general guidance is to bootstrap anything in builds which can be done in a non-stateful-to-machine way;  I'll do some reading on Maven and if it makes sense we can add it to these machines (though it will take a bit to roll out)

PR #44, opened 2019-04-24
TPCH Benchmark fixes

adamsitnik:
cc @rapoth @imback82


suhsteve:
@adamsitnik thanks for the fix.  Reran the q13 in the cluster and the perfs  were very close to the original runs.

suhsteve:
LGTM

adamsitnik:
>  Reran the q13 in the cluster and the perfs were very close to the original runs.

@suhsteve thanks for letting me know! I will continue my perf investigation

PR #46, opened 2019-04-24
Make F# example app more idiomatic

slang25:
I've tidied up the F# example app discovery code so that it is more in keeping with F# idioms.

### DIFF
--- a/examples/Microsoft.Spark.FSharp.Examples/Program.fs
+++ b/examples/Microsoft.Spark.FSharp.Examples/Program.fs
@@ -5,23 +5,18 @@
 module Microsoft.Spark.Examples.Main
 
 open System
-open System.Collections.Generic
-open System.Linq
 open System.Reflection
-open System.Runtime.InteropServices
 
-let printUsage (examples : IEnumerable<string>) =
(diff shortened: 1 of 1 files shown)


examples/Microsoft.Spark.FSharp.Examples/Program.fs
```diff
@@ -5,23 +5,18 @@
 module Microsoft.Spark.Examples.Main
 
 open System
-open System.Collections.Generic
-open System.Linq
 open System.Reflection
-open System.Runtime.InteropServices
 
-let printUsage (examples : IEnumerable<string>) =
+let printUsage (examples : string seq) =
```
cartermp: Not a dealbreaker by any means, but `seq<string>` is preferred here as per the style guide: https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#use-prefix-syntax-for-generics-foot-in-preference-to-postfix-syntax-t-foo

slang25: Thanks @cartermp, I didn't know that

Issue #13, opened 2019-04-24
[Documentation]: Get Started Experience on Ubuntu


rapoth:
At the moment, the Get Started section in README.md talks about how users can get started on a Windows machines using .NET Core. We need a similar set of instructions for Ubuntu.

**Success Criteria**
User has clear instructions on:
- Setting up all the pre-requisites
- Creating a new .NET Core project
- Obtain the Nuget package on Ubuntu
- Compiling the .NET app on Ubuntu
- Running the app using spark-submit

lqdev:
# Getting Started with Spark.NET on Ubuntu

These instructions help get you started with ML.NET on Ubuntu 18.04

## Download and Install Prerequisites

- [Java 1.8](https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html)
- [Spark 2.4.x](https://spark.apache.org/downloads.html)
- Maven
- [.NET Core SDK 2.1](https://dotnet.microsoft.com/download/dotnet-core/2.1)

### Install Java

Navigate to the following [link](https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) and download `jdk-8u211-linux-x64.tar.gz`.

Then, extract the contents of the `tar.gz` folder:

```bash
tar -xvzf jdk-8u211-linux-x64.tar.gz
```

Add Java 1.8 to the list of Java versions on your system:

```bash
sudo update-alternatives --install "/usr/bin/java" "java" "/home/$USER/jdk1.8.0_211/bin/java" 1500
sudo update-alternatives --install "/usr/bin/javac" "javac" "/home/$USER/jdk1.8.0_211/bin/javac" 1500
sudo update-alternatives --install "/usr/bin/javaws" "javaws" "/home/$USER/jdk1.8.0_211/bin/javaws" 1500
```

Follow the prompts and select 1.8 as the version:

```
sudo update-alternatives
```

### Install Maven

Enter the following commands into the terminal: 

```bash
mkdir -p ~/bin/maven
cd ~/bin/maven
wget https://www-us.apache.org/dist/maven/maven-3/3.6.0/binaries/apache-maven-3.6.0-bin.tar.gz
tar -xvzf apache-maven-3.6.0-bin.tar.gz
ln -s apache-maven-3.6.0 current
```

### Install Spark

Download the latest version of Spark:

```bash
[...]

PR #50, opened 2019-04-25
getting started ubuntu instructions

lqdev:
PR for #13


docs/getting-started/getting-started-ubuntu-spark-net.md
```diff
@@ -0,0 +1,187 @@
...
+These instructions help get you started with Spark on Ubuntu 18.04
+
+## Download and Install Prerequisites
+
+- [Java 1.8](https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html)
+- [Spark 2.4.x](https://spark.apache.org/downloads.html)
+- Maven
+- [.NET Core SDK 2.1](https://dotnet.microsoft.com/download/dotnet-core/2.1)
+
+### Install Java
```
suhsteve: Instead of reiterating the install steps for java, spark, netcore and maven, we can instead reference docs/building/ubuntu-instructions.md#pre-requisites and update those instructions accordingly if necessary.   This will help if we ever need to update these instructions in the future.

lqdev: Updated. I added a link to the [home page](https://github.com/dotnet/spark#get-started) for the pre-requisites. One thing to note though is maven is not required to create an app. This is just in the event a user wants to build from source. Otherwise, they can just take the [Microsoft.Spark.Worker release package](https://github.com/dotnet/spark/releases) which seems like the easiest way to go about it.

lqdev: Another approach would be to have another document showing how to set up pre-requisites (without maven and cloning the repo to build from source).

docs/getting-started/getting-started-ubuntu-spark-net.md
```diff
@@ -0,0 +1,100 @@
...
+        {
+            var spark = SparkSession.Builder().GetOrCreate();
+            var df = spark.Read().Json("people.json");
+            df.Show();
+        }
+    }
+}
+```
+
+Add the following content to your `HelloSpark.csproj` file inside the `Project` tags:
```
imback82: Can we follow the same flow as the Windows example in the main README.md?

Basically, you will publish first, go to the published directory, create the input file, run (meaning no need to update csproj file) and the `spark-submit` command will be:
```bash
spark-submit \
--class org.apache.spark.deploy.DotnetRunner \
--master local \
./microsoft-spark-2.4.x-0.1.0.jar \
./HelloSpark
```

docs/getting-started/getting-started-ubuntu-spark-net.md
suhsteve: Let's try to keep the Windows and Linux instruction layout/structure as similar as possible.  Please see https://github.com/dotnet/spark#get-started  .  After your changes are merged, we can pull out the windows instructions into it's own file (similar to yours) and provide links from the main readme to both.

docs/getting-started/getting-started-ubuntu-spark-net.md
lqdev: Updated

suhsteve:
LGTM

lqdev:
Thanks @suhsteve and @imback82 . Is this ready for merge?

PR #54, opened 2019-04-25
Migrated nuspec to csproj and improved packaging

### LINKED ISSUE #36: Use `dotnet pack` against a csproj to create nuget package
We are currently using `nuget.exe` and a `.nuspec` file to create the Spark NuGet package. We should instead use `dotnet pack` and MSBuild to do this.

The advantages are that we can:

1. Maintain the dependencies in a single place (the .csproj).
2. Use MSBuild properties to control things like the version number. This allows for pre-release daily builds to have different version numbers.
3. It works cross-platform
4. We can take advantage of new features, like .snupkg packages for symbols, source link, and the `<repository  />` tag in the .nuspec which tells users which git repo and commit this package was built from.

@imback82 @safern @rapoth

### LINKED ISSUE #37: Microsoft.Spark nuget package shouldn't work on all TFMs
The Microsoft.Spark nuget package has a `build\Microsoft.Spark.targets` file. Since this file isn't contained in a TFM folder, nuget will think this package works on any TFM, even TFMs that aren't compatible with `netstandard2.0`.

See https://github.com/dotnet/machinelearning/pull/370 for the same issue that came up in ML.NET.

We should move that `.targets` file into a `build\netstandard2.0` folder.

### LINKED ISSUE #38: Microsoft.Spark nuget package shouldn't have symbols files in it
We are currently packaging the `.pdb` symbols file into the NuGet package. But we should be using the new `.snupkg` format instead.

See https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg

The reason we don't want symbols in the NuGet package is because it bloats every application that uses this package. Symbols are only be used when debugging, so publishing an application shouldn't need them.

### LINKED ISSUE #40: Add SourceLink support
We aren't currently embedding SourceLink information into the PDB - that allows a debugger to automatically find and pull the source files.

https://github.com/dotnet/sourcelink has info

```
C:\dotnet>dotnet tool install --global sourcelink
Since you just installed the .NET Core SDK, you will need to reopen the Command Prompt window before running the tool you installed.
You can invoke the tool using the following command: sourcelink
Tool 'sourcelink' (version '3.0.0') was successfully installed.
…
```
```
C:\dotnet>sourcelink test C:\pkg\lib\netstandard2.0\Microsoft.Spark.pdb | find /i "failed"
sourcelink test failed
```

(Note you need 2.1 installed currently to run the tool: https://github.com/ctaggart/SourceLink/issues/380)

alefranz:
This fixes #36, fixes #37, fixes #38, fixes #40

- Migrated from nuspec to csproj
- Fixed targets location
- Fixed package dependencies
- Fixed package to be recognised as netstandard2.0
- Moved debug symbols to symbols package
- Added Sourcelink
- Fixed package dependencies

Before / After
image

Symbols package
image

Package size is actually bigger now
image

For the Azure Pipelines I'm flying blind, but hopefully should work correctly and not break the signature.
I don't think the symbol package needs to be signed.

- [X] There's a descriptive title that will make sense to other developers some time from now. 
- [X] There's associated issues. All PR's should have issue(s) associated - unless a trivial self-evident change such as fixing a typo. You can use the format `Fixes #nnnn` in your description to cause GitHub to automatically close the issue(s) when your PR is merged.
- [X] Your change description explains what the change does, why you chose your approach, and anything else that reviewers should know.


.gitignore
```diff
@@ -180,6 +180,8 @@ PublishScripts/
 
 # NuGet Packages
 *.nupkg
+# NuGet Symbol Packages
+*.snupkg
```
alefranz: Updated as in https://github.com/github/gitignore/pull/3041

azure-pipelines.yml
```diff
@@ -229,11 +229,13 @@ jobs:
...
-
-    - task: NuGetCommand@2
+ 
+    - task: DotNetCoreCLI@2
+      displayName: 'Create NuGet packages'
       inputs:
         command: pack
-        packagesToPack: '$(Build.SourcesDirectory)\src\csharp\Microsoft.Spark.nuspec'
+        projects: '$(solution)'
+        arguments: '--no-build --configuration $(buildConfiguration) --output $(Build.ArtifactStagingDirectory)'
```
alefranz: I'm not really sure if --no-build is really needed even if the binaries had been signed

azure-pipelines.yml
```diff
@@ -229,11 +229,13 @@ jobs:
...
         msbuildVersion: 15.0
-
-    - task: NuGetCommand@2
+ 
+    - task: DotNetCoreCLI@2
+      displayName: 'Create NuGet packages'
       inputs:
         command: pack
-        packagesToPack: '$(Build.SourcesDirectory)\src\csharp\Microsoft.Spark.nuspec'
+        projects: '$(solution)'
```
imback82: How about using `$(Build.SourcesDirectory)\src\csharp\Microsoft.Spark.csproj` and remove `IsPackable` in other csproj files? @eerhardt @safern what's the recommended way here?

alefranz: I prefer this approach as it doesn't require changes every time you add new projects as you usually want them as packages (test project have `IsPackable` set to false by default and for the examples I should move the flag to the build.props).
That's why I also used wildcards in the csproj to include the `jar`s
However, currently the sign steps is not currently using wildcards for the project name so that would need to be changed as well.

Let me know what do you think. Happy to switch to just pack that project directly if that is your preference or you don't see new packaged projects added in the near future.

examples/Microsoft.Spark.FSharp.Examples/Microsoft.Spark.FSharp.Examples.fsproj
```diff
@@ -1,11 +1,12 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFrameworks>net461;netcoreapp2.1</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.1</TargetFrameworks>
     <RootNamespace>Microsoft.Spark.Examples</RootNamespace>
     <AssemblyName>Microsoft.Spark.FSharp.Examples</AssemblyName>
+      <IsPackable>False</IsPackable>
```
eerhardt: (nit) the whitespace here is off

azure-pipelines.yml
eerhardt: I think it is good policy to specify `--no-build` here. We strictly do not want to build/compile the binary while doing `pack`. Better to be explicit IMO.

---
In reply to: [278708264](https://github.com/dotnet/spark/pull/54#discussion_r278708264) [](ancestors = 278708264)

eerhardt:
>     - task: NuGetToolInstaller@0

I believe you can remove this `NuGetToolInstaller` line now that we aren't using `nuget.exe`

---
Refers to: azure-pipelines.yml:182 in 3bd65d5. [](commit_id = 3bd65d5bcd859296fefc4c02713c2c2456b355e8, deletion_comment = False)

examples/Microsoft.Spark.FSharp.Examples/Microsoft.Spark.FSharp.Examples.fsproj
eerhardt: You can create a `Directory.Build.props` file in the root `examples` folder and put this `<IsPackable>` setting in it, so it doesn't need to be duplicated across all examples.

---
In reply to: [278753970](https://github.com/dotnet/spark/pull/54#discussion_r278753970) [](ancestors = 278753970)

azure-pipelines.yml
eerhardt: @imback82 - do you envision any more nupkgs in the future? I agree with @alefranz here - it is best to have the .yml file say "run pack", and have the projects themselves decide whether they are packable or not. It allows for more packages without having to update this step, which may be missed.

But if we only ever envision a single NuGet package, then hard-coding the one project to pack would be OK.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -1,10 +1,22 @@
...
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <Description>.NET for Apache Spark</Description>
+    <PackageReleaseNotes>Initial release</PackageReleaseNotes>
+    <PackageProjectUrl>https://github.com/dotnet/spark</PackageProjectUrl>
+    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
+    <PackageTags>spark;dotnet;csharp</PackageTags>
+
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
eerhardt: Now we are creating `.snupkg` packages, but we aren't doing anything with them in the build. Can we publish them somewhere? Like in a "BuildArtifact"?

eerhardt: Ideally we would publish them to the Microsoft symbol server, but that is probably outside of the scope of this PR. We can do that in the future.

---
In reply to: [278755885](https://github.com/dotnet/spark/pull/54#discussion_r278755885) [](ancestors = 278755885)

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -15,4 +27,18 @@
...
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-18618-05" PrivateAssets="All"/>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="..\..\scala\microsoft-spark-*\target\microsoft-spark-*.jar" Link="jars\%(Filename)%(Extension)">
+      <Pack>true</Pack>
+      <PackagePath>jars\%(Filename)%(Extension)</PackagePath>
+    </Content>
+    <Content Include="..\build\**" Link="build\%(Filename)%(Extension)">
```
eerhardt: What do you feel about moving this file into a `src/csharp/Microsoft.Spark/build/netstandard2.0` folder?

I have 2 concerns:

1. If we add other nupkgs that need build scripts, they can't go into this root `build` directory, or else they would start being packed here.
2. The .csproj shouldn't be deciding where this file goes. Instead the file system should.  If you put a file in `build/foo/bar/myFile.targets`, it should be in the same folder in the package. 

So instead of the above, you could have:

```xml
<Content Include="build\**\*" Pack="true" PackagePath="build" />
```

And it would pick up any file/folder structure under the `build` folder.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -1,10 +1,22 @@
...
     <RootNamespace>Microsoft.Spark</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(MSBuildProjectName).xml</DocumentationFile>
+
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <Description>.NET for Apache Spark</Description>
+    <PackageReleaseNotes>Initial release</PackageReleaseNotes>
+    <PackageProjectUrl>https://github.com/dotnet/spark</PackageProjectUrl>
+    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
```
eerhardt: This is already set in [https://github.com/dotnet/spark/blob/master/src/csharp/Directory.Build.props](https://github.com/dotnet/spark/blob/master/src/csharp/Directory.Build.props). No need to duplicate it here.

azure-pipelines.yml
safern: Agreed

azure-pipelines.yml
safern: I think there is another nuget package comming for the worker. I think packages should decide if they are packable or not. Also, that way locally if you want to build packages you can just do, dotnet pack Solution.sln and it will generate the packages that the repo builds.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
safern: I think we should also include the License and Third-Parties files that we define on the root of the repo. This is a convention we've followed in other repos. https://github.com/dotnet/iot/blob/master/Directory.Build.props#L21

azure-pipelines.yml
imback82: Sounds good. Let's keep it this way. Thanks for the feedbacks.

examples/Microsoft.Spark.FSharp.Examples/Microsoft.Spark.FSharp.Examples.fsproj
imback82: Cool, this was the solution I was looking for. Thanks @eerhardt !

alefranz:
> > ```
> > - task: NuGetToolInstaller@0
> > ```
> 
> I believe you can remove this `NuGetToolInstaller` line now that we aren't using `nuget.exe`
> 
> Refers to: azure-pipelines.yml:182 in [3bd65d5](https://github.com/dotnet/spark/commit/3bd65d5bcd859296fefc4c02713c2c2456b355e8). [](commit_id = [3bd65d5](https://github.com/dotnet/spark/commit/3bd65d5bcd859296fefc4c02713c2c2456b355e8), deletion_comment = False)

@eerhardt - I thought this was needed to install the `MicroBuildSigningPlugin`, can you confirm it is not needed? I can try to import the pipeline so I can test this job (as it doesn't run for PRs)

examples/Microsoft.Spark.FSharp.Examples/Microsoft.Spark.FSharp.Examples.fsproj
alefranz: Done

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
alefranz: I added a step to publish them as artifacts. Does it help?

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
alefranz: I was trying to not move things around :) but agree it is better with this approach. Done!

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
alefranz: removed. I also moved some other settings which I believe to all packages there.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
alefranz: I've added that as well as moved the Licence expression and some other generic bits to the props file.
Please note this means the third party notices file will show up in the Solution Explorer for each project

eerhardt:
> I thought this was needed to install the MicroBuildSigningPlugin

I don't believe this is needed for the MicroBuildSigningPlugin. We are using an older version of that plugin on ML.NET https://github.com/dotnet/machinelearning/blob/master/build/vsts-ci.yml#L134, and it doesn't require the `NuGetToolInstaller`. Maybe @safern knows for sure, but I believe we can remove the `NuGetToolInstaller`.

eerhardt:
>     <None Include="$(MSBuildThisFileDirectory)../jars/microsoft-spark*.jar">

Is this still the right path now that this file has moved?

---
Refers to: src/csharp/Microsoft.Spark/build/netstandard2.0/Microsoft.Spark.targets:3 in 008abc5. [](commit_id = 008abc5c39be7aba2ed215e93a889d4932f13e83, deletion_comment = False)

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -15,4 +19,13 @@
     <PackageReference Include="System.Memory" Version="4.5.2" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-18618-05" PrivateAssets="All"/>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="..\..\scala\microsoft-spark-*\target\microsoft-spark-*.jar" Link="jars\%(Filename)%(Extension)"
+      Pack="true" PackagePath="jars\%(Filename)%(Extension)" />
```
eerhardt: (nit) A more readable multi-line format here would be:

```xml
<Content Include="..\..\scala\microsoft-spark-*\target\microsoft-spark-*.jar"
                Link="jars\%(Filename)%(Extension)"
                Pack="true"
                PackagePath="jars\%(Filename)%(Extension)" />
```

It's easier to see each attribute. You don't have to parse the string, or miss something way off to the right.

alefranz: done

alefranz:
@eerhardt 

I removed the NugetInstall step.

I'm still not able to test on my Azure DevOps account as I don't know how to get the Signing plugin.
`Could not queue the build because there were validation errors or warnings. Job Package: Step references task 'MicroBuildSigningPlugin' at version '2' which does not exist.`

If you could please try to manually run a build of this PR so the `Build.Reason` will not be `PullRequest` and it will run all the tasks, so we can verify that it works correctly.

> > ```
> > <None Include="$(MSBuildThisFileDirectory)../jars/microsoft-spark*.jar">
> > ```
> 
> Is this still the right path now that this file has moved?
> 

I haven't moved the jars, as they are installed by maven. Not sure if needed and in scope.

eerhardt:
> I haven't moved the jars, as they are installed by maven. Not sure if needed and in scope.

You didn't move the jars, but you did move the `.targets` file that is referencing the jars using `$(MSBuildThisFileDirectory)../jars/microsoft-spark*.jar`. `$(MSBuildThisFileDirectory)` is the directory that the current MSBuild file (i.e. the `Microsoft.Spark.targets` file) is contained in. Since you moved the .targets file down a directory, you need two `../` here to move up 2 directories.

eerhardt:
I've kicked off a test of the official build using your changes @alefranz. I'll let you know how it goes.

alefranz:
> > I haven't moved the jars, as they are installed by maven. Not sure if needed and in scope.
> 
> You didn't move the jars, but you did move the `.targets` file that is referencing the jars using `$(MSBuildThisFileDirectory)../jars/microsoft-spark*.jar`. `$(MSBuildThisFileDirectory)` is the directory that the current MSBuild file (i.e. the `Microsoft.Spark.targets` file) is contained in. Since you moved the .targets file down a directory, you need two `../` here to move up 2 directories.

Ah sorry, I didn't realize you were referring to the targets file. Fixed it.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -15,4 +19,15 @@
     <PackageReference Include="System.Memory" Version="4.5.2" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-18618-05" PrivateAssets="All"/>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="..\..\scala\microsoft-spark-*\target\microsoft-spark-*.jar"
+      Link="jars\%(Filename)%(Extension)"
```
eerhardt: (nit) I wasn't clear about the formatting, but we typically align the attributes underneath the attribute on the first line. This isn't 100% necessary to fix, but if you are changing something else, consider this.

alefranz: no worries, fixed

alefranz:
Updated top of description to correctly specify the issues it fixes, also added #40

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
imback82: How about `Visible="false"`?
```
<Content Include="$(MSBuildThisFileDirectory)..\..\THIRD-PARTY-NOTICES.TXT" Pack="true" PackagePath="\"  Visible="false"/>
```

eng/Sign.proj
```diff
@@ -36,7 +36,7 @@
     </ItemGroup>
 
     <ItemGroup Condition="'$(SignNugetPackages)' == 'true'">
-      <FilesToSign Include="$(OutDir)*.nupkg" Exclude="$(OutDir)*.symbols.nupkg">
+      <FilesToSign Include="$(OutDir)*.nupkg">
```
safern: I think we should also sign snupkg, right? Since the plan is for it to go into nuget? @eerhardt what do you think?

safern: Opps we merged right at the same moment I commented on this, but @imback82 I think we want to solve this.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
safern: Yes, let's fix this on a follow up PR 😄

eng/Sign.proj
eerhardt: I guess it depends on what we plan on doing with them... if we just use it to publish the symbols to the Microsoft Symbol Server during the build, they probably don't need to be signed.
But if we do plan on uploading them to nuget.org, then maybe they need to be signed?

@dagood @mikem8361 - for thoughts/recommendations on what should be done with `.snupkg` files.

dagood: @tmat knows the most about `snupkg`s, I think.

tmat: snupkgs should be published to NuGet.org **and** the PDBs to Microsoft Symbol servers.

tmat:
Have you considered using [Arcade SDK](https://github.com/dotnet/arcade/blob/master/Documentation/ArcadeSdk.md)? It's kind of sad to see folks spending a time on their infrastructure reinventing the wheel for the hundredth time when the same work have already been done and it's easy to use.

Issue #55, opened 2019-04-25
[BUG]: DataFrameWriter PartitionBy fails - cannot find matching method class org.apache.spark.sql.DataFrameWriter.partitionBy


GoEddie:
**Describe the bug**

DataFrameWriter.PartitionBy fails to execute. PartitionBy takes an array of strings as the parameters (params string[]) and when the scala code tries to find the matching method on ```org.apache.spark.sql.DataFrameWriter``` it fails because it is expecting either a Seq or a String[] but instead (*I think* at least) it is looking for one string argument for every column passed in so String, String.

19/04/25 21:55:50 WARN DotnetBackendHandler: cannot find matching method class org.apache.spark.sql.DataFrameWriter.partitionBy. Candidates are:
19/04/25 21:55:50 WARN DotnetBackendHandler: partitionBy(class [Ljava.lang.String;)
19/04/25 21:55:50 WARN DotnetBackendHandler: partitionBy(interface scala.collection.Seq)
...
19/04/25 21:55:50 ERROR DotnetBackendHandler: args:
19/04/25 21:55:50 ERROR DotnetBackendHandler: argType: java.lang.String, argValue: a
19/04/25 21:55:50 ERROR DotnetBackendHandler: argType: java.lang.String, argValue: b
[2019-04-25T20:55:50.0447941Z] [TEDTOP] [Error] [JvmBridge] JVM method execution failed: Nonstatic method partitionBy failed for class 13 when called with 2 arguments ([Index=1, Type=String, Value=a], [Index=2, Type=String, Value=b], )

**To Reproduce**

If you use the basic example to read a parquet file then call PartitionBy:

```csharp
            SparkSession spark = SparkSession
                .Builder()
                .AppName(".NET Spark SQL basic example")
                .Config("spark.some.config.option", "some-value")
                .GetOrCreate();

            var df = spark.Read().Load(args[0]);
            df.Write()
                .PartitionBy("favorite_color")
                .SaveAsTable("people_partitioned_bucketed");
```

run it using:

```powershell
spark-submit.cmd --class org.apache.spark.deploy.DotnetRunner `
 --master local  `
 C:\github\dotnet-spark\src\scala\microsoft-spark-2.4.x\target\microsoft-spark-2.4.x-0.1.0.jar `
 Microsoft.Spark.CSharp.Examples.exe Sql.Basic %SPARK_HOME%\examples\src\main\resources\users.parquet

```

**Expected behavior**

I would expect to be able to partition when writing

[...]

PR #56, opened 2019-04-25
DataFrameWriter PartitionBy fails

imback82:
Addresses #55 

Will be adding E2E tests.


imback82:
Unit tests will be addressed in #59

PR #62, opened 2019-04-26
 Create ConvertTpchCsvToParquetApp.scala

wentaowu:


benchmark/scala/src/main/scala/com/microsoft/tpch/ConvertTpchCsvToParquetApp.scala
```diff
@@ -0,0 +1,137 @@
+/*
+ * Licensed to the .NET Foundation under one or more agreements.
+ * The .NET Foundation licenses this file to you under the MIT license.
+ * See the LICENSE file in the project root for more information.
+ */
+
+package mypackage
```
imback82: package com.microsoft.tpch

benchmark/scala/src/main/scala/com/microsoft/tpch/ConvertTpchCsvToParquetApp.scala
```diff
@@ -0,0 +1,137 @@
...
+import org.apache.spark.sql.{SaveMode, SparkSession}
+
+object ConvertTpchCsvToParquetApp {
+  def main(args: Array[String]): Unit = {
+    val spark = SparkSession.builder()
+      .appName(this.getClass.getName)
+      .getOrCreate()
+
+    // The 1st arg should be the folder containing TPCH tables.
+    val tpchSourceFolderPath = args(1)
```
imback82: I thought it is 0 based, no?

benchmark/scala/src/main/scala/com/microsoft/tpch/ConvertTpchCsvToParquetApp.scala
```diff
@@ -0,0 +1,137 @@
...
+ */
+
+package mypackage
+
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{SaveMode, SparkSession}
+
+object ConvertTpchCsvToParquetApp {
+  def main(args: Array[String]): Unit = {
+    val spark = SparkSession.builder()
```
imback82: something like:
```scala
if (args.length != 2) {
  println("Usage:")
  println("\t<spark-submit> --master local[*] --class com.microsoft.tpch.ConvertTpchCsvToParquetApp microsoft-spark-benchmark-<version>.jar")
  println("\t\t<input-directory-with-TPCH-tables> <output-directory>")
  return
}
```
?

benchmark/scala/src/main/scala/com/microsoft/tpch/ConvertTpchCsvToParquetApp.scala
laserljy: This make sense, thanks.

benchmark/scala/src/main/scala/com/microsoft/tpch/ConvertTpchCsvToParquetApp.scala
laserljy: I just tested this, it is zero based. My bad, I was looking at the wrong example. Thanks.

benchmark/scala/src/main/scala/com/microsoft/tpch/ConvertTpchCsvToParquetApp.scala
laserljy: That is a good idea, args checking is always needed actually. Thanks!

A small modification on the printout, just following the format of [submitting application](https://spark.apache.org/docs/latest/submitting-applications.html). The instruction for submitting spark for this conversion application will be the same as the following printout.

`    
    if (args.length != 2) {
      println(
        "Error: incorrect command to submit the application, please follow the instruction and specify the path to the source and destination directory as arguments."
      )
      println("The correct command usage should be:")
      println("\t./bin/spark-submit \\")
      println("\t\t--class com.microsoft.tpch.ConvertTpchCsvToParquetApp \\")
      println("\t\t--master local[*] \\")
      println("\t\tmicrosoft-spark-benchmark-<version>.jar \\")
      println(
        "\t\tpath-to-source-directory-with-TPCH-tables path-to-destination-directory-to-save-parquet-file"
      )
      return
    }
`

benchmark/scala/src/main/scala/com/microsoft/tpch/ConvertTpchCsvToParquetApp.scala
```diff
@@ -0,0 +1,152 @@
...
+package com.microsoft.tpch
+
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{SaveMode, SparkSession}
+
+object ConvertTpchCsvToParquetApp {
+  def main(args: Array[String]): Unit = {
+    if (args.length != 2) {
+      println(
+        "Error: incorrect command to submit the application, please follow the instruction and specify the path to the source and destination directory as arguments."
```
imback82: did you run scala-fmt on this? I think you need to turn off the formatting rule if you want to go over 120 max chars.
Also, let's just follow https://github.com/dotnet/spark/blob/master/benchmark/scala/src/main/scala/com/microsoft/tpch/App.scala for error message for consistency.

laserljy: I see. Somehow the formatting rule did not break the long string into 2 lines of strings in my IntelliJ, I manually made it two lines. Following the examples, the if loop goes like this:

```scala
    if (args.length != 2) {
      println("Usage:")
      println("\t<spark-submit> --master local[*] --class com.microsoft.tpch.ConvertTpchCsvToParquetApp microsoft-" +
        "spark-benchmark-<version>.jar")
      println("\t\t<path-to-source-directory-with-TPCH-tables> <path-to-destination-directory-to-save-parquet-file>")
      return
    }
```

@wentaowu Could you please make a change accordingly, thanks.

PR #63, opened 2019-04-29
Improve the pickling deserializaiton performance

adamsitnik:
I have been analysing all the profiles and come up to the conclusion that deserialization, communication via sockets and serialization are the three most common perf bottlenecks in the functional benchmarks.

I did not want to rewrite the sockets to other IPC without further discussing it first so I started working on the deserialization perf.

After getting into the internals of existing unplickler I decided to rewrite it.

In this PR I am introducing a new type called `BufferUnpickler` which is circa a two times faster than `Unpickler`.

The idea is simple: we know the length of the message up front. Instead of using `MaxLengthReadStream` (which adds an overhead to every call to `ReadByte` and `Read`) and checking if we have enough of data in a few places I decided to read the bytes once, to an array from ArrayPool. So we do the IO once (instead of multiple calls to ReadByte and Read) and in most of the cases (when messageSize < ArrayPool.Shared.MaxSize) we don't allocate any managed memory. Morever, we check if we have enough of data only once, after reading it from the stream.

Once we have everything in a memory buffer we wrap in into `ReadOnlySpan<byte>` and pass to `BufferUnpickler`. `BufferUnpickler` processes the data in a similar way `Unpickler` did. The difference is that:
- `stream.ReadByte` is just `span[0]`
- `stream.Read(buffer, 0, size)` is just `span.Slice(0, size)`
This helps to avoid a lof of virtual method calls and `Array.Copy`.

The gain I got for deserializaiton of sample input from Q1 benchmark is x2.

|                                     Method |      Mean |     Error |    StdDev | Ratio |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------------------------------------- |----------:|----------:|----------:|------:|-------:|-------:|------:|----------:|
| GetUnpickledObjectsFromMaxLengthReadStream | 10.921 us | 0.0163 us | 0.0145 us |  1.00 | 1.7242 | 0.1068 |     - |  10.65 KB |
|              GetUnpickledObjectsFromBuffer |  5.083 us | 0.0110 us | 0.0103 us |  0.47 | 1.7166 | 0.1068 |     - |   10.6 KB |

In all the profiles the deserialization takes also two times less time now:

obraz

obraz

obraz

obraz


src/csharp/Microsoft.Spark/Pyrolite/UnpickleStack.cs
```diff
@@ -16,7 +16,7 @@ namespace Razorvine.Pickle
 [SuppressMessage("ReSharper", "InconsistentNaming")]
 internal class UnpickleStack {
 	public readonly object MARKER;
-    private const int DefaultCapacity = 4;
+    private const int DefaultCapacity = 128;
```
adamsitnik: this small change was very important: the unpickler adds every deserialized object to this collection.

if we deserialize 100 doubles, it would allocate an array of 4 objects, double it to 8 (and copy all elements), 16, 32, 64, 128 etc..

128 is better because we hardly ever want to deserialize and send only a few objects from one process to another

src/csharp/Microsoft.Spark/RDD/Collector.cs
```diff
@@ -66,18 +66,10 @@ internal interface IDeserializer
...
-            private static MaxLengthReadStream s_slicedReadStream;
-
             private readonly BinaryFormatter _formater = new BinaryFormatter();
 
             public object Deserialize(Stream stream, int length)
             {
-                MaxLengthReadStream readStream = s_slicedReadStream ??
-                    (s_slicedReadStream = new MaxLengthReadStream());
-
-                readStream.Reset(stream, length);
```
adamsitnik: I have removed `MaxLengthReadStream` because it's not used.

As you can see in the line below the original stream is passed to the formatter, not the instance of `MaxLengthReadStream`.

src/csharp/Microsoft.Spark/Pyrolite/BufferUnpickler.cs
```diff
@@ -0,0 +1,859 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
imback82: Should we contribute this to https://github.com/irmen/Pyrolite? Note that we copied Pyrolite code because the dll was not being signed with a strong name as of last week. Now that it is being signed, we plan to remove Pyrolite code from our repo.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -118,20 +116,31 @@ internal class PicklingSqlCommandExecutor : SqlCommandExecutor
...
-                    //   |null|
-                    //   |  11|
-                    //   +----+
-                    object[] inputRows = PythonSerDe.GetUnpickledObjects(readStream);
+                    byte[] buffer = ArrayPool<byte>.Shared.Rent(messageLength);
+                    object[] inputRows = null;
+
+                    try
+                    {
+                        if (inputStream.Read(buffer, 0, messageLength) != messageLength)
```
eerhardt: Should we loop until we've read that many bytes or until Read returns 0?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -38,5 +43,13 @@ internal static object[] GetUnpickledObjects(Stream s)
             Debug.Assert(unpickledItems != null);
             return (unpickledItems as object[]);
         }
+
+        internal static object[] GetUnpickledObjects(ReadOnlySpan<byte> memory)
```
eerhardt: (nit) naming the parameter `memory` feels a bit odd, since it is a Span.

src/csharp/Microsoft.Spark/Pyrolite/BufferUnpickler.cs
adamsitnik: It depends on how fast we want to go.

The BufferUnpickler introduced in this PR offers the same features as the original Unipicker. 

The thing is that we don't need all of the features it offers and we could make it tailored to our needs and even faster.

An example: the Pickler uses internally a stack of deserialized objects. If we want to deserialize a hundred double pairs (two doubles), it's going to allocate a hundred of `object[2]` arrays with two doubles in each array and store them on the mentioned stack. It's ok for general purpose.

But in our case, the `SqlCommandExecutor` always iterates over the deserialized `IEnumerable<object[]>` in a loop and it uses only one element at a time. So instead of allocating `new object[2]` for every pair of doubles, we could have one and make the pickler return a custom `IEnumerable<object[]>` which would be reusing the cached array

```cs
object[] inputRows = PythonSerDe.GetUnpickledObjects(new ReadOnlySpan<byte>(buffer, 0, messageLength));
for (int i = 0; i < inputRows.Length; ++i)
    outputRows.Add(commandRunner.Run(0, inputRows[i]));
```

Could become:

```cs
IEnumerable<object[]> inputRows = PythonSerDe.GetUnpickledObjects(new ReadOnlySpan<byte>(buffer, 0, messageLength));

foreach(object[] input row in inputRows)
    outputRows.Add(commandRunner.Run(0, inputRow)); // input row would be always the same instance of array (depending on size), but pointing to different values
```

adamsitnik: Most probably I should contribute it to the Pyrolite now and in the future if we decided to go the extreme path, imlement a tailored solution.

src/csharp/Microsoft.Spark/Pyrolite/BufferUnpickler.cs
```diff
@@ -0,0 +1,859 @@
...
+    /// <summary>
+    /// Unpickles an object graph from a pickle data memory buffer. Supports all pickle protocol versions.
+    /// Maps the python objects on the corresponding java equivalents or similar types.
+    /// This class is NOT threadsafe! (Don't use the same unpickler from different threads)
+    /// See the README.txt for a table with the type mappings.
+    /// </summary>
+    [SuppressMessage("ReSharper", "InconsistentNaming")]
+    [SuppressMessage("ReSharper", "MemberCanBePrivate.Global")]
+    [SuppressMessage("ReSharper", "InvertIf")]
+    internal class BufferUnpickler
```
eerhardt: Do we really need to duplicate the whole class? There appears to be a lot of duplication between this and Unpickler.cs.

One pattern I used in the Arrow repo was to have the "public" class be the same, and then internally under the covers used a "strategy" pattern for the "implementation" over a Stream or Memory.

Public class with 2 constructors:
[https://github.com/apache/arrow/blob/master/csharp/src/Apache.Arrow/Ipc/ArrowStreamReader.cs#L37-L48](https://github.com/apache/arrow/blob/master/csharp/src/Apache.Arrow/Ipc/ArrowStreamReader.cs#L37-L48)

2 internal implementations:
[https://github.com/apache/arrow/blob/master/csharp/src/Apache.Arrow/Ipc/ArrowStreamReaderImplementation.cs](https://github.com/apache/arrow/blob/master/csharp/src/Apache.Arrow/Ipc/ArrowStreamReaderImplementation.cs)
[https://github.com/apache/arrow/blob/master/csharp/src/Apache.Arrow/Ipc/ArrowMemoryReaderImplementation.cs](https://github.com/apache/arrow/blob/master/csharp/src/Apache.Arrow/Ipc/ArrowMemoryReaderImplementation.cs)

And an internal base class for shared logic:
[https://github.com/apache/arrow/blob/master/csharp/src/Apache.Arrow/Ipc/ArrowReaderImplementation.cs](https://github.com/apache/arrow/blob/master/csharp/src/Apache.Arrow/Ipc/ArrowReaderImplementation.cs)

thoughts?

adamsitnik:
I am going to close this PR and send a new one to https://github.com/irmen/Pyrolite

@eerhardt you are right, a lot of code duplication was unnecessary. I am going to reduce it without sacrificing the perf.

PR #69, opened 2019-04-29
Add LICENSE file to Nuget package

### LINKED ISSUE #42: Add license and 3rd party notices to nuget package
We aren't currently putting the license and 3rd party notice files in our nuget package. We should add them.

imback82:
Fixes #42

### DIFF
--- a/src/csharp/Directory.Build.props
+++ b/src/csharp/Directory.Build.props
@@ -34,6 +34,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Content Include="$(MSBuildThisFileDirectory)..\..\LICENSE"
+         Pack="true"
+         PackagePath="\"
+         Visible="false" />
     <Content Include="$(MSBuildThisFileDirectory)..\..\THIRD-PARTY-NOTICES.TXT"
              Pack="true"
              PackagePath="\"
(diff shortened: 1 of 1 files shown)


src/csharp/Directory.Build.props
```diff
@@ -34,6 +34,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Content Include="$(MSBuildThisFileDirectory)..\..\LICENSE"
```
eerhardt: I wonder if we should change the `PackageLicenseExpression` to `<PackageLicenseFile>LICENSE</PackageLicenseFile>` instead.

That way there is only 1 "truth" or "master" license - the file that is in the root of the repo. If we ever changed that file (to say Apache 2.0), the nuget package would say it is `MIT`, but this LICENSE file would say `Apache.

src/csharp/Directory.Build.props
```diff
@@ -34,6 +34,10 @@
...
+         Pack="true"
```
eerhardt: (nit) it is nice to have the attributes lined up vertically with the attribute on the first line. See the `<Content` xml below for an example.

src/csharp/Directory.Build.props
imback82: fixed.

src/csharp/Directory.Build.props
imback82: Fixed.

PR #70, opened 2019-04-29
Getting ready for releasing v0.2.0

imback82:


docs/release-notes/0.2/release-0.2.md
```diff
@@ -0,0 +1,8 @@
+# .NET for Apache Spark 0.2 Release Notes
+
+Below are some of the highlights from this release.
```
rapoth: Since the first version did not have a release notes, I suggest adding a one-time retrospective release notes here (e.g., we support Spark versions blah) so that people who look at this in Nuget have a good idea of what this does and what it doesn't.

suhsteve:
LGTM

PR #71, opened 2019-04-30
Update build pipeline to use .NET Core 2.2.105

imback82:
#69 changed `<PackageLicenseExpression>MIT</PackageLicenseExpression>` to `<PackageLicenseFile>LICENSE</PackageLicenseFile>`. However, this introduced the following error in the Nuget generation pipeline:
`error NU5030: The license file 'LICENSE' does not exist in the package.`

This seems to be related to https://github.com/NuGet/Home/issues/7591 and the dotnet installed on the build machine is 2.1.503.

The quick solution is to install .NET Core 2.2.105 in the build machines.

Also, I cleaned up `IsPackable` so that by default it is false everywhere and we have only one project where `IsPackable` is set to true.


eerhardt:
@chcosta @mattgal - do you know how to get the latest `2.2` .NET SDK installed on Pool:
NetCoreInternal-Int-Pool? https://dev.azure.com/dnceng/internal/_settings/agentqueues?_a=agents&queueId=116&view=jobs

This bug has been fixed with the latest SDK, but it appears the build agents don't have the 2.2 SDK on them:

```
C:\Program Files\dotnet\sdk\2.1.503\Sdks\Microsoft.NET.Sdk\Sdk\Sdk.props
```

cc @safern

MattGal:
@eerhardt :
The configuration of these machines is controlled via the yaml files in this internal AzDO repo: https://dnceng.visualstudio.com/internal/_git/dotnet-helix-machines 

However, we'd generally advise for things which can be easily bootstrapped like a dotnet SDK that a build bootstrap them itself; is there a specific reason why this is not possible?

eerhardt:
> However, we'd generally advise for things which can be easily bootstrapped like a dotnet SDK that a build bootstrap them itself; is there a specific reason why this is not possible?

No, I believe doing that is possible.

@imback82 - It looks like a simple way to enable this is to use the `DotNetCoreInstaller@0` task:

https://docs.microsoft.com/en-us/azure/devops/pipelines/languages/dotnet-core?view=azure-devops#build-environment

Can we use that with version `2.2.105`?

Another (potentially long-term) option here would be to use the `dotnet/arcade` tooling, which will also bootstrap the .NET SDK version. But I believe that would be more work which may not be beneficial right now.

safern:
> @imback82 - It looks like a simple way to enable this is to use the DotNetCoreInstaller@0 task:
> https://docs.microsoft.com/en-us/azure/devops/pipelines/languages/dotnet-core?view=azure-devops#build-environment
> Can we use that with version 2.2.105?

Yes I think that is the best option at the moment.

azure-pipelines.yml
```diff
@@ -23,6 +23,10 @@ jobs:
   pool: Hosted VS2017
 
   steps:
+  - task: DotNetCoreInstaller@0
+    inputs:
+    version: '2.2.105'
```
safern: Can we instead define a variable in the global variables and share it on both steps? That way we're always consistent on the version we install for bot jobs.

imback82: Yes. Will do. Thanks!

imback82:
Thanks @safern / @eerhardt . I verified that this works in the internal repo.

azure-pipelines.yml
```diff
@@ -25,7 +26,7 @@ jobs:
   steps:
   - task: DotNetCoreInstaller@0
     inputs:
-      version: '2.2.105'
+      version: '$(dotnetCoreVersion)'
```
safern: Nit: you don't need to quote this.

PR #73, opened 2019-04-30
Updating Readme to reflect the use of dotnet.exe

rapoth:
On Windows, when you use .NET Core 2.1, building the solution will produce a DLL as opposed to an EXE. This means the user would have to pass in the path to dotnet.exe and not the DLL itself.


suhsteve:
LGTM

PR #76, opened 2019-04-30
Sign snupkg packages during build process

imback82:
Currently, only the nuget packages are signed. Update build pipieline to sign snupkg packages as well.

### DIFF
--- a/eng/Sign.proj
+++ b/eng/Sign.proj
@@ -41,6 +41,12 @@
       </FilesToSign>
     </ItemGroup>
 
+    <ItemGroup Condition="'$(SignNugetPackages)' == 'true'">
+      <FilesToSign Include="$(OutDir)*.snupkg">
+        <Authenticode>NuGet</Authenticode>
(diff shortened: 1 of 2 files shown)


eng/Sign.proj
```diff
@@ -41,6 +41,12 @@
       </FilesToSign>
     </ItemGroup>
 
+    <ItemGroup Condition="'$(SignNugetPackages)' == 'true'">
+      <FilesToSign Include="$(OutDir)*.snupkg">
```
eerhardt: You can collapse this into the above `ItemGroup`:

```xml
    <ItemGroup Condition="'$(SignNugetPackages)' == 'true'">
      <FilesToSign Include="$(OutDir)*.nupkg;$(OutDir)*.snupkg">
        <Authenticode>NuGet</Authenticode>
      </FilesToSign>
    </ItemGroup>
```

It saves a little bit on duplication.

imback82: Thanks!

PR #80, opened 2019-05-01
Worker installation script changes to deal with versioning

rapoth:
This PR brings changes to accommodate versioning in our releases.

### DIFF
--- a/deployment/install-worker.sh
+++ b/deployment/install-worker.sh
@@ -5,54 +5,69 @@
 # This is a helper script to install the worker binaries on your Apache Spark cluster
 #
 # Usage:
-# ./install-script.sh <cloud-provider> <path-to-worker> <local-worker-installation-path>
+# ./install-script.sh <release-provider> <path-to-worker-release> <local-worker-installation-path>
 #
 # Sample usage:
-# ./install-script.sh 
-#   azure 
-#   abfs://<blobcontainer>@<gen2storageaccount>.dfs.core.windows.net/<path>/Microsoft.Spark.Worker.netcoreapp2.1.linux-x64-0.1.0.tar.gz 
-#   /usr/local/bin
+# ./install-script.sh
+#    github
+#    https://github.com/dotnet/spark/releases/download/v0.1.0/Microsoft.Spark.Worker.netcoreapp2.1.linux-x64-0.1.0.tar.gz
+#    /usr/local/bin
+#
+# or if you have your Worker release on filesystem like ABFS, here's how the path would
+# look like:
+# ./install-script.sh
+#    azure
+#    abfs://<blobcontainer>@<gen2storageaccount>.dfs.core.windows.net/<path>/Microsoft.Spark.Worker.netcoreapp2.1.linux-x64-0.1.0.tar.gz
+#    /usr/local/bin
+#
 ##############################################################################
 
 set +e
+
+# Uncomment if you want full tracing (for debugging purposes)
+#set -o xtrace
+
 # Cloud Provider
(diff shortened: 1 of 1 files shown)


deployment/install-worker.sh
```diff
@@ -1,40 +1,69 @@
...
+# Description:
+# This is a helper script to install the worker binaries on your Apache Spark cluster
+#
+# Usage:
+# ./install-script.sh <cloud-provider> <path-to-worker> <local-worker-installation-path>
+#
+# Sample usage:
+# ./install-script.sh 
+#   azure 
+#   abfs://<blobcontainer>@<gen2storageaccount>.dfs.core.windows.net/<path>/Microsoft.Spark.Worker.netcoreapp2.1.linux-x64-0.1.0.tar.gz
```
imback82: I am wondering if we should move to `wget` and let the user to point to our release on GitHub (verified on HDInsight). We don't have to do this in this PR, but we may want to track this if we want to simplify the experience.
cc: @suhsteve

deployment/install-worker.sh
```diff
@@ -1,40 +1,69 @@
...
+
 # The path on the executor nodes where Microsoft.Spark.Worker executable is installed.
-DEST_PATH=$3
+LOCAL_WORKER_PATH=$3
+
 # The path where all the dependent libraies are installed so that it doesn't
-# pollute the $DEST_PATH.
-DEST_PATH_BINARIES=$DEST_PATH/microsoft.spark.worker
+# pollute the $LOCAL_WORKER_PATH.
+SPARKDOTNET_ROOT=$LOCAL_WORKER_PATH/spark-dot-net
```
imback82: `dotnet`

deployment/install-worker.sh
```diff
@@ -1,40 +1,69 @@
...
+#   /usr/local/bin
+##############################################################################
+
+set +e
 # Cloud Provider
 CLOUD_PROVIDER=$1
+
 # Path where packaged worker file (tgz) exists.
-WORKER_PATH=$2
+WORKER_PATH_OR_URI=$2
```
imback82: I think we need SRC_ and DEST_ . LOCAL_WORKER_PATH is still not clear.

deployment/install-worker.sh
```diff
@@ -1,40 +1,69 @@
...
-sudo rm -rf $DEST_PATH_BINARIES
+# Extract version
+IFS='-' read -ra BASE_FILENAME <<< "$(basename $WORKER_PATH_OR_URI .tar.gz)"
+VERSION=${BASE_FILENAME[2]}
+
+IFS='.' read -ra VERSION_CHECK <<< "$VERSION"
+[[ ${#VERSION[@]} != 3 ]] || { echo >&2 "Version check does not satisfy. Raise an issue here: https://github.com/dotnet/spark"; exit 1; }
+
+# Path of the final worker binaries (the one we just downloaded and extracted)
+CURRENT_WORKER_PATH=$SPARKDOTNET_ROOT/Microsoft.Spark.Worker-$VERSION
```
imback82: why is this current and the comment is final?

rapoth: Done

deployment/install-worker.sh
rapoth: Done

deployment/install-worker.sh
rapoth: Done

deployment/install-worker.sh
rapoth: Added the capability.

PR #81, opened 2019-05-01
Remove formatPath() that adds ".\" or "./" to the executable.

imback82:
`formatPath()` adds ".\" or "." to the .NET executable. This forces the user to specify a full path for the .NET executable passed to `spark-submit`. For example,
```
spark-submit --class org.apache.spark.deploy.DotnetRunner microsoft-spark-<version>.jar dotnet HelloSpark.dll
```
is not allowed, since `formatPath()` changes `dotnet` to `.\dotnet.exe` even if dotnet is in the PATH.

### DIFF
--- a/src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/deploy/DotnetRunner.scala
+++ b/src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/deploy/DotnetRunner.scala
@@ -71,11 +71,6 @@ object DotnetRunner extends Logging {
         // Reuse windows-specific formatting in PythonRunner.
         dotnetExecutable = PythonRunner.formatPath(resolveDotnetExecutable(driverDir, args(1)))
         otherArgs = args.slice(2, args.length)
-      } else if (new File(args(0)).isDirectory) {
-        // Local mode where .NET application is located in the given directory.
-        // Reuse windows-specific formatting in PythonRunner.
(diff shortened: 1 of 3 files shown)


src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/deploy/DotnetRunner.scala
```diff
@@ -71,11 +71,6 @@ object DotnetRunner extends Logging {
         // Reuse windows-specific formatting in PythonRunner.
         dotnetExecutable = PythonRunner.formatPath(resolveDotnetExecutable(driverDir, args(1)))
         otherArgs = args.slice(2, args.length)
-      } else if (new File(args(0)).isDirectory) {
```
imback82: Removing this since there is no use case for this.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/deploy/DotnetRunner.scala
```diff
@@ -200,17 +195,6 @@ object DotnetRunner extends Logging {
     resolvedExecutable
   }
 
-  // When executing in YARN cluster mode, the name of the
```
imback82: Note that for YARN cluster mode, the user has to ship the executable with dependencies via Zip file and the full path for the executable is constructed from it. Thus, this function is not needed.

suhsteve:
LGTM

Issue #78, opened 2019-04-30
Automate packaging release artifacts


imback82:
The build pipieline should automatically produce the following worker binaries for release:
  * Microsoft.Spark.Worker.net461.win-x64-0.1.0.zip
  * Microsoft.Spark.Worker.netcoreapp2.1.linux-x64-0.1.0.tar.gz
  * Microsoft.Spark.Worker.netcoreapp2.1.win-x64-0.1.0.zip

imback82:
Closed by #82

PR #82, opened 2019-05-01
Automate packaging release artifacts

suhsteve:
Address #78


script/package-worker.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+NUGET_DIR=$1
+WORKER_DIR=$2
+OUTPUT_DIR=$3
+
+VERSION=`ls ${NUGET_DIR}/Microsoft.Spark.*.nupkg | head -n 1 | xargs -n 1 basename | cut -d'.' -f3-5`
```
rapoth: You might want to take a look at how I did this in [install-worker.sh](https://github.com/dotnet/spark/blob/master/deployment/install-worker.sh#L46) and potentially put in a version check in case the version string is not as expected? (just to be on the safe side)

azure-pipelines.yml
```diff
@@ -280,3 +280,31 @@ jobs:
       inputs:
         pathtoPublish: '$(Build.ArtifactStagingDirectory)/SymbolPackages'
         artifactName:  Microsoft.Spark.Binaries
+
+- ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
```
imback82: do you need a separate job? Can't you do this in a previous job?

suhsteve: Unfortunately it seems the previous job's image doesn't have bash.  I think it is possible to use the previous job, but it would require more steps as well as hard coding the version / runtime / framework values.

suhsteve: I'll look into this a bit more.

imback82: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/bash?view=azure-devops wouldn't work?

azure-pipelines.yml
```diff
@@ -280,3 +280,31 @@ jobs:
...
+      vmImage: ubuntu-16.04
+
+    steps:
+    - task: DownloadBuildArtifacts@0
+      displayName: Download Build Artifacts
+      inputs:
+        artifactName: Microsoft.Spark.Binaries
+        downloadPath: $(Build.ArtifactStagingDirectory)
+
+    - task: Bash@3
```
safern: This has a simplified syntax:

```
- bash: script/package-worker.sh Microsoft.Spark.Binaries Microsoft.Spark.Binaries/Microsoft.Spark.Worker Microsoft.Spark.Worker.Packages
  workingDirectory: $(Build.ArtifactStagingDirectory)
  displayName: Package Microsoft.Spark.Worker
```

azure-pipelines.yml
```diff
@@ -280,3 +280,31 @@ jobs:
...
         pathtoPublish: '$(Build.ArtifactStagingDirectory)/SymbolPackages'
         artifactName:  Microsoft.Spark.Binaries
+
+- ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+  - job: PackageWorker
+    dependsOn:
+      - Package
+    displayName: Package Microsoft.Spark.Worker
+    pool:
+      vmImage: ubuntu-16.04
```
safern: You can change this to just be: `pool: Hosted Ubuntu 16.04`

azure-pipelines.yml
safern: You wouldn't need this condition again, since it is already defined above before the package job. So by just adding the right indentation to this job, they both would share the same condition.

suhsteve: Removed the separate job

azure-pipelines.yml
suhsteve: Removed job

azure-pipelines.yml
suhsteve: Using powershell instead

script/package-worker.sh
suhsteve: changed to using powershell.   Don't think we need to check this unless we are going to change the way the nuget filename is going to be named.

PR #91, opened 2019-05-05
[WIP]: Adding multiple Spark Master environment variables

rapoth:
When running in Azure Databricks (see below: Create Job -> Select JAR), spark.master is not set but instead MASTER is set to point to the Master URI. This change introduces the check for MASTER in addition to spark.master.

image

### DIFF
--- a/src/csharp/Microsoft.Spark/Services/ConfigurationService.cs
+++ b/src/csharp/Microsoft.Spark/Services/ConfigurationService.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using System.Configuration;
 using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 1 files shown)


src/csharp/Microsoft.Spark/Services/ConfigurationService.cs
```diff
@@ -22,7 +23,7 @@ internal sealed class ConfigurationService : IConfigurationService
         public const string WorkerReadBufferSizeEnvName = "spark.dotnet.worker.readBufferSize";
         public const string WorkerWriteBufferSizeEnvName = "spark.dotnet.worker.writeBufferSize";
 
-        private const string SparkMasterEnvName = "spark.master";
+        private readonly string[] SparkMasterEnvName = new string[] { "spark.master", "MASTER" };
```
imback82: I don't think this is the right fix. I think we should just merge DefaultConfiguration/LocalConfiguration/DebugConfiguration into one, so that the configuration creation is independent of spark master env variable.

imback82:
This will be addressed in #92.

rapoth:
Thank you!

PR #92, opened 2019-05-06
Simplify configuration service.

imback82:
This addresses #91 where we remove dependency on spark.master environment variable.


joperezr:
After this change, the instructions here: https://github.com/dotnet/spark/blob/master/docs/getting-started/windows-instructions.md and here: are no longer valid since they show the wrong environment variable name for the worker process. I will submit a PR soon to fix the docs in this repo.

PR #93, opened 2019-05-06
F# project (Microsoft.Spark.FSharp.Examples) build failure on Linux

imback82:
Building Microsoft.Spark.FSharp.Examples on Ubuntu fails with the following message:

```
FSC : error FS2014: A problem occurred writing the binary '/tmp/spark/examples/Microsoft.Spark.FSharp.Examples/obj/Debug/netcoreapp2.1/Microsoft.Spark.FSharp.Examples.dll':
A call to StrongNameSignatureSize failed (Invalid Public Key blob) [/tmp/spark/examples/Microsoft.Spark.FSharp.Examples/Microsoft.Spark.FSharp.Examples.fsproj]
```

This seems to be related to:
https://github.com/Microsoft/visualfsharp/issues/680#issuecomment-147905529
https://stackoverflow.com/questions/3733148/problems-signing-a-dll-in-f-on-mono

Currently, I am disabling signing since we are not publishing this dll anyway.

@eerhardt @safern any thoughts?


eerhardt:
> Currently, I am disabling signing since we are not publishing this dll anyway.

We did the same thing in ML.NET:

https://github.com/dotnet/machinelearning/pull/600/files#diff-254a1dfbeae9416ca2b453611c894683R7

From @dsyme:

> This is due to an F# compiler issue when using /publicsign+ (i.e. think the problem is when using a test key that only containing a public key token. Or something like that). I'll add an F# compiler bug

> In the meantime we work around by not using /publicsign for the F# test assembly (since it's only a test assembly it doesn't matter either way)

@dsyme - do you know if that F# compiler issue got fixed? If so, which version of the .NET Core SDK was it fixed in?

PR #106, opened 2019-05-14
Add config option for installing .NET Core runtime in shared mode

### LINKED ISSUE #105: Add config option for installing .NET Core runtime in shared mode
Giving the option within the `db-init.sh` initialisation script to install .NET Core Runtime in shared mode would allow for smaller application bundles.

When running `dotnet publish` then bundling within a .zip file, it produces around 200 files and an almost 30MB .zip file.  Having the .NET Core Runtime installed on the workers in shared mode allows us to compile with the `--self-contained=false` flag producing around 20 files which bundle into a less than 1MB .zip file.

AeroXuk:
This adds config options for installing the .NET Core Runtime in shared mode.  Allowing for smaller app bundles.

Fixes #105


suhsteve:
Thank you for your interest and support, but this is not a change we can accept at this time for the following reasons:

1) We would like to keep the worker installation and application execution experience as similar as possible between the various cloud platforms.  Other scenarios would need to be updated and verified.

2) The worker and applications are self-published in order to have a bit more control over the dotnet versions used to run the binaries.  This also allows us to avoid having to install/overwrite dotnet versions that users may already have installed on their clusters.

AeroXuk:
@suhsteve Thank you for your review. Would information on how to switch from self-published to shared install be better as a documentation piece?  I could see other people wanting smaller application packages if they intend to have many applications deployed to DBFS for different scenarios.

PR #111, opened 2019-05-17
Improve Pickling performance

adamsitnik:
This PR contains perf improvements introduced to Pyrolite in the following PRs:

https://github.com/irmen/Pyrolite/pull/64
https://github.com/irmen/Pyrolite/pull/67

Depending on the scenario (bools, double, strings, integers) the serialization and deserialization should be from 20% to 100% faster.


eerhardt:
>         internal static object[] GetUnpickledObjects(Stream s)

Do we still need the `GetUnpickledObjects(Stream s)` overload?

---
Refers to: src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs:34 in d4100f5. [](commit_id = d4100f5e7df1a0694315c0be9116bfbf36b923a8, deletion_comment = False)

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -38,5 +39,18 @@ internal static object[] GetUnpickledObjects(Stream s)
...
+
+        /// <summary>
+        /// Unpickles objects from memory buffer
+        /// </summary>
+        /// <param name="buffer">Pickled byte buffer</param>
+        /// <returns>Unpicked objects</returns>
+        internal static object[] GetUnpickledObjects(ReadOnlyMemory<byte> buffer)
+        {
+            var unpickler = new Unpickler();
+            var unpickledItems = unpickler.loads(buffer, stackCapacity: 100); // 100 because spark typically sends batches of 100 rows
```
eerhardt: Should the `100` be configurable at all?  I know that for the Arrow format, the number of rows in a batch is configurable.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -153,22 +162,22 @@ internal class PicklingSqlCommandExecutor : SqlCommandExecutor
...
+                s_output = new byte[estimatedMaxSize];
 
             Pickler pickler = s_pickler ?? (s_pickler = new Pickler(false));
-            pickler.dump(rows, writeOutputStream);
+            pickler.dumps(rows, ref s_output, out int bytesWritten);
 
-            if (writeOutputStream.Position == 0)
+            if (bytesWritten == 0)
             {
                 throw new Exception("Message buffer cannot be null.");
```
eerhardt: Not really your change - but this exception message isn't illustrative of what happened. Would it be possible to fix this message while you are here?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -118,28 +117,38 @@ internal class PicklingSqlCommandExecutor : SqlCommandExecutor
...
+                    }
 
                     for (int i = 0; i < inputRows.Length; ++i)
                     {
                         // Split id is not used for SQL UDFs, so 0 is passed.
                         outputRows.Add(commandRunner.Run(0, inputRows[i]));
                     }
 
-                    WriteOutput(outputStream, outputRows);
+                    WriteOutput(outputStream, outputRows, messageLength);
```
eerhardt: The input `messageLength` doesn't really have anything to do with the output message length. This feels like an odd association/heuristic.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
adamsitnik: @imback82 is the number of rows in a batch from spark configurable when using Pickling?

I have used 100 because this is what I have seen in the profiler. Also, this is only the initial capacity, not max capacity.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
adamsitnik: Good point, I have updated the message.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
adamsitnik: The output size should be <= input (I have verified that for Q1, Q8 and Q13). It's going to be allocated once and then reused many times. That was the best estimation I could come up with based on the available data.

adamsitnik:
> Do we still need the GetUnpickledObjects(Stream s) overload?

Good point. I have refactored the code and now we have only `GetUnpickledObjects(Stream s, int messageSize)`

src/csharp/Microsoft.Spark/RDD/Collector.cs
```diff
@@ -66,18 +65,10 @@ internal interface IDeserializer
...
             private readonly BinaryFormatter _formater = new BinaryFormatter();
 
             public object Deserialize(Stream stream, int length)
             {
-                MaxLengthReadStream readStream = s_slicedReadStream ??
-                    (s_slicedReadStream = new MaxLengthReadStream());
-
-                readStream.Reset(stream, length);
-
                 return _formater.Deserialize(stream);
```
adamsitnik: important: I have removed the usage of `MaxLengthReadStream` and the type itself because it was not really used here.

```cs
_formater.Deserialize(stream); // stream istead of readStream is passed to the formatter
```

should it be removed? or changed to be actually used?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
eerhardt: Sounds good to me.
Can we add a comment here with that reasoning?

adamsitnik: Sure, I did it right after you asked this question.

https://github.com/dotnet/spark/blob/674a83d1d146e2053d573fda15fdc2a4c2b2d885/src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs#L136

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
imback82: 100 is hard coded on the Spark side: https://github.com/apache/spark/blob/d9bcacf94b93fe76542b5c1fd852559075ef6faa/sql/core/src/main/scala/org/apache/spark/sql/execution/python/BatchEvalPythonExec.scala#L77

Would it hurt the performance a lot if we just use default (4)?

src/csharp/Microsoft.Spark/RDD/Collector.cs
imback82: I think this can be removed.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -131,15 +125,15 @@ internal class PicklingSqlCommandExecutor : SqlCommandExecutor
...
+                    object[] inputRows = PythonSerDe.GetUnpickledObjects(inputStream, messageLength);
 
                     for (int i = 0; i < inputRows.Length; ++i)
                     {
                         // Split id is not used for SQL UDFs, so 0 is passed.
                         outputRows.Add(commandRunner.Run(0, inputRows[i]));
                     }
 
-                    WriteOutput(outputStream, outputRows);
+                    WriteOutput(outputStream, outputRows, estimatedMaxSize: messageLength); // estimatedMaxSize is equal to messageLength because the produced output should be always smaller or of the same size as input
```
imback82: Same here on the max chars in a line.

Also, I didn't get "because the produced output should be always smaller or of the same size as input". How can we guarantee this if you have an UDF like:
```C#
var udf = Udf<string, string>(str => str + str);
```
Am I misunderstanding this comment?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -153,22 +147,22 @@ internal class PicklingSqlCommandExecutor : SqlCommandExecutor
...
         /// <param name="stream">Stream to write to</param>
         /// <param name="rows">Rows to write to</param>
-        private void WriteOutput(Stream stream, IEnumerable<object> rows)
+        /// <param name="estimatedMaxSize">Estimated max size of the serialized output</param>
+        private void WriteOutput(Stream stream, IEnumerable<object> rows, int estimatedMaxSize)
         {
-            MemoryStream writeOutputStream = s_writeOutputStream ??
-                (s_writeOutputStream = new MemoryStream());
-            writeOutputStream.Position = 0;
+            if (s_outputBuffer == null)
```
imback82: https://github.com/dotnet/spark/blob/master/docs/coding-guidelines/csharp-coding-style.md
"A single line statement block must go with braces."

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
adamsitnik: @imback82 thanks for the confirmation. I have set the capacity to avoid unnecessary copies when unpickling and I would prefer to keep it that way.

If we want to unpickle a list of 100 doubles (or any other type) currently an array of 4 elements is allocated, then 8 and 4 are copied, then 16 and 8 are copied, then 32 and 16 are copied then 64 and 32 are copied then 128 and 64 are copied. 

So we allocate 6 arrays of 252 elements in total and copy 124 elements. Setting the capacity to 100 allows us to allocate the array only once and perform 0 copies.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -26,17 +29,29 @@ static PythonSerDe()
...
-            return (unpickledItems as object[]);
+            byte[] buffer = ArrayPool<byte>.Shared.Rent(messageLength);
+
+            try
+            {
+                SerDe.ReadBytes(s, buffer, messageLength);
+
+                // Not making any assumptions about the implementation and hence not a class member.
+                var unpickler = new Unpickler();
+                var unpickledItems = unpickler.loads(new ReadOnlyMemory<byte>(buffer, 0, messageLength), stackCapacity: 100); // 100 because spark typically sends batches of 100 rows
```
imback82:
```C#
object unpickledItems = ...
```

src/csharp/Microsoft.Spark/Interop/Ipc/SerDe.cs
```diff
@@ -183,6 +182,32 @@ public static byte[] ReadBytes(Stream s, int length)
...
+            if (length < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(length), length, "length can't be negative.");
+            }
+            if (length == 0)
+            {
+                return;
+            }
+
+            int bytesRead;
```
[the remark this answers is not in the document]
adamsitnik: I find the design of `TryReadBytes` confusing:

- it returns false when it can't read any bytes (it's ok, it's a Try method)
- it throws an exception when the read is incomplete `totalBytesRead < messageLength` (I would not expect a Try method to throw)

IMHO it should throw in both cases or return false in both. Introducing a new method that always throws when something is wrong seemed like a better approach to me. But it's only a opinion, I can change the code if you want.

src/csharp/Microsoft.Spark.UnitTest/Sql/RowTests.cs
```diff
@@ -86,7 +86,7 @@ public void RowConstructorTest()
             var pickledBytes = pickler.dumps(new[] { row1, row2 });
 
             // Note that the following will invoke RowConstructor.construct().
-            var unpickledData = PythonSerDe.GetUnpickledObjects(new MemoryStream(pickledBytes));
+            var unpickledData = PythonSerDe.GetUnpickledObjects(new MemoryStream(pickledBytes), pickledBytes.Length);
```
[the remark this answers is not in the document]
adamsitnik: I am sorry for not following the rules, my IDE did not produce any warnings and I have missed it.

#fixed

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
adamsitnik: I have changed the parameter name to "sizeHint" and simplified the comment.

The goal is to guess the max size in order to allocate the array only once.
The pickler is going to resize the buffer if needed (it's passed to the method by reference)

src/csharp/Microsoft.Spark/Interop/Ipc/SerDe.cs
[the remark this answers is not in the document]
stephentoub: > I would not expect a Try method to throw

Try* methods often throw.  The Try pattern isn't about eliminating all exceptions: it's explicitly about picking one category of exceptions that's likely to happen and turning just that one category into a Boolean result.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
stephentoub: @imback82, what situation are you optimizing for here?  Doesn't a stack of 100 suggest ~100 columns?  I'd expect this change would actually be a deoptimization for the common case where there are only a few columns.  Am I misunderstanding?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -30,14 +25,9 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
...
-            while (((length = SerDe.ReadBytesLength(inputStream)) != null) &&
-                (length.GetValueOrDefault() > 0))
+            while (((length = SerDe.ReadBytesLength(inputStream)) != null) && (length.Value > 0))
             {
-                MaxLengthReadStream readStream = s_slicedReadStream ??
-                    (s_slicedReadStream = new MaxLengthReadStream());
-
-                readStream.Reset(inputStream, length.GetValueOrDefault());
-                var unpickledObjects = PythonSerDe.GetUnpickledObjects(readStream);
+                object[] unpickledObjects = PythonSerDe.GetUnpickledObjects(inputStream, length.Value);
```
stephentoub: Why did you change these from GetValueOrDefault to Value?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
stephentoub: Ah, I was misremembering the implemetation.  Looks like the stack ends up being used to also hold the deserialized row results.  Ok.

adamsitnik: @stephentoub  The unpickler pushes every deserialized object to an internal stack. It is used to support tuples, list, arrays and more.

Let's consider the following example: we want to serialize a List (spark always excepts a list as a result, it throws for arrays) of arrays of two doubles each:

```
var toSerialize = new List<object>()
{
   new object[2] { 1.0, 2.0 },
   new object[2] { 3.0, 4.0 },
}
```

The pickling format represents it as:

```
Opcodes.EMPTY_LIST (byte marker)
Opcodes.MARK
Opcodes.BINFLOAT
1.0 (8 bytes Int64 BigEndianInteger)
Opcodes.BINFLOAT
2.0
Opcodes.TUPLE2
Opcodes.BINFLOAT
3.0
Opcodes.BINFLOAT
4.0
Opcodes.TUPLE2
Opcodes.APPENDS
```

Now when it deserializes the provided input:

- it pushes everything to the mentioned stack
- when it encounters TUPLE2 mark it pops the last two items and allocates an array of two elements and stores poped items in it. It also pushes the array to the stack.
- when it encounters APPENDS mark it pops everything from the stack until it hits MARK, stores it in a list, reverts it (it's a stack) and then returns the list


Adding the items to the stack and resizing it was visible in the profiles I collected after optimizing the way we read/write from/to an array so I have introduced this new parameter to provide a hint to the unpickler and save few % when deserializing.

stephentoub: Right, that's what I meant by "Looks like the stack ends up being used to also hold the deserialized row results." :wink:

stephentoub: Given that, isn't 100 generally too small?  When processing the last of 100 rows, won't we already be ~100 deep, and we'll grow the array to handle the processing of the last row?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
adamsitnik: Because the loop checks if it's not null, and when it's not null it is guaranteed to have a value.

```cs
while (((length = SerDe.ReadBytesLength(inputStream)) != null) && (length.Value > 0))
```

https://github.com/dotnet/spark/blob/35048c9cfa324336497055d6543fa46b8d2de247/src/csharp/Microsoft.Spark/Interop/Ipc/SerDe.cs#L215-L230

stephentoub: @adamsitnik, have you looked at the implementations of `Nullable<T>.Value` vs `Nullable<T>.GetValueOrDefault`?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
adamsitnik: > Right, that's what I meant by

I got an email with your first comment, started writing the response, posted it and then it turned out that in the meantime you wrote another one saying that you don't need the explanation anymore ;P

I hope that others will find it useful ;)

adamsitnik: > Given that, isn't 100 generally too small?

After setting it to 100 it got much smaller in the profiles. I was thinking about bigger number, but was not sure. Most benchmarks have batches of arrays of 100 items, some have of 100 items where every item is a tuple (so we would need 100).

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
adamsitnik: I did not and I did not know about the difference.  I am going to revert it.

BTW we should really write a Roslyn analyzer for this kind of micro-optimizations.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
stephentoub: > some have of 100 items where every item is a tuple (so we would need 100

With 100 rows of just one double per row, doesn't the pickling stack actually need to hold 101 elements?

And with tuples, when processing the last tuple, wouldn't the reading of the last element push all of the elements of the tuple onto the stack, such that we'd grow beyond 100?

I'm not convinced any of this really matters, but if we're going to play games with the initial size, we should make sure it's big enough for the common cases to avoid whatever overheads we're trying to avoid.  I would suggest running all of the various TPCH benchmarks that are relevant and keeping track of the high-water mark on the stack size.

adamsitnik: You are right. The stack size should be... 102 (I forgot about the markers). I have verified that it's the perfect number for Q1, Q13 and Q8.

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -30,14 +25,9 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
...
-            while (((length = SerDe.ReadBytesLength(inputStream)) != null) &&
-                (length.GetValueOrDefault() > 0))
+            while (((length = SerDe.ReadBytesLength(inputStream)) != null) && (length.GetValueOrDefault() > 0))
             {
-                MaxLengthReadStream readStream = s_slicedReadStream ??
-                    (s_slicedReadStream = new MaxLengthReadStream());
-
-                readStream.Reset(inputStream, length.GetValueOrDefault());
-                var unpickledObjects = PythonSerDe.GetUnpickledObjects(readStream);
+                object[] unpickledObjects = PythonSerDe.GetUnpickledObjects(inputStream, length.Value);
```
eerhardt: Should `.Value` be `.GetValueOrDefault()` instead?  (as discussed above)

src/csharp/Microsoft.Spark/Interop/Ipc/SerDe.cs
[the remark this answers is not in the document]
eerhardt: What was the resolution here? IMO it would be nice if we could just have a single pattern in the code.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -131,15 +125,17 @@ internal class PicklingSqlCommandExecutor : SqlCommandExecutor
...
+                    object[] inputRows = PythonSerDe.GetUnpickledObjects(inputStream, messageLength);
 
                     for (int i = 0; i < inputRows.Length; ++i)
                     {
                         // Split id is not used for SQL UDFs, so 0 is passed.
                         outputRows.Add(commandRunner.Run(0, inputRows[i]));
                     }
 
-                    WriteOutput(outputStream, outputRows);
+                    // most probably input > output. If not, pickler increases the buffer
```
imback82: To me, "most probably" is not true because it entirely depends on the UDFs. I would rewrite it to something like (and I think the code is clear enough without `sizeHint` variable):

```C#
// The initial (estimated) buffer size for pickling rows is set to the size of input pickled rows
// because the number of rows are the same for both input and output.
WriteOutput(outputStream, outputRows, messageLength)
```

src/csharp/Microsoft.Spark/Interop/Ipc/SerDe.cs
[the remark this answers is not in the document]
adamsitnik: I removed the method and reused `TryReadBytes`

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
adamsitnik: #done

adamsitnik:
@eerhardt @stephentoub @imback82 I have addressed all the code review comments, PTAL.

PR #113, opened 2019-05-20
Enable arcade support

eerhardt:
This is the first round of using dotnet/arcade as the build infrastructure.

It doesn't make use of everything in arcade, but it is the first step to using arcade, and we can incrementally make changes as we go to make use of more and more arcade functionality (for example the signtool which will sign nugets and their contained assemblies in a single step).

It is easiest to review by ignoring the first commit "Copy eng\common from Aracde repo". It is a strict copy of the `eng\common` folder that is defined by the arcade repo. It will get sync'd automatically when we start taking new versions of arcade using the dependency flow infrastructure.

This should solve #40.


eerhardt:
OK, this is now ready for review. Please take a look and let me know what you think.

azure-pipelines.yml
```diff
@@ -23,31 +22,15 @@ jobs:
   displayName: Build and Test Sources
   pool: Hosted VS2017
 
+  variables:
+    ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+      _OfficialBuildIdArgs: /p:OfficialBuildId=$(BUILD.BUILDNUMBER)
+
   steps:
   - task: DotNetCoreInstaller@0
```
safern: Do we still need to install the DotNetCoreInstaller? if you call build.cmd with the -ci flag, when installing the SDK stated on the global.json, it will add it to the path using VSO variables. So once you've restored it, you can call DotNetCoreCLI tasks without installing it manually.

eerhardt:
>     - task: DotNetCoreInstaller@0

I think this can now be removed on this leg. We no longer invoke DotNetCoreCLI tasks on this leg.

---
Refers to: azure-pipelines.yml:155 in 31c6cae. [](commit_id = 31c6cae334b6549e85a48220741027e75b131123, deletion_comment = False)

azure-pipelines.yml
```diff
@@ -64,6 +47,20 @@ jobs:
       filename: script\download-spark-distros.cmd
       arguments: $(Build.BinariesDirectory)
 
+  - script: build.cmd -pack
```
safern: Use the -ci flag, it cleans up stuff before and after building, plus adds `RepoRoot\.dotnet` to the path.

azure-pipelines.yml
```diff
@@ -64,6 +47,20 @@ jobs:
...
+              -c $(buildConfiguration)
+              $(_OfficialBuildIdArgs)
+              /p:PublishSparkWorker=true
+              /p:SparkWorkerPublishDir=$(Build.ArtifactStagingDirectory)\Microsoft.Spark.Worker
+    displayName: '.NET build'
+
+  - task: DotNetCoreCLI@2
+    displayName: '.NET unit tests'
+    inputs:
+      command: test
```
safern: Does `build.cmd -test /p:ProjectToBuild=**\*UnitTest\*.csproj` work?

eerhardt: No, unfortunately it doesn't. We can't run tests using arcade's test runner yet because it doesn't support coverlet, so we would lose code coverage. I started an internal discussion on the Arcade Working Group on this topic, but didn't get very far. Basically https://github.com/dotnet/arcade/issues/1946 is the blocking issue.

safern: Ok, cool. Once that gets resolved we can think of updating this.

eng/Versions.props
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <VersionPrefix>0.2.0</VersionPrefix>
+    <PreReleaseVersionLabel>prerelease</PreReleaseVersionLabel>
+    <RestoreSources>
+      $(RestoreSources);
+      https://api.nuget.org/v3/index.json;
+      https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;
```
safern: We're no longer publishing to myget. Please use:

`https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;`

eerhardt: This came from copying it from Directory.Build.props. I'm not sure if we need it at all. I'll try removing it.

azure-pipelines.yml
eerhardt: Nice! I didn't know this, and it worked out perfectly.

I verified that each of the test commands are executed with `.dotnet\dotnet.exe` from the repo:

```
[command]D:\a\1\s\.dotnet\dotnet.exe test D:\a\1\s\src\csharp\Microsoft.Spark.UnitTest\Microsoft.Spark.UnitTest.csproj --logger trx --results-directory D:\a\_temp --configuration Release /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura
```

azure-pipelines.yml
eerhardt: Done.

azure-pipelines.yml
```diff
@@ -235,36 +181,25 @@ jobs:
...
+          **/*.snupkg
         targetFolder: $(Build.ArtifactStagingDirectory)/Packages
-
-    - task: CopyFiles@2
-      displayName: Copy snupkg to publish
-      inputs:
-        sourceFolder: $(Build.ArtifactStagingDirectory)
-        contents: 'Microsoft.Spark*.snupkg'
-        targetFolder: $(Build.ArtifactStagingDirectory)/SymbolPackages
+        flattenFolders: true
```
safern: Just curious what does this do?

src/csharp/Microsoft.Spark.Worker.UnitTest/Microsoft.Spark.Worker.UnitTest.csproj
```diff
@@ -1,17 +1,14 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="coverlet.msbuild" Version="2.4.0">
+    <PackageReference Include="coverlet.msbuild" Version="2.6.1">
```
safern: Can this package reference be shared across UnitTest projects?

src/csharp/Microsoft.Spark.Worker/Microsoft.Spark.Worker.csproj
```diff
@@ -5,9 +5,13 @@
     <TargetFrameworks>net461;netcoreapp2.1</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.1</TargetFrameworks>
     <RootNamespace>Microsoft.Spark.Worker</RootNamespace>
-    <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(MSBuildProjectName).xml</DocumentationFile>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.Spark.Worker.UnitTest" />
```
safern: Nice, I didn't know this was possible. Would've saved me a lot of pain when I did the signing work.

src/csharp/Microsoft.Spark.E2ETest/Microsoft.Spark.E2ETest.csproj
```diff
@@ -5,13 +5,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="coverlet.msbuild" Version="2.4.0">
+    <PackageReference Include="coverlet.msbuild" Version="2.6.1">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
```
safern: Why can we remove these now?

eerhardt: The need to be because they are automatically included by arcade when the project is a "test" project. If I don't remove them I get errors saying it was referenced twice. (once implicitly)

azure-pipelines.yml
eerhardt: It puts the files directly in the root folder (i.e. flattened) instead of having the same directory structure underneath.

So if you copy:

```
\a
   \b
        foo.exe
```

to a folder `\temp`, you would get `\temp\b\foo.exe`.  But here I want the `foo.exe` to be directly under `\temp`, and not to preserve the folder hierarchy.

src/csharp/Microsoft.Spark.E2ETest/Microsoft.Spark.E2ETest.csproj
safern: Ah I see, that is why I couldn’t find where these references were defined. (Arcade magic)

azure-pipelines.yml
safern: Cool. Good to know, thanks for explaining!

src/csharp/Microsoft.Spark.E2ETest/SparkFixture.cs
```diff
@@ -127,7 +127,7 @@ private void BuildSparkCmd(out string filename, out string args)
             string classArg = "--class org.apache.spark.deploy.DotnetRunner";
             string curDir = AppDomain.CurrentDomain.BaseDirectory;
             string jarPrefix = GetJarPrefix(sparkHome);
-            string scalaDir = $"{curDir}{sep}..{sep}..{sep}..{sep}..{sep}..{sep}scala";
+            string scalaDir = $"{curDir}{sep}..{sep}..{sep}..{sep}..{sep}..{sep}src{sep}scala";
```
imback82: did we change Scala directory structure? This also runs locally right?

imback82:
> It is easiest to review by ignoring the first commit "Copy eng\common from Aracde repo". It is a strict copy of the `eng\common` folder that is defined by the arcade repo. It will get sync'd automatically when we start taking new versions of arcade using the dependency flow infrastructure.

So do we plan to remove files under eng\common?

imback82:
I see this error message in https://dev.azure.com/dnceng/public/_build/results?buildId=198008, but the build task is marked as success. Is this expected?
```
D:\a\1\s\.packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.19257.7\tools\SymStore.targets(61,5): error MSB3073: The command ""D:\a\1\s\.packages\microsoft.diasymreader.pdb2pdb\1.1.0-beta1-62506-02\tools\Pdb2Pdb.exe" "D:\a\1\s\artifacts\bin\Microsoft.Spark.FSharp.Examples\Release\net461\Microsoft.Spark.FSharp.Examples.exe" /out "D:\a\1\s\artifacts\SymStore\Release\Microsoft.Spark.FSharp.Examples\net461\Microsoft.Spark.FSharp.Examples.pdb" /verbose /srcsvrvar SRC_INDEX=public" exited with code 2. [D:\a\1\s\examples\Microsoft.Spark.FSharp.Examples\Microsoft.Spark.FSharp.Examples.fsproj]
```

src/csharp/Microsoft.Spark.Worker.UnitTest/Microsoft.Spark.Worker.UnitTest.csproj
eerhardt: Yes. Fixed.

src/csharp/Microsoft.Spark.E2ETest/SparkFixture.cs
eerhardt: No, the Scala directory structure didn't change.

What DID change is where all the .NET outputs go.  Before the output directory used to be in a folder underneath the `$(RepoRoot)\src\csharp` folders. Now all the outputs of the .NET projects go into the `$(RepoRoot)\artifacts` folder. So that's why I needed to add `src` to this path, because `artifacts` and `src` are at the same level.

Yes, this also runs locally.

eerhardt:
> So do we plan to remove files under eng\common? 

No, these stay in the repo. But you don't modify them here- the "master" version is located in https://github.com/dotnet/arcade/tree/master/eng/common. When we take new versions of arcade, these files are updated automatically.

> Also, do you have an internal build to share so I can double check the nuget and worker binaries?

https://dev.azure.com/dnceng/internal/_build/results?buildId=198161

> I see this error message in https://dev.azure.com/dnceng/public/_build/results?buildId=198008, but the build task is marked as success. Is this expected?

No, looking into it now.

src/csharp/Microsoft.Spark.E2ETest/SparkFixture.cs
imback82: Got it, thanks. Does this mean we need to update docs that depend on the output path? (https://github.com/dotnet/spark/blob/master/docs/building/windows-instructions.md#run-samples)

eerhardt: Good catch! Will update.

eerhardt:
OK, all feedback is now addressed, and I've updated the branch to the latest code. This should be ready for a final review.

> I see this error message in https://dev.azure.com/dnceng/public/_build/results?buildId=198008, but the build task is marked as success. Is this expected?

The issue here is caused by an F# bug: https://github.com/dotnet/fsharp/issues/6832. But also, our `Examples` projects were defaulting to `IsShipping=true` in the arcade SDK, which was causing some extra processing during the build. Setting `IsShipping=false` on the projects that aren't shipping opts out of the processing that was causing the issue. And since we only have F# examples, and not shipping binaries, we are able to work around the issue.

imback82:
> > Also, do you have an internal build to share so I can double check the nuget and worker binaries?
> 
> https://dev.azure.com/dnceng/internal/_build/results?buildId=198161

I checked the nupkg and worker binaries and they are appended with "-prerelease.19272.6". What's the recommended way of releasing from now on? We were planning to release without the prerelease tag.

azure-pipelines.yml
```diff
@@ -134,19 +127,12 @@ jobs:
...
       inputs:
-        sourceFolder: $(Build.SourcesDirectory)/src/csharp/Microsoft.Spark/bin/$(buildConfiguration)
+        sourceFolder: $(Build.SourcesDirectory)/artifacts/packages/$(buildConfiguration)/Shipping
         contents: |
-          **/*.dll
-          **/*.pdb
-        targetFolder: $(Build.ArtifactStagingDirectory)/BuildArtifacts/src/csharp/Microsoft.Spark/bin/$(buildConfiguration)
+          **/*.nupkg
+          **/*.snupkg
+        targetFolder: $(Build.ArtifactStagingDirectory)/BuildArtifacts/artifacts/packages/$(buildConfiguration)/Shipping
```
safern: If the target folder is `$(Build.ArtifactStagingDirectory)/BuildArtifacts` and we make the `contents` be: `**/Shipping/**/*.nupkg` and `**/Shipping/**/*.snupkg` and then the sourceFolder being `(Build.SourcesDirectory)` it should preserve the folder structure, right?

eerhardt:
> What's the recommended way of releasing from now on? We were planning to release without the prerelease tag.

Daily builds get a build number appended to the end. That’s how you know it isn’t an officially released NuGet package - it has a daily build number. We can push those daily builds to feeds that aren’t NuGet.org. When you are ready to ship, you queue a build with a variable set to build without the build number at the end. And then you release that package.

azure-pipelines.yml
eerhardt: Yes I think so. I can try changing it that way tomorrow morning.

safern: Not required to merge but it’s a nice cleanup.

PR #114, opened 2019-05-21
Fix a bug that produces an incorrect JSON for UDF return types.

### LINKED ISSUE #112: Array/Map return types for UDF do not work correctly.
I have a scenario where for structured streaming input and for each event/row i have to write a custom logic/function which can return multiple rows.

looks like for return type UDF only supports basic type and not list/array.

Any workaround for this?

for sample my UDF is something like below, so that i can explode to create multiple rows.
`
         Func<Column, Column> ToUpperList = Udf<string, string[]>((arg) =>
            {
                var ret = new string[] { arg, arg.ToUpper()};
                return ret;
            });

            var query = inStream
                .Select(Explode( ToUpperList(Col("InputEventColumn"))))
                .WriteStream()
                .OutputMode("append")
                .Format("console")
                .Start();`

imback82:
Fixes #112.


src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
```diff
@@ -119,16 +119,38 @@ public sealed class ShortType : IntegralType
     }
 
     /// <summary>
-    /// Represents a decimal type (not implemented).
+    /// Represents a decimal type.
     /// </summary>
     public sealed class DecimalType : FractionalType
     {
+        internal static Regex s_fixedDecimal =
+            new Regex(@"decimal\s*\((\d+),\s*(\d+)\)", RegexOptions.Compiled);
```
suhsteve: Are the \s* required ?

PySpark's spark/python/pyspark/sql/types.py:
```python
    def jsonValue(self):
        return "decimal(%d,%d)" % (self.precision, self.scale)
```

And the below:
```csharp
        public override string SimpleString => $"decimal({_precision},{_scale})";
```

doesn't contain spaces.

imback82: Yes. The true source should be https://github.com/apache/spark/blob/1e87694f2bc0dea3eaaef936dc8552b3ba421539/sql/catalyst/src/main/scala/org/apache/spark/sql/types/DataType.scala#L117

suhsteve: Let's add the missing \s* after the (\d+) capture groups.

src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
```diff
@@ -49,47 +82,50 @@ internal enum PythonEvalType
...
-            // See spark.sql.datatypes.datatype.scala for reference.
-            var types = new Type[] { type };
-            var returnTypeFormat = "{0}";
-
-            if (type.IsArray)
+            if (s_returnTypes.TryGetValue(type, out string value))
             {
-                types[0] = type.GetElementType();
-                returnTypeFormat = "array<{0}>";
+                return $"\"{value}\"";
```
suhsteve:
```suggestion
                return $@"""{value}""";
```

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
imback82: Sure. Let me change that.

src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
imback82: fixed.

suhsteve:
LGTM

PR #124, opened 2019-05-29
Use arcade's signing tool to sign all artifacts.

### LINKED ISSUE #119: Microsoft.Spark.dll is not signed
It stopped being signed when we moved from nuspec into using `dotnet pack`. The reason of it is because apparently the --no-build switch is being ignored by the DotNetCoreCLI task when command = pack, so it rebuilds the binaries and overrides the signed ones with non-signed. I tried changing it to use the custom command and with that respects it. However, that is not enough, since we restored and build the binaries in another agent. So we would need to run restore again before running `dotnet pack --no-build`. 

@eerhardt has a PR open to use `arcade` as part of the infrastructure, which has built-in support for signing a nupkg, and its contents. You give it the `nupkg`, it will open it, sign its contents, re-package it and sign the `nupkg`. So I don't think it is worth fixing it now, until that PR is merged. 

Once that is merged, we can produce the `nupkg` in the first build step. Then on the signing step, we download it, and just give arcade's signtool the nupkg to sign. That will simplify our build and avoid this kind of issues.

cc: @imback82

eerhardt:
This completely redoes the way we do signing to use the signtool from arcade.

All the signing happens in a single step now, it unpacks the `.nupkg`, extracts the files that need signing, signs them, puts them back in the `.nupkg` and signs the `.nupkg`.

It also de-dupes assemblies, so duplicate assemblies are only signed a single time. ex. We only need to sign `Microsoft.Spark.dll` one time, but it is placed in all the places that use it.

Fix #119 

/cc @tmat - in case you are interested in reviewing. But feel free to skip.


azure-pipelines.yml
```diff
@@ -147,6 +147,10 @@ jobs:
       name: NetCoreInternal-Int-Pool
       queue: buildpool.windows.10.amd64.vs2017
 
+    variables:
+      ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+        _OfficialBuildIdArgs: /p:OfficialBuildId=$(BUILD.BUILDNUMBER)
```
eerhardt: I don't know how to define this variable in a single spot. I don't really speak yaml. So if anyone has a suggestion on how define this variable once in the yaml file and reuse it across both jobs, I'm all ears.

eng/Signing.props
```diff
@@ -0,0 +1,15 @@
+<Project>
+  <ItemGroup>
+    <ItemsToSign Include="$(SparkPackagesDir)/**/*.nupkg" />
+    <ItemsToSign Include="$(SparkPackagesDir)/**/*.snupkg" />
+    
+    <ItemsToSign Include="$(SparkWorkerPublishDir)/**/Microsoft.Spark.dll" />
+    <ItemsToSign Include="$(SparkWorkerPublishDir)/**/Microsoft.Spark.Worker.exe" />
+    <ItemsToSign Include="$(SparkWorkerPublishDir)/**/Microsoft.Spark.Worker.dll" />
```
imback82: For Linux, the assembly can be just `Microsoft.Spark.Worker`. Do we need to sign it? (I vaguely remember signing only for Windows assemblies).

eerhardt: We don't need to sign the assemblies on Linux, since there is no signing validation. Also - typically Linux binaries are built on a Linux machine, where signing is not possible.

From my understanding, the things that need to be signed are:

- Binaries executed on Windows machines, along with any Windows installer files, i.e. msi.
- nupkg files. 
- vsix files, but we don't make any of those.
- macOS installer files, i.e. pkg. but we don't make any of those.

azure-pipelines.yml
safern: Can't you define it in the global variables section?

eng/Signing.props
```diff
@@ -0,0 +1,15 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <!-- extend arcade's FileExtensionSignInfo with snupkg information -->
+    <FileExtensionSignInfo Include=".snupkg" CertificateName="NuGet" />
```
safern: Do we need to specify the certificate name for .jar files? or is that already built-in into arcade's signtool?

azure-pipelines.yml
eerhardt: I did that, and it failed to start the build. I'm sure I did something wrong, but I assume it had something to do with the `    ${{ if ` at the beginning. Do those always need to go last in the list?

eng/Signing.props
eerhardt: Already, built in:

https://github.com/dotnet/arcade/blob/8052d0f693173d9bb99f6fa10611a6f4aafd4eb0/src/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj#L34

eng/Signing.props
```diff
@@ -0,0 +1,15 @@
+<Project>
+  <ItemGroup>
+    <ItemsToSign Include="$(SparkPackagesDir)/**/*.nupkg" />
```
safern: Can we default these two `Dir` variables to a default location in the repo? I think there is a `PackagesDir` property defined in arcade for the shipping packages.

eng/Signing.props
safern: Great!

azure-pipelines.yml
safern: That is what I thought, I don't think you can globally defined variables with a condition, I don't think it is allowed because the global section is expanded before evaluating the ${{ expressions at compile time.

safern: We could technically use a variable template, but I don't think it is worth to just avoid duplicating 1 variable.

eng/Signing.props
eerhardt: > Can we default these two Dir variables to a default location in the repo? 

We could, but I think that would just make the build more complicated than it is now. 

Today, we:
1. build the nupkg and publish the worker in the first leg, then put them in an artifact.
2. On the second leg, we download the artifacts into the Staging directory. Sign everything and package the worker.
3. Publish the signed and packaged assets to the artifact.

If we put these two `Dir` variables into the repo, we would need to copy files around before and after (2) above.

> I think there is a PackagesDir property defined in arcade for the shipping packages.

Arcade has `$(ArtifactsPackagesDir)`

https://github.com/dotnet/arcade/blob/8052d0f693173d9bb99f6fa10611a6f4aafd4eb0/src/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj#L17

But using that would mean the same copying files above, so I took the easier path of defining two new properties just for the signing operation. And the SparkWorkerPublishDir gets reused below when packaging the worker.

azure-pipelines.yml
```diff
@@ -183,7 +183,8 @@ jobs:
     - task: DotNetCoreCLI@2
       displayName: Package Microsoft.Spark.Worker
       inputs:
-        command: msbuild
+        command: custom
+        custom: msbuild
         projects: eng\PackageWorker.proj
```
tmat: Why does this not build in the previous step?

eerhardt: I didn't fold packaging the worker into build/pack phase because I didn't see sign tool having support for `.tar.gz` files.  `PackageWorker.proj` builds 3 archives:

- 2 `.zip` files for Windows.
  - netfx
  - netcoreapp
- 1 `.tar.gz` file for Linux netcoreapp

Granted, it probably isn't critical that the managed assemblies in the `.tar.gz` file aren't signed. But it "feels wrong" to ship unsigned managed binaries, even if they are only going to be used on non-Windows.

Thoughts? Do you think we could ship unsigned binaries in the `.tar.gz` files? Or maybe we can add support in arcade for `.tar.gz` files?

As for keeping the packaging happening after signing, but doing it in the same build step, I guess I didn't know how to do that "the arcade way". How do I say:

`eng\common\build.ps1 -restore -sign --??` - what goes in `--??` to build an extra project?

Can I do:

`eng\common\build.ps1 -restore -sign -projects eng\PackageWorker.proj`

? Will that call signing first, and then build the `eng\PackageWorker.proj`? I thought it would be the other way around - build the project, then sign. Which is not what we want.

https://github.com/dotnet/arcade/blob/b5016f5688dc8ca9f3e4811ee7e2e86ad8907a40/src/Microsoft.DotNet.Arcade.Sdk/tools/Build.proj#L215-L243

I guess, looking at that code now, I see the `AfterSigning.proj`. I could hook into that to do the `PackageWorker`. Thoughts?

tmat: Could you file an issue on Arcade to enable signing artifacts in `.tar.gz`? I don't know if there is a good .NET API that allows us to re-package `.tar.gz` but we should look into it.

eerhardt:
Ok, I got a good build https://dev.azure.com/dnceng/internal/_build/results?buildId=205030. I verified that all the artifacts are signed correctly. And I've addressed all feedback.

Please take a look and provide any additional feedback.

azure-pipelines.yml
eerhardt: Done - https://github.com/dotnet/arcade/issues/2915.

eng/Signing.props
imback82: Got it, thanks.

azure-pipelines.yml
```diff
@@ -147,6 +147,10 @@ jobs:
       name: NetCoreInternal-Int-Pool
```
safern: Small Nit: This job no longer builds the packages, should we change the display name to Sign Artifacts?

eerhardt: Sorry - I merged before I saw this. I will do this next time I'm in here.

safern: No problem. I approved before realizing the name was off 😄

PR #125, opened 2019-05-30
Fix VS 2017 build

eerhardt:
The VS 2017 F# compiler doesn't like the new AssemblyInformationAttribute added by dotnet/arcade.

Also, rename the signing leg of the build.

### DIFF
--- a/examples/Microsoft.Spark.FSharp.Examples/Microsoft.Spark.FSharp.Examples.fsproj
+++ b/examples/Microsoft.Spark.FSharp.Examples/Microsoft.Spark.FSharp.Examples.fsproj
@@ -7,6 +7,8 @@
     <RootNamespace>Microsoft.Spark.Examples</RootNamespace>
     <AssemblyName>Microsoft.Spark.FSharp.Examples</AssemblyName>
     <PublicSign>false</PublicSign>
+    <!-- workaround https://github.com/dotnet/fsharp/issues/4822 to support VS 2017 -->
+    <NoWarn>2003;$(NoWarn)</NoWarn>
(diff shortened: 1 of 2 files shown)


examples/Microsoft.Spark.FSharp.Examples/Microsoft.Spark.FSharp.Examples.fsproj
```diff
@@ -7,6 +7,8 @@
     <RootNamespace>Microsoft.Spark.Examples</RootNamespace>
     <AssemblyName>Microsoft.Spark.FSharp.Examples</AssemblyName>
     <PublicSign>false</PublicSign>
+    <!-- workaround https://github.com/dotnet/fsharp/issues/4822 to support VS 2017 -->
```
imback82: Looks like if we move to vs 2019, this will be fixed right? If so, do we have buildpool for vs2019?

eerhardt: In the official builds (and building from the command-line using `build.cmd|sh`) we always use the exact same version of the .NET Core SDK (the one that is in the root global.json file). So it doesn't matter what machine you are running on.

When building inside of VS 2019, there are no issues. I was doing this on my main dev machine while doing the arcade work.

However, moving to my other dev box, which only has VS 2017, I was getting a build error from the F# project. Suppressing this warning allows the solution to build successfully. It is just an example project, so I can't imagine any down sides to suppressing it.

My intention here is that devs working in this repo shouldn't be forced to install VS 2019.

Issue #45, opened 2019-04-24
Arrow enabled implementation and comparison with pySpark


voltcode:
It seems that using Arrow enabled Pandas could greatly improve pySpark efficiency. Did you compare against Arrow enabled Pandars or the version without? Is this detail something you could highlight in the comparison? As far as I know ,NET does not have proper Arrow bindings, and it would be of great benefit if it did, did you consider creating it? There's a bit of a discussion in https://github.com/dotnet/machinelearning/issues/69 if you need more context.

rapoth:
@voltcode: Thank you for your comment! We totally agree and we are working on more benchmarks as I type this :) For a comparison against Arrow enabled Pandas in Python, having efficient .NET support for Arrow is a must. 

We are already working very closely with the Arrow community towards getting the initial version done over the next few weeks. @eerhardt from our team has been contributing several PRs lately to [Apache Arrow](https://github.com/apache/arrow): https://github.com/apache/arrow/pull/4017,  https://github.com/apache/arrow/pull/5019, https://github.com/apache/arrow/pull/4839, https://github.com/apache/arrow/pull/4502, https://github.com/apache/arrow/pull/4737, https://github.com/apache/arrow/pull/4543, https://github.com/apache/arrow/pull/4435

Our plan is that once we have a reasonable Arrow .NET integration, we will run more benchmarks. This is a great point that you made. We are very open to hearing your take on any configurations we should pay special attention to. Since benchmarking is a sensitive topic, we want to be totally transparent about what we are running and how we are running - you can find all the benchmarking code we are using [here](/benchmark). Our goal is to have community members guide us towards getting sensible benchmarks.

eerhardt:
@voltcode - Arrow enabled Pandas UDFs are something we are actively working on. The preliminary numbers are promising when there are lots of records being serialized between the JVM and .NET, and we are working on making them better.

> As far as I know ,NET does not have proper Arrow bindings, and it would be of great benefit if it did, did you consider creating it? 

[...]

PR #127, opened 2019-05-31
Implement VectorUdf and use it in Queries 1 and 8 of TPCH benchmarks.

eerhardt:
As requested in #45, this brings Vector UDF support to .NET for Apache Spark.

See https://databricks.com/blog/2017/10/30/introducing-vectorized-udfs-for-pyspark.html for the equivalent Python capabilities. This first round of implementation supports `PandasUDFType.SCALAR` type of UDFs, but grouped UDFs can be implemented in the future.

On an Azure DS5 Ubuntu 18 box, I'm seeing the following results in these two TPCH queries using the `1 GB` dataset and a single core `--master local`:

|                      Query|      One-at-a-time|    Vectorized|
|---------------------------- |----------:|----------:|
| Query 1 | 10.3 s | 8.3 s |
| Query 8 | 30.7 s | 29.1 s |

## Notes
1. I created a new "Experimental" project because exposing VectorUdf using the Arrow types directly is not the long term plan. The hope is that there will be a fully featured `DataFrame` type in .NET that can be created from an Arrow RecordBatch, with usable APIs. VectorUdf will be considered an "experimental" API until we have a stable long-term plan for how to pass columnar data to a .NET method.
2. For other TPCH queries that have the same logic to compute the discount, using VectorUdf gave me worse performance numbers. My assumption is that the other queries aren't dealing with as much data, and converting to a columnar format (Arrow) is outweighing the benefits of vectorization.
3. I could have used the new C# hardware intrinsics APIs in .NET Core 3.0 to perform the vectorization, but that would either force everyone to use a preview 3.0 SDK, or I would have to perform unnatural build gymnastics to allow building without the 3.0 SDK. So instead, I chose to use `Vector<T>`, which still uses vectorization under the covers and works on the current targeted frameworks.


benchmark/csharp/Tpch/VectorizedFunctions.cs
```diff
@@ -0,0 +1,121 @@
...
+                    Vector<double> r = exP * ((ones - dis) * (ones + t));
+
+                    for (int i = 0; i < Vector<double>.Count; i++)
+                    {
+                        builder.Append(r[i]);
+                    }
+
+                    pExtendedPrice += 4;
+                    pDiscount += 4;
+                    pTax += 4;
```
stephentoub: Should these instead be `+= Vector<double>.Count`?

benchmark/csharp/Tpch/VectorizedFunctions.cs
```diff
@@ -0,0 +1,121 @@
...
+            fixed (double* pStartExtendedPrice = price.Values)
+            fixed (double* pStartDiscount = discount.Values)
+            fixed (double* pStartTax = tax.Values)
+            {
+                double* pExtendedPrice = pStartExtendedPrice;
+                double* pDiscount = pStartDiscount;
+                double* pTax = pStartTax;
+                double* pExtendedPriceEnd = pExtendedPrice + length;
+
+                double* pVectorizationEnd = pExtendedPriceEnd - 4;
```
stephentoub: `- Vector<double>.Count`?

benchmark/csharp/Tpch/VectorizedFunctions.cs
```diff
@@ -0,0 +1,121 @@
...
+            Vector<double> ones = new Vector<double>(1d);
+
+            fixed (double* pStartExtendedPrice = price.Values)
+            fixed (double* pStartDiscount = discount.Values)
+            {
+                double* pExtendedPrice = pStartExtendedPrice;
+                double* pDiscount = pStartDiscount;
+                double* pExtendedPriceEnd = pExtendedPrice + length;
+
+                double* pVectorizationEnd = pExtendedPriceEnd - 4;
```
stephentoub: Same comment on all these hard-coded 4s.  Will stop commenting on them.

src/csharp/Microsoft.Spark.Experimental/Sql/ExperimentalFunctions.cs
```diff
@@ -0,0 +1,256 @@
...
+    public static class ExperimentalFunctions
+    {
+        /// <summary>Creates a Vector UDF from the specified delegate.</summary>
+        /// <typeparam name="A1">Specifies the type of the first argument to the UDF.</typeparam>
+        /// <typeparam name="RT">Specifies the return type of the UDF.</typeparam>
+        /// <param name="udf">The Vector UDF function implementation.</param>
+        /// <returns>
+        /// A delegate that returns a <see cref="Column"/> for the result of the Vector UDF.
+        /// </returns>
+        public static Func<Column, Column> VectorUdf<A1, RT>(Func<A1, RT> udf)
```
stephentoub: Nit: are the names `A1` and `RT` similar to what we use elsewhere in Spark?  They're not exactly descriptive :)

src/csharp/Microsoft.Spark/Sql/ArrowArrayHelpers.cs
```diff
@@ -14,720 +13,100 @@ namespace Microsoft.Spark.Sql
...
-
-        private static IArrowArray ToStringArrowArray(string[] array)
-        {
-            var valueOffsets = new ArrowBuffer.Builder<int>();
-            var valueBuffer = new ArrowBuffer.Builder<byte>();
-            var offset = 0;
-
-            // TODO: Use array pool and encode directly into the array.
-            foreach (string str in array)
+            if (typeof(T) == typeof(Int8Array))
```
stephentoub: Can you make all of these `else if`s?

stephentoub:
Nice.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -51,60 +48,100 @@ public void TestCollect()
...
-                SetUseArrowValue(originalUseArrow);
+        }
+
+        [Fact]
+        public void TestVectorUdf()
+        {
+            // Single UDF.
+            Func<Column, Column, Column> udf1 = VectorUdf<Int32Array, StringArray, StringArray>(
+                (ages, names) => (StringArray)ToArrowArray(
+                        Enumerable.Range(0, names.Length)
```
imback82: alignment is off by 4. (Line 116 looks correct)

benchmark/csharp/Tpch/VectorizedFunctions.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Runtime.CompilerServices;
+using Apache.Arrow;
+
+namespace Tpch
+{
+    internal static class VectorizedFunctions
+    {
+        internal unsafe static DoubleArray ComputeTotal(DoubleArray price, DoubleArray discount, DoubleArray tax)
+        {
+            if (price.Length != discount.Length || price.Length != tax.Length)
```
imback82:
```C#
if ((price.Length != discount.Length) || (price.Length != tax.Length))
{
    throw new ArgumentException("Arrays need to be the same length");
}
```

benchmark/csharp/Tpch/VectorizedFunctions.cs
```diff
@@ -0,0 +1,121 @@
...
+                throw new ArgumentException("Arrays need to be the same length");
+
+            int length = price.Length;
+            var builder = new ArrowBuffer.Builder<double>(length);
+            Vector<double> ones = new Vector<double>(1d);
+
+            fixed (double* pStartExtendedPrice = price.Values)
```
imback82: Will this run a lot faster than using `GetValue()`? I am thinking whether we should write this in a simpler way for the first version.

benchmark/csharp/Tpch/VectorizedFunctions.cs
```diff
@@ -0,0 +1,121 @@
...
+    internal static class VectorizedFunctions
+    {
+        internal unsafe static DoubleArray ComputeTotal(DoubleArray price, DoubleArray discount, DoubleArray tax)
+        {
+            if (price.Length != discount.Length || price.Length != tax.Length)
+                throw new ArgumentException("Arrays need to be the same length");
+
+            int length = price.Length;
+            var builder = new ArrowBuffer.Builder<double>(length);
+            Vector<double> ones = new Vector<double>(1d);
```
imback82: `var`

benchmark/csharp/Tpch/VectorizedFunctions.cs
```diff
@@ -0,0 +1,121 @@
...
+                    pExtendedPrice += 4;
+                    pDiscount += 4;
+                    pTax += 4;
+                }
+
+                while (pExtendedPrice <= pExtendedPriceEnd)
+                {
+                    builder.Append(*pExtendedPrice * (1 - *pDiscount) * (1 + *pTax));
+
+                    pExtendedPrice++;
```
imback82: Please use prefix increment operator throughout the files changed: https://github.com/dotnet/spark/blob/master/docs/coding-guidelines/csharp-coding-style.md.

src/csharp/Microsoft.Spark.UnitTest/UdfWrapperTests.cs
```diff
@@ -136,107 +138,116 @@ private void ValidatePicklingWrapper(int numArgs, dynamic udfWrapper)
...
+                StringArray, StringArray, StringArray, StringArray, StringArray, StringArray, StringArray, StringArray, StringArray, StringArray, StringArray>(
                     (str1, str2, str3, str4, str5, str6, str7, str8, str9, str10)
-                        => str1 + str2 + str3 + str4 + str5 + str6 + str7 + str8 + str9 + str10);
+                        => Concat(str1, str2, str3, str4, str5, str6, str7, str8, str9, str10));
             ValidateArrowWrapper(10, udfWrapper);
         }
 
+        private static StringArray Concat(params StringArray[] arrays)
+        {
+            StringBuilder builder = new StringBuilder();
```
imback82: `var`

src/csharp/Microsoft.Spark.UnitTest/UdfWrapperTests.cs
```diff
@@ -136,107 +138,116 @@ private void ValidatePicklingWrapper(int numArgs, dynamic udfWrapper)
...
+            int length = arrays[0].Length;
+            string[] resultStrings = new string[length];
```
imback82: `var`

src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
```diff
@@ -151,172 +164,214 @@ internal static JvmObjectReference CreatePythonFunction(IJvmBridge jvm, byte[] c
...
+            return (PicklingDelegate)
+                new PicklingUdfWrapper<
+                    T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, TResult>(udf).Execute;
+        }
+
+        internal static Delegate CreateVectorUdfWrapper<T1, TResult>(Func<T1, TResult> udf)
+            where T1 : IArrowArray
+            where TResult : IArrowArray
+        {
+            return (ArrowDelegate)
```
imback82: nit: Can we have the same style as `CreateUdfWrapper` for the first few?

src/csharp/Microsoft.Spark/Sql/UDFRegistration.cs
```diff
@@ -254,7 +254,7 @@ private void Register<TResult>(string name, Delegate func)
                     name,
                     pythonFunction,
                     GetDataType<TResult>(),
-                    (int)UdfUtils.GetPythonEvalType(),
+                    (int)UdfUtils.PythonEvalType.SQL_BATCHED_UDF,
```
imback82: I think we should allow registering VectorUdf through this API as well (this API is registering UDFs so that they can be invoked from SQL statements).

I will check PySpark and update this thread.

benchmark/csharp/Tpch/VectorizedFunctions.cs
eerhardt: Doh. I copied this code from some 3.0 hardware intrinsics code and forgot to update these constants. Thanks for the catch.

benchmark/csharp/Tpch/VectorizedFunctions.cs
eerhardt: It actually runs end-to-end in the exact same time whether I use a vectorized implementation, or just add them up one row at a time. I've tried it multiple times and I get the same execution time for query `1v`.

I'll change this to a much more simple approach:

```C#
        internal static DoubleArray ComputeTotal(DoubleArray price, DoubleArray discount, DoubleArray tax)
        {
            if ((price.Length != discount.Length) || (price.Length != tax.Length))
            {
                throw new ArgumentException("Arrays need to be the same length");
            }

            int length = price.Length;
            var builder = new ArrowBuffer.Builder<double>(length);
            ReadOnlySpan<double> prices = price.Values;
            ReadOnlySpan<double> discounts = discount.Values;
            ReadOnlySpan<double> taxes = tax.Values;
            for (int i = 0; i < length; i++)
            {
                builder.Append(prices[i] * (1 - discounts[i]) * (1 + taxes[i]));
            }

            return new DoubleArray(
                builder.Build(),
                nullBitmapBuffer: ArrowBuffer.Empty,
                length: length,
                nullCount: 0,
                offset: 0);
        }
```

eerhardt: UPDATE:

I ran query `8v` and I am seeing a slight change in perf:

|                      Query|      Normal UDF|    VectorUdf with `Vector<T>`| VectorUdf with simple `for`|
|---------------------------- |----------:|----------:|----------:|
| Query 8 | 29.7 s | 28.6 s | 29.0 s |

imback82: I actually like the simple approach since that is how people would normally write (I think).

We can revisit this when we start benchmarking TPCH for the cluster run (along with hardware intrinsics). What do you think?

eerhardt: That sounds like a good plan to me.

benchmark/csharp/Tpch/VectorizedFunctions.cs
eerhardt: These are now all gone with resolving https://github.com/dotnet/spark/pull/127#discussion_r289927023

src/csharp/Microsoft.Spark.Experimental/Sql/ExperimentalFunctions.cs
eerhardt: I copied these from https://github.com/dotnet/spark/blob/eb26baa46200bfcbe3e1080e650f335853d9990e/src/csharp/Microsoft.Spark/Sql/Functions.cs#L3604.

@imback82 - was there a specific reason to use `A1` there? Should I update all these to `T1`?

src/csharp/Microsoft.Spark/Sql/ArrowArrayHelpers.cs
```diff
@@ -14,720 +13,100 @@ namespace Microsoft.Spark.Sql
...
-                new[] { ArrowBuffer.Empty, builder.Build() });
-
-            return ArrowArrayFactory.BuildArray(data);
-        }
-
-        private static byte[] CreateRawBytesForBoolArray(int boolArrayLength)
-        {
-            int byteLength = boolArrayLength / 8;
-            if (boolArrayLength % 8 != 0)
+            if (typeof(T) == typeof(BooleanArray))
```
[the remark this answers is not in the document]
eerhardt: No they aren't structs, they are classes. https://github.com/apache/arrow/blob/cdedd85f9c8545ab3c98e7b9de938a28d1ec9e11/csharp/src/Apache.Arrow/Arrays/BooleanArray.cs#L20

But I was not aware of the limitation. Why wouldn't the JIT be able to see `if (typeof(BooleanArray) == typeof(BooleanArray))` is always `true` when JIT'ing the generic method? Similarly, when `if (typeof(Int32Array) == typeof(BooleanArray))` is always `false` and prune the branch?

src/csharp/Microsoft.Spark/Sql/ArrowArrayHelpers.cs
eerhardt: Yes, will do.

src/csharp/Microsoft.Spark.Experimental/Sql/ExperimentalFunctions.cs
imback82: It should be from https://github.com/apache/spark/blob/8486680b34a02af5a6975ae8fc6e0b465e4f4536/sql/core/src/main/scala/org/apache/spark/sql/functions.scala#L3961

imback82: I think we can follow `Func<>`'s type parameter names if needed.

eerhardt: So would you rather continuing using `A1`, `A2`, etc for the type names like scala? Or should I use `T1`, `T2`, like `Func<>`?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
eerhardt: Fixed. There were a few other places that were off as well.

src/csharp/Microsoft.Spark.Experimental/Sql/ExperimentalFunctions.cs
eerhardt: I updated both to use `T1`, `T2`, etc.

src/csharp/Microsoft.Spark/Sql/UDFRegistration.cs
eerhardt: I looked through the PySpark docs - https://spark.apache.org/docs/latest/api/python/pyspark.sql.html?highlight=udf#pyspark.sql.UDFRegistration.

Obviously we can't do the Python decorator pattern in C#, but it's pretty straight-forward to add a `_spark.Udf().RegisterVector` method. I didn't add an "evalType" publicly yet, since we only support one "vector" UDF evalType so far. But when we add grouped support, we can add the evalType parameter as well.

eerhardt:
I've responded to all feedback and updated the PR. Please take a look.

src/csharp/Microsoft.Spark/Sql/ArrowArrayHelpers.cs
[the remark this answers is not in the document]
eerhardt: Ah, this [comment](https://github.com/ashmind/SharpLab/issues/99#issuecomment-302850959) explains it.

I changed it to be `else if`. The "good" thing is this method shouldn't be on a hot path, as it is only necessary when Spark calls the UDF with no records.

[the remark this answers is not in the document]
stephentoub: Sorry I didn't see your question yesterday.  Yeah, the JIT generates code for a generic method for each value type that's used to instantiate it, but it shares a single generated piece of code for all reference types used to instantiate it.  So any optimizations that rely on the JIT being able to see the specific type of T at compile time generally only apply to value types.

benchmark/csharp/Tpch/TpchFunctionalQueries.cs
```diff
@@ -202,6 +226,40 @@ internal void Q8()
               .Show();
         }
 
+        internal void Q8v()
+        {
+            Func<Column, Column> getYear = Udf<string, string>(x => x.Substring(0, 4));
```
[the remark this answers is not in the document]
eerhardt: No, it doesn't.  I implemented both getYear and isBrazil as VectorUdfs, and it seems to slightly degrade the perf:

|                      Run|   With  Udf|   With VectorUdf|
|---------------------------- |----------:|----------:|
| #1 | 28.845 s | 29.246 s |
| #2 | 28.900 s | 29.527 s |
| #3 | 29.143 s | 29.328 s |

Intuitively, I don't think it will help much because these UDFs are not called with a lot of data at once, but instead they are called many times with only a few records at a time. Looking at a log output:

<details>
    <summary>Click to expand</summary>

        [Info] [DaemonWorker] Started .NET DaemonServer with port 39275.
        [Info] [DaemonWorker] New connection accepted for TaskRunner [1]
        [Info] [DaemonWorker] Pool snapshot: [NumThreads:1], [NumTaskRunners:1]
        [Info] [TaskRunner] [1] Starting with ReuseSocket[True].
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:827630
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:822015
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:820344
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:820366
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:820475
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:820506
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:820600
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:249279
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:43
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:37
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:45
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:37
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:34
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:37
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:42
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:46
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:49
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:45
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:45
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:31
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:49
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:46
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:45
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:36
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:32
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:54
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:42
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:50
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:27
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:39
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:52
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:38
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:32
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:33
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:42
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:34
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:39
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:55
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:46
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:59
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:31
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:40
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:40
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:34
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:36
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:51
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:45
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:30
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:47
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:49
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:56
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:33
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:47
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:47
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:45
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:38
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:47
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:42
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:58
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:36
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:36
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:34
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:42
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:51
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:49
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:42
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:34
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:51
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:47
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:51
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:37
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:52
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:34
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:34
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:60
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:45
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:28
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:43
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:39
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:34
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:53
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:37
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:62
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:53
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:49
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:50
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:51
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:52
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:39
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:47
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:43
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:39
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:35
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:39
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:47
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:52
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:42
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:40
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:45
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:38
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:35
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:35
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:49
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:46
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:42
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:53
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:51
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:37
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:40
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:37
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:33
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:46
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:55
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:52
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:50
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:40
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:49
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:50
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:42
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:50
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:39
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:50
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:53
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:40
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:50
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:38
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:52
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:50
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:34
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:54
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:35
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:35
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:49
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:46
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:42
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:46
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:39
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:49
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:39
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:46
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:55
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:46
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:45
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:39
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:54
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:38
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:33
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:40
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:35
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:43
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:46
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:35
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:48
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:53
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:44
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:31
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:37
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:49
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:45
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:41
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:53
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:62
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:35
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:40
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:30
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:37
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:51
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:38
        [Debug] [TaskRunner] [1] Received END_OF_STREAM signal.
        [Info] [TaskRunner] [1] Processed a task: readComplete:True, entries:47
</details>

You can see the first 8 processed tasks contain hundreds of thousands of entries. But then after that, it suddenly drops to ~40 entries per invocation. The first 8 tasks are processing the OrderLines (which there are millions of records). But the tasks after that are the joined orders.

If we could get Spark to process more entries per invocation, I think we could speed this query up.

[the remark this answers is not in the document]
imback82: I think it's a bit non-intuitive to have a vector UDF for one and regular UDFs for others. Since this Q8v is for vector UDFs, can we convert all UDFs to use VectorUDFs? This way, I think it is easier to benchmark VectorUDF vs. UDF.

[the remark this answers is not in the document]
eerhardt: > it's a bit non-intuitive to have a vector UDF for one and regular UDFs for others.

I'm not sure I fully agree. I would think the typical scenario would be to use the function that makes the most sense for the scenario at hand. I wouldn't expect users to use the same one across the board just because it makes sense in one spot.

I can change it to use all VectorUDFs for this function, if the goal is to compare one-at-a-time UDFs vs. the VectorUDFs across the board. But I thought the goal was to show the fastest way possible to execute the query.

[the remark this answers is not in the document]
rapoth: I had a slightly different idea in mind (which makes this conversation even more interesting :) ). There are a few reasons to get VectorUDFs across the board, in my opinion:

  1. It allows not only for a one-at-a-time comparison but also allows us to do a holistic comparison to understand the overall benefits for a given workload.
  2. We can use the comparison to actually evaluate whether it is worthwhile to profile a specific case and understand if anything else is needed. Optionally, we can get any other curious mind to look and contribute.
  3. Gives us a framework to suggest to customers who have been asking about this on whether VectorUDFs really help them or not for their use case (ideally, I was expecting this would make for an excellent blog post in the future that talks about when to use VectorUDFs and when not to).

I haven't paid too much attention to the effort required to enable this across the board though so I apologize in advance as I don't mean to rat-hole this discussion. Just thought I'll share my two cents :)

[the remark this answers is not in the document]
eerhardt: If the goal is to do a holistic comparison, then I would think you would want every permutation implemented, correct?

For example, if I had 2 UDFs in my query, and one UDF was faster written as a Vector UDF and the other was faster as a normal UDF, I would want the full table of `Normal/Normal`, `Normal/Vector`, `Vector/Normal`, `Vector/Vector`, to compare the differences. If I was only comparing `Normal/Normal` to `Vector/Vector`, the time differences might wash each other out, and it looks like it doesn't make a difference. When really the `Vector/Normal` permutation is the fastest and the `Normal/Vector` combination is the slowest.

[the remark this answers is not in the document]
imback82: I was thinking we could do "all normal UDFs" vs "all VectorUdFs", then analyze to come up with hybrid that will perform the best.

But, for the scope of this PR, I think the current approach is fine. We can sync up on the plan for TPCH benchmark in a separate thread or offline.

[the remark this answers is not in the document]
rapoth: I agree. We can keep it simple and do all normal vs. all vector to begin with. I do not want to delay this PR so I'll leave it to your judgement.

src/csharp/Microsoft.Spark.Experimental/Sql/ExperimentalFunctions.cs
```diff
@@ -0,0 +1,256 @@
...
+    public static class ExperimentalFunctions
+    {
+        /// <summary>Creates a Vector UDF from the specified delegate.</summary>
+        /// <typeparam name="T1">Specifies the type of the first argument to the UDF.</typeparam>
+        /// <typeparam name="RT">Specifies the return type of the UDF.</typeparam>
+        /// <param name="udf">The Vector UDF function implementation.</param>
+        /// <returns>
+        /// A delegate that returns a <see cref="Column"/> for the result of the Vector UDF.
+        /// </returns>
+        public static Func<Column, Column> VectorUdf<T1, RT>(Func<T1, RT> udf)
```
imback82: If we are following `Func<>`'s type parameter names, should we also change RT to TResult? (Also, T1 should be just T, for `Func` taking one T.)

src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
```diff
@@ -151,172 +164,211 @@ internal static JvmObjectReference CreatePythonFunction(IJvmBridge jvm, byte[] c
...
-                return (PicklingDelegate)
-                    new PicklingUdfWrapper<
-                        T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, TResult>(udf).Execute;
-            }
+            return (PicklingDelegate)
+                new PicklingUdfWrapper<
+                    T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, TResult>(udf).Execute;
+        }
+
+        internal static Delegate CreateVectorUdfWrapper<T1, TResult>(Func<T1, TResult> udf)
```
imback82: just `T`?

src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
```diff
@@ -151,172 +164,211 @@ internal static JvmObjectReference CreatePythonFunction(IJvmBridge jvm, byte[] c
...
-                return (ArrowDelegate)new ArrowUdfWrapper<TResult>(udf).Execute;
-            }
-            else
-            {
-                return (PicklingDelegate)new PicklingUdfWrapper<TResult>(udf).Execute;
-            }
+            return (PicklingDelegate)new PicklingUdfWrapper<TResult>(udf).Execute;
         }
 
         internal static Delegate CreateUdfWrapper<T1, TResult>(Func<T1, TResult> udf)
```
imback82: just `T`?

src/csharp/Microsoft.Spark.Experimental/Sql/ExperimentalFunctions.cs
eerhardt: Fixed all generic type names to match `Func<>`

suhsteve:
LGTM

PR #134, opened 2019-06-07
Update Microsoft.Spark.Worker exe path

suhsteve:
The `Microsoft.Spark.Worker` exe path was constructed based on the following conditions:

1) If `DOTNET_WORKER_DIR` environment variable is set, then use this to construct the Worker exe path
2) else if the Microsoft.Spark.Worker exists in the current assemblies directory, then use this to construct the path
3) otherwise assume that `Microsoft.Spark.Worker` exe can be accessed using the PATH env variable.

We remove condition 2 in this PR.
  - This condition works well if the `Microsoft.Spark.Worker` binary is colocated with the executing assembly during a `--master local` spark-submit run.  However this introduces issues when issuing a non-local spark-submit.  The driver will inform the executors that it has found the Microsoft.Spark.Worker located in a local path found on the driver.  This path may not exist on the executors and may result in a file not found error.
  - `Microsoft.Spark.Worker` binary location will either 1) need to be installed in a location that is part of the PATH env variable on the executor nodes or 2) `DOTNET_WORKER_DIR` needs to be defined on the driver node that specifies where the `Microsoft.Spark.Worker` can be found on the executors.


src/csharp/Microsoft.Spark.E2ETest/IpcTests/SparkContextTests.cs
```diff
@@ -1,44 +1,44 @@
...
-            sc.SetJobGroup("group id", "description");
-            sc.SetJobGroup("group id", "description", true);
-
-            sc.ClearJobGroup();
-
-            string filePath = $"{TestEnvironment.ResourceDirectory}people.txt";
-            sc.AddFile(filePath);
-            sc.AddFile(filePath, true);
-
+// Licensed to the .NET Foundation under one or more agreements.
```
suhsteve: Majority of the .cs files have CRLF line endings.  A few of the checked in .cs files had LF line endings.  The .cs files in this PR that look like they have no change were modified to have CRLF line endings to be consistent.

imback82:
Can you add a description please?

imback82:
> We remove condition 2 in this PR.
Can you explain why, for the record? Also, I think you need to update net461 README: https://github.com/dotnet/spark/blob/master/docs/building/windows-instructions.md#using-visual-studio-for-net-framework-461

src/csharp/Microsoft.Spark.Worker/Microsoft.Spark.Worker.csproj
```diff
@@ -4,6 +4,7 @@
     <OutputType>Exe</OutputType>
     <TargetFrameworks>net461;netcoreapp2.1</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.1</TargetFrameworks>
+    <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
```
suhsteve: This will allow the `dotnet publish -f netcoreapp2.1 -r <runtime>` to publish without error for the Tpch and Examples projects.

imback82: We decided not to go with this approach. You can search the issue in the old repo. 

Now that we have only one way to install worker, i think we should just remove dependencies on the worker.

Issue #120, opened 2019-05-22
[BUG]: The object in the lambda closure cannot be deserialized


suhsteve:
**Describe the bug**
The object in the lambda closure cannot be deserialized.

**To Reproduce**
Class Library: *ExternalLibrary.dll*
```csharp
using System;

namespace ExternalLibrary
{
    [Serializable]
    public class ExternalClass
    {
        private int _i;

        public ExternalClass(int i)
        {
            _i = i;
        }

        public int Compute()
        {
            return 2 * _i;
        }
    }
}
```

Console App: *SparkExample*
```csharp
using ExternalLibrary;
using Microsoft.Spark.Sql;
using static Microsoft.Spark.Sql.Functions;

namespace SparkExample
{
    class Program
    {
        static void Main(string[] args)
        {
            var spark = SparkSession.Builder().GetOrCreate();
            var df = spark.Read().Json(args[0]);

            var ex = new ExternalClass(100);
            var udf = Udf<string, int>(str => ex.Compute());
            df.Select(udf(df["name"])).Show();
        }
    }
}
```

1. Compile the above code.
2. Copy the following assemblies to the **Microsoft.Spark.Worker** folder:
    - ExternalLibrary.dll
    - SparkExample.dll
3. Run *org.apache.spark.deploy.DotnetRunner* in [Debug Mode](https://github.com/dotnet/spark/blob/master/docs/developer-guide.md#how-to-debug)
4. Run the SparkExample console app with the path to the *people.json* file as the first argument.

**Error Message**
```
[Error] [JvmBridge] org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 27.0 failed 1 times, most recent failure: Lost task 0.0 in stage 27.0 (TID 27, localhost, executor driver): org.apache.spark.api.python.PythonException: System.Runtime.Serialization.SerializationException: Unable to find assembly 'ExternalLibrary, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.
   at System.Runtime.Serialization.Formatters.Binary.BinaryAssemblyInfo.GetAssembly()
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.GetType(BinaryAssemblyInfo assemblyInfo, String name)
[...]

PR #135, opened 2019-06-11
Load Assembly referenced from within a lambda closure

suhsteve:
Please see example in issue #120 

The assembly containing the type used in the lambda closure is not loaded as part of the Worker's application domain.  When the Worker deserializes the object within the lambda, the type must be loaded from an external Assembly.

This issue seems to be isolated to dotnet core.  Using .NET framework, the assembly automatically loaded.


src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -299,5 +308,45 @@ private static Delegate CreateUdfWrapperDelegate<T>(Type type, object[] paramete
...
+        /// 2) The directory of the application
+        /// If the assembly is not found in the above locations, the exception from
+        /// Assembly.LoadFrom() will be propagated.
+        /// </summary>
+        /// <param name="assemblyName">The FullName of the assembly to load</param>
+        /// <returns>The loaded assembly</returns>
+        /// <exception cref = "System.IO.FileNotFoundException" > Thrown when the FullName of
+        /// the loaded assembly does not match the FullName of the assembly to load.</exception>
+
+        private static Assembly LoadAssembly(string assemblyName)
```
imback82: How about `LoadAssembly` inside UdfSerDe.cs?

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -299,5 +308,45 @@ private static Delegate CreateUdfWrapperDelegate<T>(Type type, object[] paramete
...
+        {
+            var sep = Path.DirectorySeparatorChar;
+            var asmSimpleName = assemblyName.Split(',').FirstOrDefault();
+            Assembly asm;
+            try
+            {
+                asm = Assembly.LoadFrom(
```
eerhardt: Have you checked out https://docs.microsoft.com/en-us/dotnet/framework/deployment/best-practices-for-assembly-loading? I don't think `Assembly.LoadFrom` is recommended anymore. But instead, using AssemblyLoadContext.

/cc @stephentoub

eerhardt: See also the way MSBuild loads assemblies for an example:

https://github.com/microsoft/msbuild/blob/fc10ea8ce260b764bb9fa5033b327af9fefcaabe/src/Shared/TaskEngineAssemblyResolver.cs#L127

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
suhsteve: Removed `LoadAssembl`y in CommandSerDe.cs and leveraging the one in UdfSerDe.cs

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
suhsteve: Thanks for the suggestion and finding the article.  For now we will create issue #138 and revisit the assembly loading in another PR.

azure-pipelines.yml
```diff
@@ -55,14 +55,14 @@ jobs:
     inputs:
       command: test
       projects: '**/*UnitTest/*.csproj'
-      arguments: '--configuration $(buildConfiguration) /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura'
```
suhsteve: Removing due to errors related to https://github.com/tonerdo/coverlet/issues/210

suhsteve:
This seems to be related to https://github.com/dotnet/coreclr/issues/11498 .  Let me investigate this a bit more.

azure-pipelines.yml
suhsteve: Will implement necessary changes in #138

imback82:
@suhsteve what will be the effort to do #138? Since both this PR and #138 are breaking changes, it will be nice to have this in one release.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
suhsteve: @eerhardt I investigated this a bit more.  The AssemblyLoadContext relies is only available for netcore and not net framework.  Ideally we would want to do something like the msbuild example, using #if preprocessor directives and choosing to use Assembly.LoadFrom or AssemblyLoadContext depending on the framework.  However since Microsoft.Spark library is targeting netstandard2.0 , i'm not sure how to go about doing this.   Would you have any suggestions on how we can do this ?

eerhardt: I think there are 2 possibilities:

1. Detect at runtime if we are running on .NET Framework - which you can use https://source.dot.net/#CoreFx.Private.TestUtilities/System/PlatformDetection.cs,24.
2. Multi-target Microsoft.Spark to build for both `net45` and `netcoreapp2.1`. But we would still need to target `netstandard2.0` as well so other `netstandard` libraries could reference Spark. This would involve more work and build time, since we would be compiling the assembly 3 times.

My suggestion would be to try (1) above, and fallback to (2) if necessary.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -68,7 +71,50 @@ internal sealed class FieldData
...
             public object Value { get; set; }
+
+            public void GetObjectData(SerializationInfo info, StreamingContext context)
+            {
+                info.AddValue("TypeData", TypeData, typeof(TypeData));
+
+                var bf = new BinaryFormatter();
+                using (var ms = new MemoryStream())
+                {
+                    bf.Serialize(ms, Value);
```
imback82: I guess null case is also tested?

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -68,7 +71,50 @@ internal sealed class FieldData
...
+                    {
+                        ms.Seek(0, SeekOrigin.Begin);
+                        DeserializeType(TypeData);
+                        Value = bf.Deserialize(ms);
+                    }
+                }
+            }
         }
 
         internal static UdfData Serialize(Delegate udf)
```
imback82: Can you add few tests to Microsoft.Spark.UnitTest for UdfSerDe class? (And an example with TargetData - it doesn't have to an external dll).

imback82:
> @suhsteve what will be the effort to do #138? Since both this PR and #138 are breaking changes, it will be nice to have this in one release.

Let's release this one first and come back to #138.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
suhsteve: @eerhardt thanks for your suggestions.

I've tried (1) and while it compiles, it fails at runtime, where it tries to load the incompatible library.  I suspect that the net461 Microsoft.Spark.Worker attempts to load it because Microsoft.Spark.dll still contains references to the library.

(2) worked, but we run into the concerns you brought up regarding building separate libraries targeting multiple frameworks.  What do you think about the following approach?

Add an AssemblyLoader.cs class to Microsoft.Spark (netstandard2.0) that looks like the following:
```csharp
    internal static class AssemblyLoader
    {
        internal static Func<string, Assembly> Loader { get; set; } = Assembly.Load;

        internal static Assembly LoadAssembly(string manifestModuleName)
        {
            var sep = Path.DirectorySeparatorChar;
            try
            {
                return Loader(
                    $"{Directory.GetCurrentDirectory()}{sep}{manifestModuleName}");
            }
            catch (FileNotFoundException)
            {
                return Loader(
                    $"{AppDomain.CurrentDomain.BaseDirectory}{sep}{manifestModuleName}");
            }
        }
```

In Microsoft.Spark.Worker, which targets (net461 and netcoreapp2.1), we can add `#if preprocessor directives` and set the above AssemblyLoader.Loader property to use AssemblyLoadContext if the target framework is netcore.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
suhsteve: null case is currently checked in
```csharp
        private static TargetData SerializeTarget(object target)
        {
...
                    ValueData = new ValueData()
                    {
                        TypeData = (value != null) ? SerializeType(value.GetType()) : default,
                        Value = value
                    }
```

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
suhsteve: I am loading the assembly by way of the delegate mentioned previously.  This seemed like the cleanest approach.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
suhsteve: Added some test cases.

src/csharp/Microsoft.Spark.Worker/Microsoft.Spark.Worker.csproj
```diff
@@ -8,6 +8,10 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <PropertyGroup Condition="$(TargetFramework.StartsWith('net4'))">
+    <DefineConstants>$(DefineConstants);FEATURE_ASSEMBLY_LOADFROM</DefineConstants>
```
eerhardt: You don't need to define your own constant for this. There are built-in constants generated for the TargetFramework that is currently being compiled.

In your C# code you can either say:

```C#
#if NETFRAMEWORK

// code only compiles for any .NET Framework version (net461, net45, etc)

#endif
```

or 

```C#
#if NET461

// code only compiles for net461

#endif
```

suhsteve: Thanks!

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -158,15 +227,15 @@ private static TargetData SerializeTarget(object target)
         private static object DeserializeTargetData(TargetData targetData)
         {
             Type targetType = DeserializeType(targetData.TypeData);
-            var target = Activator.CreateInstance(targetType);
+            var target = FormatterServices.GetUninitializedObject(targetType);
```
suhsteve: `Activator.CreateInstance(targetType)'` requires targetType to have a ctor().  `FormatterServices.GetUninitializedObject(targetType);` allows the creation of a new instance of targetType without having to call the ctor().

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -197,16 +266,15 @@ private static TypeData SerializeType(Type type)
         /// <returns>The loaded assembly</returns>
         private static Assembly LoadAssembly(string manifestModuleName)
         {
-            var sep = Path.DirectorySeparatorChar;
             try
             {
-                return Assembly.LoadFrom(
-                    $"{Directory.GetCurrentDirectory()}{sep}{manifestModuleName}");
+                return AssemblyLoader(
+                    Path.Combine(Directory.GetCurrentDirectory(), manifestModuleName));
```
eerhardt: Would it be a bad idea to check `File.Exists` before the `try-catch`? If there are many calls here, throwing and catching exceptions isn't the best.

suhsteve: Sounds like a good idea.  I implemented your suggestion.  Also got rid of the try/catch while deserializing ValueData by utilizing a cache.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
imback82: But this is a serializer for `ValueData`. What will happen if `Value` is set to null? `bf.Serialize(ms, Value);` is still safe?

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -184,30 +244,62 @@ private static TypeData SerializeType(Type type)
...
+
+        /// <summary>
+        /// Return the cached assembly, otherwise attempt to load and cache the assembly
+        /// in the following order:
+        /// 1) Search the assemblies loaded in the current app domain.
+        /// 2) Load the assembly from disk using manifestModuleName.
+        /// </summary>
+        /// <param name="assemblyName">The full name of the assembly</param>
+        /// <param name="manifestModuleName">Name of the module that contains the assembly</param>
+        /// <returns></returns>
```
imback82: missing comment.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -184,30 +244,62 @@ private static TypeData SerializeType(Type type)
...
-        /// Assembly.LoadFrom() will be propagated.
+        /// AssemblyLoader() will be propagated.
         /// </summary>
         /// <param name="manifestModuleName">The name of assembly to load</param>
         /// <returns>The loaded assembly</returns>
         private static Assembly LoadAssembly(string manifestModuleName)
         {
-            var sep = Path.DirectorySeparatorChar;
-            try
+            string assemblyPath1 =
```
imback82: can we give a better name than 1 and 2?

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -184,30 +244,62 @@ private static TypeData SerializeType(Type type)
...
         private static Type DeserializeType(TypeData typeData) =>
             s_typeCache.GetOrAdd(typeData,
-                td => LoadAssembly(typeData.ManifestModuleName).GetType(typeData.Name));
+                td => LoadAssembly(td.AssemblyName, td.ManifestModuleName).GetType(td.Name));
+
+        /// <summary>
+        /// Return the cached assembly, otherwise attempt to load and cache the assembly
+        /// in the following order:
+        /// 1) Search the assemblies loaded in the current app domain.
+        /// 2) Load the assembly from disk using manifestModuleName.
```
imback82: Is this same behavior as `AssemblyLoadContext`? If so, can you add a comment?

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -184,30 +244,62 @@ private static TypeData SerializeType(Type type)
...
+        /// <param name="assemblyName">The full name of the assembly</param>
+        /// <param name="manifestModuleName">Name of the module that contains the assembly</param>
+        /// <returns></returns>
+        private static Assembly LoadAssembly(string assemblyName, string manifestModuleName)
+        {
+            return s_assemblyCache.GetOrAdd(
+                assemblyName,
+                _ =>
+                {
+                    foreach (Assembly asm in AppDomain.CurrentDomain.GetAssemblies())
```
imback82: what's the size of the array retruned by `GetAssemblies()` when you run the worker?

src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
```diff
@@ -0,0 +1,178 @@
...
+                // Without Closure
+                Func<int, int> udf = i => 10 * i;
+                VerifyUdfSerDe(udf, false, false);
+            }
+
+            {
+                // With Closure
+                TestClass tc = new TestClass(20);
+                Func<int, int> udf = i => tc.MultiplyBy(i);
+                VerifyUdfSerDe(udf, true, false);
```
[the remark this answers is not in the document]
suhsteve: The lambda will create an anonymous class as the target, where the field of the target will be TestClass.  I set it to false because there didn't seem to be an easy way of defining a custom equals method for lambda functions.   I modified the test to instead just check the target fields for equality.

src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
```diff
@@ -0,0 +1,178 @@
...
+                    Assert.Equal(udf.Target, deserializedUdf.Target);
+                }
+            }
+        }
+
+        private void VerifyUdfData(Delegate udf, UdfSerDe.UdfData udfData, bool hasClosure)
+        {
+            VerifyTypeData(udf.Method.DeclaringType, udfData.TypeData);
+            Assert.Equal(udf.Method.Name, udfData.MethodName);
+            VerifyTargetData(udf.Target, udfData.TargetData, hasClosure);
```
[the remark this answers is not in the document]
suhsteve: Overrode `Equals` in the classes related to UdfData

src/csharp/Microsoft.Spark.Worker/Program.cs
```diff
@@ -5,6 +5,11 @@
 using System;
 using System.Diagnostics;
 
+#if !NETFRAMEWORK
```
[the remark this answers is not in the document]
suhsteve: Sounds good.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
suhsteve: Modified so ValueData can only be created via constructor.  Will throw an exception if value is null.  This will prevent TypeData and Value properties getting set to null.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
suhsteve: net461 starts at ~11 and netcoreapp2.1 ~26

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
suhsteve: https://docs.microsoft.com/en-us/dotnet/framework/deployment/best-practices-for-assembly-loading#avoid-loading-multiple-versions-of-an-assembly-into-the-same-context  suggests that we can use this method to check loaded assemblies.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -212,7 +212,7 @@ private sealed class UdfWrapperData
 
             foreach (UdfSerDe.FieldData field in fields)
             {
-                SerializeUdfs((Delegate)field.Value, curNode, udfWrapperNodes, udfs);
+                SerializeUdfs((Delegate)field.ValueData?.Value, curNode, udfWrapperNodes, udfs);
```
imback82: I think `ValueData` is always non-null. Should we either assert or  `field.ValueData.Value`? Otherwise, we will get null exception down the stream, so it may be a bit confusing.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -54,21 +62,175 @@ internal sealed class UdfData
...
+
+            public bool Equals(TargetData other)
+            {
+                if ((other == null) ||
+                    !TypeData.Equals(other.TypeData) ||
+                    (Fields?.Length != other.Fields?.Length))
+                {
+                    return false;
+                }
+                else if (Fields == null && other.Fields == null)
```
imback82: shouldn't this by just `if`?

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -54,21 +62,175 @@ internal sealed class UdfData
...
+                {
+                    return true;
+                }
+
+                for (int i = 0; i < Fields.Length; ++i)
```
imback82: you can use `SequenceEqual`

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
imback82: also, according to https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#layout-conventions
it should be:
```C#
if ((Fields == null) && (other.Fields == null))
```
(check other places as well)

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -184,30 +344,62 @@ private static TypeData SerializeType(Type type)
...
+                {
+                    foreach (Assembly asm in AppDomain.CurrentDomain.GetAssemblies())
+                    {
+                        if (asm.FullName.Equals(assemblyName))
+                        {
+                            return asm;
+                        }
+                    }
+
+                    return LoadAssembly(manifestModuleName);
```
imback82: can we add a comment this this can potentially load multiple dlls with the same name? (or we can fix the race condition in this PR)

src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
```diff
@@ -0,0 +1,150 @@
...
+
+                Delegate deserializedUdf = UdfSerDe.Deserialize(udfData);
+                Assert.Equal(udf.GetType(), deserializedUdf.GetType());
+                Assert.Equal(udf.Method, deserializedUdf.Method);
+                Assert.Equal(udf.Target.GetType(), deserializedUdf.Target.GetType());
+
+                FieldInfo[] expectedFields = udf.Target.GetType().GetFields();
+                FieldInfo[] actualFields = deserializedUdf.Target.GetType().GetFields();
+                Assert.Equal(expectedFields.Length, actualFields.Length);
+                for (int i = 0; i < expectedFields.Length; ++i)
```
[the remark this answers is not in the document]
suhsteve: In `VerifyUdfData` above we verify that the UdfData is was properly serialized/deserialized.  However in this case, we are verifying that the `Delegate deserializedUdf = UdfSerDe.Deserialize(udfData);` was properly deserialized into a `Delegate`.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
suhsteve: fixed race condition using Lazy<Assembly>().Value

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
suhsteve: ValueData can potentially be null.  I updated the UdfSerDeTests with an example.

src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
```diff
@@ -0,0 +1,152 @@
...
+            byte[] serializedUdf,
+            bool hasClosure)
+        {
+            using (var ms = new MemoryStream(serializedUdf, false))
+            {
+                var bf = new BinaryFormatter();
+                var udfData = (UdfSerDe.UdfData)bf.Deserialize(ms);
+                VerifyUdfData(UdfSerDe.Serialize(udf), udfData, hasClosure);
+
+                Delegate deserializedUdf = UdfSerDe.Deserialize(udfData);
```
imback82: I think we need `VerifyDelegate()` and `VerifyUdfData` separately to be clear.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -54,21 +62,168 @@ internal sealed class UdfData
...
+                    Equals(fieldData);
+            }
+
+            public bool Equals(FieldData other)
+            {
+                return (other != null) &&
+                    TypeData.Equals(other.TypeData) &&
+                    (Name == other.Name) &&
+                    (((ValueData == null) && (other.ValueData == null)) ||
+                    ((ValueData != null) && ValueData.Equals(other.ValueData)));
```
imback82: don't you need a tab here?

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
imback82: great!

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
imback82: Which test is covering this? Can `ValueData` be null in this context? Note that it's a delegate in this context (not capture). Can you check comments in line 70 in this file?

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
suhsteve: Was auto formatted.  Manually adding 4 space indent here.

src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
suhsteve: Added VerifyDelegate

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
suhsteve: Oh you're right.  My tests do not touch `CommandSerde.cs` .  Looks like it only gets to this case when the current node is a non-UDF, a MapUdfWrapper or WorkerChainHelper.  I'll revert this change.

src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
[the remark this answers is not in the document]
suhsteve: Refactored to make things easier to follow.

src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
```diff
@@ -0,0 +1,175 @@
...
+                string input = "HelloWorld";
+                string expectedResult = "TestHelloWorld";
+                Assert.Equal(expectedResult, expectedUdf(input));
+                Assert.Equal(expectedResult, ((Func<string, string>)actualUdf)(input));
+            }
+
+            {
+                // With closure where the delegate target is TestClass
+                // and target's field "_str" is set to "Test".
+                TestClass tc = new TestClass("Test");
```
imback82: How about the following? Also, I don't think we need to verify expectedUdf(input).
```C#
    TestClass tc = new TestClass("Test");
    Func<string, string> expectedUdf = tc.Concat;
    Delegate actualUdf = SerDe(expectedUdf);

    VerifyUdfSerDe(expectedUdf, actualUdf, true);
    Assert.Equal("TestHelloWorld", ((Func<string, string>)actualUdf)("HelloWorld"));
```

suhsteve: Sounds good.

Issue #140, opened 2019-06-14
Support custom worker path


danny8002:
consider the following use case:

Assume .net framework C# project (HelloWord)  will output helloword.exe which use Spark.Net. 

1) the Hello Worker c# project has a lot of dependencies 
2) when submit job to Spark with YARN:
   

- by nature,  use spark-submit with --files. as you know, --files will have long arguments, also maybe we will forget some files. 

- pack our project output as zip file, eg.  helloworld.zip 

`    spark-submit  xxxx --files hdfs://xxxx/helloword.zip  hdfs://xxx/microsoft-spark-2.3.x-0.2.0.jar helloworld.zip   helloword.exe  param_0`

  in this solution, Spark.Net (DotnetRunner) will unzip file, then start, but **worker can't get started** because Microsoft.Spark.Worker.exe is in helloword.zip file which is not get decompressed in worker machine.

so in order to let helloworld.zip get decompressed in worker, we can use **--archives** to let spark handle unzip issue. eg. 
`spark-submit  xxxx --archives hdfs://xxxx/helloword.zip  hdfs://xxx/microsoft-spark-2.3.x-0.2.0.jar 
 helloword.zip/helloword.exe param_0`

but worker still failed in starting Microsoft.Spark.Worker.exe because in worker the correct path is 
` (WorkingDirInWorker)/helloword.zip/Microsoft.Spark.Worker.exe`
 instead of 
`(WorkingDirInWorker)/Microsoft.Spark.Worker.exe`

so if **Spark.Net support configure custom worker path**. this can works perfect, 
here is commandline

`spark-submit  xxxx --archives hdfs://xxxx/helloword.zip  hdfs://xxx/microsoft-spark-2.3.x-0.2.0.jar  helloword.zip/helloword.exe param_0   helloword.zip/Microsoft.Spark.Worker.exe`

here is the sudo code

```
public class HelloWorld{
    static void Main(string[] args){
            var param0 = args[0];
            var workerName = args[1];

            // set custom worker path through env 
            Environment.SetEnvironmentVariable("XXX", workerName);

           SparkSession spark=xxxx
           
           spark.Udf().Register<string, string>("test", s=>s);

           // xxxx

    }
}

```

imback82:
Actually I was able to get `--archives worker.zip` working by doing on HDInsight (Linux)
```shell
export DOTNET_WORKER_DIR=./worker.zip
```

Did you try this? I tried this with the code in the master (commit 972825c)

[...]

PR #141, opened 2019-06-14
allow custom worker path

danny8002:

see #140.  
there are other reasons why we need this PR.
1) our production code use net47, and has no plan to migrate to dotnetCore since there are a lot of dependencies. 
2) in our Spark environment, we observed 'file missing' issue when YARN copy files from HDFS to worker node.  so if copy many files, the risk is high. 
3) we need net47 build for Spark.Net since we get this issue 
https://stackoverflow.com/questions/49925484/you-must-add-a-reference-to-assembly-netstandard-version-2-0-0-0
which suggest use net471 but currently we can't use net471 in our repo.

### DIFF
--- a/src/csharp/Microsoft.Spark/Services/ConfigurationService.cs
+++ b/src/csharp/Microsoft.Spark/Services/ConfigurationService.cs
@@ -15,6 +15,9 @@ namespace Microsoft.Spark.Services
     internal sealed class ConfigurationService : IConfigurationService
     {
         public const string WorkerDirEnvVarName = "DOTNET_WORKER_DIR";
+
+        public const string WorkerNameEnvVarName = "DOTNET_WORKER_NAME";
+
         public const string WorkerReadBufferSizeEnvVarName = "spark.dotnet.worker.readBufferSize";
         public const string WorkerWriteBufferSizeEnvVarName =
             "spark.dotnet.worker.writeBufferSize";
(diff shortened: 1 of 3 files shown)


imback82:
> 3\. we need net47 build for Spark.Net since we get this issue
>     https://stackoverflow.com/questions/49925484/you-must-add-a-reference-to-assembly-netstandard-version-2-0-0-0
>     which suggest use net471 but currently we can't use net471 in our repo.

@danny8002 can you elaborate on this? Do you get the error in the driver or from the worker that executes your UDFs? I am a bit confused looking your changes.

danny8002:
Now i am trying to repo this issue with a very simple project since my work code has modified to make it works. so for simplify this issue, I will remove net47 build target firstly.

danny8002:
without net47, any idea about this PR?

imback82:
This is taking longer since we need to verify the experience is consistent across all cloud providers we support. And we will need to update necessary README as well (contribution welcomed). Thanks for the patience.

src/csharp/Microsoft.Spark/Services/ConfigurationService.cs
```diff
@@ -15,6 +15,9 @@ namespace Microsoft.Spark.Services
     internal sealed class ConfigurationService : IConfigurationService
     {
         public const string WorkerDirEnvVarName = "DOTNET_WORKER_DIR";
+
```
imback82: let's remove this empty line.

src/csharp/Microsoft.Spark/Services/ConfigurationService.cs
```diff
@@ -62,19 +65,27 @@ public string GetWorkerExePath()
                 return _workerPath;
             }
 
+            var procName = s_procFileName;
```
imback82: `string procName =`

imback82: Can we also change
```C#
private static readonly string s_procBaseFileName = "Microsoft.Spark.Worker";
private static readonly string s_procFileName =
```
to
```C#

private static readonly string s_workerBaseName = "Microsoft.Spark.Worker";
private static readonly string s_workerName =
```

imback82: Once that is done, you can change this code block to:
```C#
string workerName = Environment.GetEnvironmentVariable(WorkerNameEnvVarName);
if (string.IsNullOrEmpty(workerName))
{
    workerName = s_workerName;
}
```

and use `workerName` in place of `procName` below.

imback82:
Actually I was able to get `--archives worker.zip` working without this change by doing
```
export DOTNET_WORKER_DIR=./worker.zip
```

Did you try this?

danny8002:
this PR is not needed any more, we can achieved it by setting DOTNET_WORKER_DIR=worker.zip/

imback82:
Closing this since this is not needed any more. (The same can be achieved via `DOTNET_WORKER_DIR`.)

Issue #145, opened 2019-06-15
[FEATURE REQUEST]: Add an ability to debug UDFs


imback82:
There has been lots of requests for debugging UDFs in Visual Studio. There is a way, but it requires code changes in the worker. Create a hook out-of-box and document the steps to debug UDFs.

imback82:
Closed by #150.

PR #150, opened 2019-06-26
Add a hook to launch a debugger from the worker process.

imback82:
This will enable #145.

The documentation on HowTo will be tracked with #151.

### DIFF
--- a/src/csharp/Microsoft.Spark.Worker/TaskRunner.cs
+++ b/src/csharp/Microsoft.Spark.Worker/TaskRunner.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Spark.Worker
         {
             s_logger.LogInfo($"[{TaskId}] Starting with ReuseSocket[{_reuseSocket}].");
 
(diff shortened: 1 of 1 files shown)


src/csharp/Microsoft.Spark.Worker/TaskRunner.cs
```diff
@@ -47,6 +49,11 @@ public void Run()
         {
             s_logger.LogInfo($"[{TaskId}] Starting with ReuseSocket[{_reuseSocket}].");
 
+            if (EnvironmentUtils.GetEnvironmentVariableAsBool("DEBUG_DOTNET_WORKER"))
```
suhsteve:
```suggestion
            if (EnvironmentUtils.GetEnvironmentVariableAsBool("DOTNET_WORKER_DEBUG"))
```

Follow same pattern as `DOTNET_WORKER_SPARK_VERSION` and `DOTNET_WORKER_DIR`

suhsteve:
LGTM

PR #167, opened 2019-07-09
Update Apache.Arrow reference to v0.14.

eerhardt:
Updating to the latest Arrow version which allows us to simplify some code, and brings in the latest perf fixes.

### DIFF
--- a/benchmark/csharp/Tpch/TpchFunctionalQueries.cs
+++ b/benchmark/csharp/Tpch/TpchFunctionalQueries.cs
@@ -10,6 +10,7 @@ using Apache.Arrow;
 using Microsoft.Spark.Sql;
 using static Microsoft.Spark.Sql.ExperimentalFunctions;
 using static Microsoft.Spark.Sql.Functions;
+using Column = Microsoft.Spark.Sql.Column;
 
 namespace Tpch
 {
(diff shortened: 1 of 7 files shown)


imback82:
Looks like TPCH build is failing.

eerhardt:
I'll fix it tomorrow.

eerhardt:
Should be fixed now.

imback82:
Looks like E2E is failing now. Does it repro locally?

eerhardt:
> Looks like E2E is failing now. 

The root cause for the failure is https://issues.apache.org/jira/browse/ARROW-5908. I'm working on getting a PR up for that fix. If I can get it into the next service release of Arrow, I will update dotnet/spark with it.

If not, there is a simple workaround I can do in the tests until we get the fix.

eerhardt:
Apache Arrow is planning a `0.14.1` release soon. The two fixes in the C# library will be in it. I will wait for that version to ship, and then update this PR.

The fixes in the patched version are:

https://github.com/apache/arrow/pull/4836
https://github.com/apache/arrow/pull/4851

eerhardt:
@imback82 @suhsteve - Apache.Arrow 0.14.1 was released yesterday. I've updated to that version which should unblock this PR. Please take a look.

PR #170, opened 2019-07-10
Compatibility check for Microsoft.Spark.dll in the worker.

imback82:
There will be a breaking change for UDF contract due to #135. Now, the worker needs to enforce some kind of versioning check to ensure that Microsoft.Spark.dll is compatible b/w the driver and worker.

### DIFF
--- a/src/csharp/Microsoft.Spark.Worker/TaskRunner.cs
+++ b/src/csharp/Microsoft.Spark.Worker/TaskRunner.cs
@@ -146,8 +146,7 @@ namespace Microsoft.Spark.Worker
                     return null;
                 }
 
-                // Check the version here.
-                payload.Version.
+                ValidateVersion(payload.Version);
 
                 DateTime initTime = DateTime.UtcNow;
 
@@ -187,7 +186,7 @@ namespace Microsoft.Spark.Worker
                     SerDe.Write(outputStream, (int)SpecialLengths.END_OF_DATA_SECTION);
                 }
 
-                LogStat(payload, commandExecutorStat, readComplete);
+                LogStat(commandExecutorStat, readComplete);
 
                 return payload;
             }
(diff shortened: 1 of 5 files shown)


suhsteve:
LGTM

danny8002:
is this because user code use new Microsoft.Spark.dll but the worker machine install an old Microsoft.Spark.Worker.exe ?

imback82:
> is this because user code use new Microsoft.Spark.dll but the worker machine install an old Microsoft.Spark.Worker.exe ?

Yes.

src/csharp/Microsoft.Spark.Worker/TaskRunner.cs
```diff
@@ -211,7 +213,22 @@ public void Stop()
...
+            // Initial version was shipped with version "1.0", so this needs to be adjusted
+            // to be compatible going forward.
+            if (versionStr == "1.0")
+            {
+                versionStr = "0.1.0";
+            }
+
+            if (new Version(versionStr) < new Version(Versions.CurrentVersion))
+            {
+                throw new Exception($"Upgrade Microsoft.Spark to '{Versions.CurrentVersion}+' from '{versionStr}+'.");
```
suhsteve: 100 char limit

src/csharp/Microsoft.Spark.Worker/TaskRunner.cs
```diff
@@ -211,7 +213,22 @@ public void Stop()
...
+            }
+        }
+
+        private void LogStat(CommandExecutorStat stat, bool readComplete)
         {
             s_logger.LogInfo($"[{TaskId}] Processed a task: readComplete:{readComplete}, entries:{stat.NumEntriesProcessed}");
```
suhsteve: 100 char limit

src/csharp/Microsoft.Spark.Worker/TaskRunner.cs
imback82: From coding style:
```
There are few exceptions to this rule:

Log message with string interpolation:
Logger.LogInfo($"This message {someVariable} is too long but try your best to fit in 100 character limit.");
```

src/csharp/Microsoft.Spark.Worker/TaskRunner.cs
imback82: same

src/csharp/Microsoft.Spark.Worker/TaskRunner.cs
suhsteve: Ok sounds good.

PR #180, opened 2019-07-19
SerDe referenced assemblies

suhsteve:
A delegate references assemblies that is not easily ascertained by inspecting its target, method and fields.  This PR addresses this by adding a assembly resolve event handler which will search the probing path and look for some combination of the assembly's simple name and known assembly extensions.


suhsteve:
@stephentoub can you take a look and let me know if this is on the right track or if there may be better alternatives?

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -0,0 +1,113 @@
...
+
+namespace Microsoft.Spark.Utils
+{
+    internal class AssemblyLoader
+    {
+        internal static Func<string, Assembly> LoadFromFile { get; set; } = Assembly.LoadFrom;
+
+        internal static Func<string, Assembly> LoadFromName { get; set; } = Assembly.Load;
+
+        internal static readonly ConcurrentDictionary<string, string> s_assemblyNameToFileName =
```
eerhardt: (nit) use a property and don't use `s_` prefix.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -212,7 +225,42 @@ private sealed class UdfWrapperData
...
+            AssemblyName[] referencedAssemblies,
+            Dictionary<string, string> udfRefAssemblies)
+        {
+            foreach (AssemblyName assemblyName in referencedAssemblies)
+            {
+                if (!udfRefAssemblies.ContainsKey(assemblyName.FullName))
+                {
+                    try
+                    {
+                        Assembly asm = Assembly.Load(assemblyName.FullName);
```
eerhardt: Should we really be forcing all these assemblies to load?

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -212,7 +225,42 @@ private sealed class UdfWrapperData
...
+                        udfRefAssemblies.Add(assemblyName.FullName, asm.ManifestModule.Name);
```
eerhardt: Is `ManifestModule.Name` guaranteed to be the name of the file on disk? I would have expected this to use `asm.Location`.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -336,64 +333,10 @@ private static TypeData SerializeType(Type type)
...
-            throw new FileNotFoundException(
-                $"Assembly files not found: '{currDirAsmPath}', '{currDomainBaseDirAsmPath}'");
-        }
+            s_typeCache.GetOrAdd(
+                typeData,
+                td => AssemblyLoader.LoadAssembly(
+                    td.AssemblyName,
+                    td.ManifestModuleName).GetType(td.Name));
     }
 }
```
eerhardt: Should we add some tests to ensure this behavior works and doesn't regress?

suhsteve:
> I don't fully understand the scenario you are solving here, nor why you need to serialize all the reference assemblies to the worker.
> 
Below are some examples where class `ExternalClass` is located in an external assembly:

In this case the `ExternalClass` is created outside the udf and it is part of the `udf1`'s closure.   When we inspect `udf1`, we can eventually find a reference to `ExternalClass` as part of the delegate target's fields, serialize that information and eventually send it to the worker.
```csharp
var ec = new ExternalClass();
Func<Column, Column> udf1 = Udf<string, string>(
    str =>
    {
         ec.Method1();
         $"hello {str}!"
    });
```

In this example where the `ExternalClass` is called as part of the `udf2`'s method instructions, I was not able to find a reference to the `ExternalClass`.  (Although we should be able to find the reference if we inspect each instruction in the body of the method)
```csharp
Func<Column, Column> udf2 = Udf<string, string>(
    str =>
    {
         var ec = new ExternalClass();
         ec.Method1();
         $"hello {str}!"
    });
```

> Typically, the referenced assemblies should either be next to the UDF assembly, or contained in the .NET core/framework directory. We should just be able to resolve the referenced assemblies on the worker by looking in one folder: where the UDF assembly is.

I think there are a couple ways to approach this problem.

1) if the Assembly FullName is something like `ExampleAssembly, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null` where the name is `ExampleAssembly`.  If it is guaranteed that the file that contains this assembly starts with `ExampleAssembly` and ends with either .dll, .exe, or has no extension, then we can have the Worker attempt to load just these files when it needs to resolve the assembly.
2) If the above is not true, then we can have the Worker attempt to load, *.dll, *.exe, and all files without extension from within the probing path.
3) The main reason I chose the above approach, is mainly because it may not always be the case where there is a direct mapping between the Assembly FullName to the file that actually contains the assembly.  I'm okay with going with either 1 or 2 if what we're currently doing is unnecessary.

eerhardt:
>  If it is guaranteed that the file that contains this assembly starts with ExampleAssembly and ends with either .dll, .exe, or has no extension, then we can have the Worker attempt to load just these files when it needs to resolve the assembly.

That's what MSBuild does when it tries resolving assemblies that a Task depends on:

https://github.com/microsoft/msbuild/blob/3457a9f5436442283d707726cd49ae3334b93334/src/Shared/CoreCLRAssemblyLoader.cs#L128

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
suhsteve: removed field.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
suhsteve: removed method.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
suhsteve: I'm not sure how guaranteed it is, but from the description it seems to be similar to asm.Location except with the path removed.

suhsteve: Since we're not sure, I'll use asm.Location just in case.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        private static readonly Dictionary<string, Assembly> s_assemblyCache =
+            new Dictionary<string, Assembly>();
+
+        private static readonly string[] s_searchPaths =
+            new[] { Directory.GetCurrentDirectory(), AppDomain.CurrentDomain.BaseDirectory };
+
+        private static readonly string[] s_extensions =
+            RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
+                    new[] { ".ni.dll", ".ni.exe", ".dll", ".exe" } :
```
eerhardt: you may want to put `.dll` first in the list, since that is typically the extension.

suhsteve: Thanks.  I'll list .exe after, which may be the next most typical extension.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -0,0 +1,145 @@
...
+        internal static void InstallHandler()
+        {
+            if (s_eventHandler == null)
+            {
+                s_eventHandler = new ResolveEventHandler(ResolveAssembly);
+                AppDomain.CurrentDomain.AssemblyResolve += s_eventHandler;
+            }
+        }
+
+        internal static void RemoveHandler()
```
eerhardt: Who calls this method?

eerhardt:
> using System;

(nit) copyright header

---
Refers to: src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs:1 in 7b4eeca. [](commit_id = 7b4eecab95be2655f71597410b545375ed031c79, deletion_comment = False)

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
suhsteve: No one calls at the moment.  I'll remove.

eerhardt: I think it might be better to be symmetrical between .NET Core and .NET Framework and have the same class responsible for hooking up the AssemblyResolve event.

---
In reply to: [306890361](https://github.com/dotnet/spark/pull/180#discussion_r306890361) [](ancestors = 306890361)

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
suhsteve: We have some tests in UdfSerDeTests in Microsoft.Spark.UnitTest.  Do you think we need to add a different type of test there ?

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
suhsteve: Sounds good, I added the AssemblyResolve event hookup code all in one place.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
eerhardt: In my opinion, yes we need different tests. The reasoning being is that the current tests aren't failing with the current code in master. So they aren't testing the scenario you are trying to enable here.

suhsteve: Added a few E2E tests.

suhsteve:
I believe I've addressed all the comments.  Let me know if there's anything else I should revisit.

external/Microsoft.Spark.E2ETest.ExternalLibrary/Microsoft.Spark.E2ETest.ExternalLibrary.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
eerhardt: Why not just put this project next to the unit tests?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSerDeTests.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            _df.Select(udf(_df["name"])).Show();
+        }
+
+        [Fact]
+        public void TestExternalStaticMethodCall()
+        {
+            Func<Column, Column> udf = Udf<string, string>(str =>
+            {
+                ExternalClass.HelloWorld();
```
imback82: Can you make this static method to return a string so that we know it is actually called?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSerDeTests.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            // Instantiate external assembly class within body of Udf.
+            Func<Column, Column> udf = Udf<string, string>(
+                    (str) =>
+                    {
+                        var ec = new ExternalClass("Hello");
+                        return ec.Concat(str);
+                    });
+
+            _df.Select(udf(_df["name"])).Show();
```
imback82: Instead of doing Show(), can we call `Collect()` and validate the output?

src/csharp/Microsoft.Spark.sln
```diff
@@ -25,6 +25,10 @@ Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Microsoft.Spark.FSharp.Exam
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Spark.Experimental", "Microsoft.Spark.Experimental\Microsoft.Spark.Experimental.csproj", "{7F276D07-6D94-49E3-A305-4B965DE7A700}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "external", "external", "{37E90B92-593A-47B6-BE89-F2F1E70FC79E}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Spark.E2ETest.ExternalLibrary", "..\..\external\Microsoft.Spark.E2ETest.ExternalLibrary\Microsoft.Spark.E2ETest.ExternalLibrary.csproj", "{4FB6EC00-0348-4E5A-93C2-5EAA60FD70B7}"
```
imback82: any reason this is outside src/csharp?

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -331,69 +250,15 @@ private static TypeData SerializeType(Type type)
             {
                 Name = type.FullName,
                 AssemblyName = type.Assembly.FullName,
-                ManifestModuleName = type.Assembly.ManifestModule.Name
+                AssemblyFileName = Path.GetFileName(type.Assembly.Location)
```
imback82: As far as I remember, `type.Assembly.ManifestModule.Name` also provided the assembly name. No?

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -0,0 +1,126 @@
...
+            {
+                if (s_assemblyCache.TryGetValue(assemblyName, out Assembly assembly))
+                {
+                    return assembly;
+                }
+
+                string simpleAsmName = new AssemblyName(assemblyName).Name;
+                foreach (string extension in s_extensions)
+                {
+                    var assemblyFileName = $"{simpleAsmName}{extension}";
```
imback82: `string` I think.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -0,0 +1,126 @@
...
+                    new[] { ".dll", ".exe", ".ni.dll", ".ni.exe" } :
+                    new[] { ".dll", ".ni.dll" };
+
+        private static readonly object s_cacheLock = new object();
+
+        /// <summary>
+        /// Return the cached assembly, otherwise attempt to load and cache the assembly
+        /// in the following order:
+        /// 1) Load the assembly from disk using assemblyFileName.
+        /// 2) Search for assemblyName in the current app domain.
```
imback82: question: what's the reason for this ordering?

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -0,0 +1,126 @@
...
+                return assembly;
+            }
+        }
+
+        /// <summary>
+        /// Return the cached assembly, otherwise look in the following probing paths,
+        /// searching for the simple assembly name and s_extension combination.
+        /// 1) The working directory
+        /// 2) The directory of the application
+        /// </summary>
```
imback82: I think it would be beneficial to add a comment why we decided to go with this order (in the cluster run, the dlls are usually shipped to the working directory, for example) since this is very specific to our use case.

external/Microsoft.Spark.E2ETest.ExternalLibrary/Microsoft.Spark.E2ETest.ExternalLibrary.csproj
suhsteve: Didn't think it would be necessary to clutter src/csharp with this project who only satisfies one set of tests.  I noticed spark had an [external](https://github.com/apache/spark/tree/master/external) folder which had some related external code so it seemed like it'd make sense to put it here.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
suhsteve: 1) if we already have a hint as to what the filename should be, then we will use this value first.
2) We will check if this assemblyname is currently can be found within the current domain/context.  Otherwise we'll try the other extensions to look for the assembly.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
suhsteve: Added some extra details regarding the probing path order.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
suhsteve: updated.  Although the type seems like it can be inferred based on the right hand value?

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
suhsteve: Eric had some concerns regarding ManifestModule.Name.  It seems for our tests, it has provided the assemblyfilename, but as I wasn't certain it's always the case, I opted to go this route.

src/csharp/Microsoft.Spark.sln
suhsteve: moved.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSerDeTests.cs
suhsteve: Collect().ToArray() called and values checked.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSerDeTests.cs
suhsteve: Changed.

external/Microsoft.Spark.E2ETest.ExternalLibrary/Microsoft.Spark.E2ETest.ExternalLibrary.csproj
suhsteve: moved to src/csharp

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSerDeTests.cs
```diff
@@ -35,19 +37,44 @@ public void TestUdfClosure()
                     return ec.Concat(str);
                 });
 
-            _df.Select(udf(_df["name"])).Show();
+            Row[] rows = _df.Select(udf(_df["name"])).Collect().ToArray();
+            Assert.Equal(3, rows.Length);
+
+            Row row1 = rows[0];
```
imback82: How about?
```C#
var expected = new string[]{ "HelloMichael", "HelloAndy", "HelloJustin" };
for (int i = 0; i < rows.Length; ++i)
{
    Row row = rows[i];
    Assert.Equal(1, row.Size());
    Assert.Equal(expected[i], row.GetAs<string>(0));
}
```

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
imback82: yes, but then it allows doing `var i = 0`. Let me update the coding style guideline on this (I thought I updated it on `var`, but I guess I didn't.).

external/Directory.Build.props
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
```
imback82: remove files under external?

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
imback82: Looks like the hint (the filename) is always there. If I ask it differently, what happens if we try to do 2) first?

suhsteve: To make the check stricter, we can just error when the assemblyFileName is not present and not worry about condition 2.

external/Directory.Build.props
suhsteve: removed.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
suhsteve: Updated and removed 2)

Issue #156, opened 2019-06-28
[FEATURE REQUEST]: Add RuntimeConfig and expose SparkSession.conf


suhsteve:
Expose the RuntimeConf to allow modifying the current SparkSession config without needing to create a new SparkSession.

```scala
  /**
   * Runtime configuration interface for Spark.
   *
   * This is the interface through which the user can get and set all Spark and Hadoop
   * configurations that are relevant to Spark SQL. When getting the value of a config,
   * this defaults to the value set in the underlying `SparkContext`, if any.
   *
   * @since 2.0.0
   */
  @transient lazy val conf: RuntimeConfig = new RuntimeConfig(sessionState.conf)
```

imback82:
closed by #184

PR #184, opened 2019-07-24
Support for SparkSession.conf

suhsteve:
This PR will allow retrieving and modifying a SparkSessions configuration without requiring a `stop` and recreation of the SparkSession.

### DIFF
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RuntimeConfigTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RuntimeConfigTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Spark.E2ETest.Utils;
 using Microsoft.Spark.Sql;
 using Xunit;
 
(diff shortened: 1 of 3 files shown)


suhsteve:
#156

src/csharp/Microsoft.Spark/Sql/RuntimeConfig.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <param name="key">Config name</param>
+        /// <param name="value">Config value</param>
+        public void Set(string key, long value) => _jvmObject.Invoke("set", key, value);
+
+        /// <summary>
+        /// Returns the value of Spark runtime configuration property for the given key.
+        /// </summary>
+        /// <param name="key">Key to use</param>
+        /// <param name="defaultValue">Default value to use</param>
+        public string Get(string key, string defaultValue) =>
```
imback82: Let's also expose `public string Get(string key)`.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RuntimeConfigTests.cs
```diff
@@ -0,0 +1,44 @@
...
+            conf.Set("longKey", 1234L);
+
+            Assert.Equal("stringValue", conf.Get("stringKey", "defaultValue"));
+            Assert.Equal("false", conf.Get("boolKey", "true"));
+            Assert.Equal("1234", conf.Get("longKey", "0"));
+
+            conf.Unset("stringKey");
+            Assert.Equal("defaultValue", conf.Get("stringKey", "defaultValue"));
+
+            Assert.False(conf.IsModifiable("missingKey"));
```
imback82: How about a test case that returns true?

src/csharp/Microsoft.Spark/Sql/RuntimeConfig.cs
suhsteve: I was considering this, but this will spark will throw a `java.util.NoSuchElementException` if the key does not exist.  This exact exception doesn't seem like it'll be propagated back through the `_jvmObject.Invoke` call.  I assume this was also why `SparkConf.cs` didn't implement this method.

imback82: We have many APIs where exceptions are thrown on the JVM side. I also, check the PySpark and they expose this function as well.

suhsteve: Added.

Issue #147, opened 2019-06-18
[Documentation] Best practice for writing UDFs


imback82:
Write a documentation on best practices for writing UDFs since there are few gotchas like static variable captures, etc.

imback82:
Closing this since #187 will handle the most cases. I will reopen this if there exists a special case for writing UDFs differently.

imback82:
Reopening since we still need to follow up with documentations capturing different scenarios of using UDFs.

PR #187, opened 2019-07-26
Expose a way to override assembly search paths used by AssemblyLoader.

imback82:
This will enable using --archives udf.zip so that you can put all the udf dlls in one zip file.
You can do this by setting AssemblySearchPathsEnvVarName:
```
export AssemblySearchPathsEnvVarName=./udf.zip
```

### DIFF
--- a/src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
+++ b/src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Spark.UnitTest
         [Serializable]
         private class TestClass
         {
-            private string _str;
+            private readonly string _str;
 
             public TestClass(string s)
             {
--- a/src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
+++ b/src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
@@ -6,18 +6,52 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
-using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 namespace Microsoft.Spark.Utils
 {
-    internal static class AssemblyLoader
+    internal static class AssemblySearchPathResolver
     {
-        // If this environment variable is set, the paths specified in this variable will have
-        // higher precedence over default search paths. Note that if a search path starts with
-        // ".", it will be replaced with the current directory.
         internal const string AssemblySearchPathsEnvVarName = "DOTNET_ASSEMBLY_SEARCH_PATHS";
 
+        // Returns the paths to search when loading assemblies in the following order of
+        // precedence:
+        // 1) Comma-separated paths specified in DOTNET_ASSEMBLY_SEARCH_PATHS environment variable.
+        //    - Note that if a path starts with ".", the working directory will be prepended.
+        // 2) The working directory.
+        // 3) The directory of the application.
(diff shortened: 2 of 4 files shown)


src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
```diff
@@ -16,7 +16,7 @@ public class UdfSerDeTests
         [Serializable]
         private class TestClass
```
rapoth: Can you update the documentation to include this feature? This is a pretty cool one so having a good e2e example would be great so I can tell everyone.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -10,6 +10,46 @@
...
+                Environment.GetEnvironmentVariable(AssemblySearchPathsEnvVarName);
+
+            if (!string.IsNullOrEmpty(searchPathsStr))
+            {
+                foreach (string searchPath in searchPathsStr.Split(','))
+                {
+                    string trimmedSearchPath = searchPath.Trim();
+                    if (trimmedSearchPath.StartsWith("."))
+                    {
+                        searchPaths.Add(
```
suhsteve: Will this run into any issues on windows `C:\some\path\.\other\path` which is a combination of `Path.Combine("C:\some\path", ".\other\path")` ?  Same question for linux.

suhsteve: I don't think it will, but asking just in case.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -10,6 +10,46 @@
 
 namespace Microsoft.Spark.Utils
 {
+    internal static class AssemblySearchPathResolver
+    {
+        internal const string AssemblySearchPathsEnvVarName = "DOTNET_ASSEMBLY_SEARCH_PATHS";
+
+        // Returns the paths to search when loading assemblies in the following order of
```
suhsteve: Can you update the comments in TryLoadAssembly  ?  You should be able to move the remarks regarding working dir here.

src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
imback82: Yes, I plan to do that as a part of #147 or a separate PR.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
imback82: This works on Linux, but I guess not on Windows?

suhsteve: Sorry I meant I don't think it will run into any issues.

imback82: Looks like this also works on Windows.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
imback82: Thanks. Updated.

src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
rapoth: Got it, thanks!

suhsteve:
LGTM

PR #188, opened 2019-07-26
Generate Microsoft.Spark.Worker-<version>.zip for Linux as well.

imback82:
Only tar.gz was generated for Linux. Generate .zip as well so that it can be used for --archives option.

### DIFF
--- a/script/package-worker.ps1
+++ b/script/package-worker.ps1
@@ -14,14 +14,13 @@ foreach ($framework in $frameworks)
         Copy-Item "$($runtime.FullName)\*" -Destination $worker_version_dir -Recurse
         $filename = "Microsoft.Spark.Worker.$framework.$runtime-$version"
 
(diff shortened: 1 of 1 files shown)


suhsteve:
Do you think we should just release .zip only and modify our other setup scripts/instructions to work with just zip ?

script/package-worker.ps1
```diff
@@ -14,14 +14,13 @@ foreach ($framework in $frameworks)
         Copy-Item "$($runtime.FullName)\*" -Destination $worker_version_dir -Recurse
         $filename = "Microsoft.Spark.Worker.$framework.$runtime-$version"
 
+        # Generate additional tar.gz worker files only for linux-x64.
```
rapoth: Given zip is universally available, do we lose anything by going with zip instead of tar.gz everywhere?

suhsteve:
LGTM

PR #193, opened 2019-07-31
Add zip files in working directory to assembly search path

### LINKED ISSUE #192: [FEATURE REQUEST]: Add Application .Zip file contents to Assembly Search Path
**Is your feature request related to a problem? Please describe.**
This is in relation to one of the new features added in v0.4.0:
> - The ability to use `--archives` option to deploy the worker binaries and assemblies that UDFs depend on (#187)

The `--archives` option only works for YARM setups.  Databricks, for example, does not use YARM so the `--archives` parameter has no effect.

**Describe the solution you'd like**
For the application .zip file to be transferred to all workers, extracted and the contents to be in the assembly search path.

**Describe alternatives you've considered**
A partial workaround is adding the applications .zip file name to the `--files` parameter meets the first requirement of transferring it to all workers.  However, it is not automatically extracted or added to the assembly search path.

Databricks Spark-Submit example:
```
["--files","DotNetSparkTest.zip","--class","org.apache.spark.deploy.dotnet.DotnetRunner","/dbfs/spark-dotnet/microsoft-spark-2.4.x-0.4.0.jar","/dbfs/apps/DotNetSparkTest.zip","DotNetSparkTest"]
```

AeroXuk:
This change extracts any zip files within the working directory and adds the extracted folders to the assembly search path.  When combined with the `--files` parameter allows you to copy the application dependencies to all workers and adds them to the assembly search path.

Fixes #192

### DIFF
--- a/src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
+++ b/src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.IO.Compression;
 using System.Reflection;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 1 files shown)


src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -56,6 +57,19 @@ internal static string[] GetAssemblySearchPaths()
             searchPaths.Add(Directory.GetCurrentDirectory());
             searchPaths.Add(AppDomain.CurrentDomain.BaseDirectory);
 
+            FileInfo[] archives = new DirectoryInfo(Directory.GetCurrentDirectory())
+                .GetFiles("*.zip");
+            foreach (var archive in archives)
+            {
```
AeroXuk: Personally, rather than extracting all .zip files within the folder. I would prefer to extract only the applications .zip file if present.  I don't, however, know the project well enough to know if it is possible to access the application .zip parameter that's passed to DotNetRunner from C#.

AeroXuk: See #194 as an alternative way of potentially implementing this.

PR #205, opened 2019-08-07
This commit adds two unit tests which mimic the reducers used in emai…

ruihmicrosoft:
…l search scenarios today. They are typical examples as it covers the two most common data transformation --- takes N rows and outputs 1 row, and takes N rows and outputs M rows. We use the two unit tests to demonstrate the feasibility to migrate existing reducers used by email search to HDI. Note this isn't a performance test yet.

1) The first reducer computes the top N contacts per user. It first groups by the user id and then collapses each group into one row by concatenating the records. A Udf is used in order to apply some string operations.
2) The second reducer takes N rows and outputs M rows. Given a succeeded search query with flattened folder and item list, the reducer explodes the input dataframe by pairing the corresponding folder and item ids. If the folder and item list contain multiple records, the reducer will generate multiple rows as output.

- [ ] There's a descriptive title that will make sense to other developers some time from now. 


imback82:
@ruihmicrosoft, we have been adding examples with some documentations under the example folder: #319, #320. If you want to push forward this PR, can you please convert this to an example, not E2E test? Thanks.

Issue #207, opened 2019-08-14
[BUG]: UDFs do not work on MacOS


cosmincatalin:
I have a very simple example where UDF's do not seem to work properly on MacOS.
Here is my set-up:

* Environment variable pointing to Spark:
```bash
➜  env | grep DOTNET_WORKER_DIR
DOTNET_WORKER_DIR=/Users/cosmin/lib/Microsoft.Spark.Worker/
```
* Contents of the folder the environment variable is pointing to:
```bash
➜  ls /Users/cosmin/lib/Microsoft.Spark.Worker/
Microsoft.Spark.Worker-0.4.0\Apache.Arrow.dll
Microsoft.Spark.Worker-0.4.0\Microsoft.CSharp.dll
Microsoft.Spark.Worker-0.4.0\Microsoft.Spark.Worker
Microsoft.Spark.Worker-0.4.0\Microsoft.Spark.Worker.deps.json
Microsoft.Spark.Worker-0.4.0\Microsoft.Spark.Worker.dll
Microsoft.Spark.Worker-0.4.0\Microsoft.Spark.Worker.pdb
Microsoft.Spark.Worker-0.4.0\Microsoft.Spark.Worker.runtimeconfig.json
Microsoft.Spark.Worker-0.4.0\Microsoft.Spark.Worker.xml
Microsoft.Spark.Worker-0.4.0\Microsoft.Spark.dll
Microsoft.Spark.Worker-0.4.0\Microsoft.Spark.pdb
Microsoft.Spark.Worker-0.4.0\Microsoft.Spark.xml
Microsoft.Spark.Worker-0.4.0\Microsoft.VisualBasic.dll
...............
Microsoft.Spark.Worker-0.4.0\libhostpolicy.dylib
Microsoft.Spark.Worker-0.4.0\libmscordaccore.dylib
Microsoft.Spark.Worker-0.4.0\libmscordbi.dylib
Microsoft.Spark.Worker-0.4.0\libsos.dylib
Microsoft.Spark.Worker-0.4.0\mscorlib.dll
Microsoft.Spark.Worker-0.4.0\netstandard.dll
Microsoft.Spark.Worker-0.4.0\sosdocsunix.txt
```
* Program listing
```csharp
using Microsoft.Spark.Sql;

namespace HelloSpark
{
    class Program
    {
        static void Main(string[] args)
        {
            var spark = SparkSession.Builder().GetOrCreate();
[...]

PR #209, opened 2019-08-14
Package Worker zip fix for Mac OSX

suhsteve:
`zip` files created using `Compress-Archive` in powershell causes incorrect behavior when using OSX's `Archiver` to extract.  There is a bug in Powershell module Microsoft.PowerShell.Archive.  This PR will update the module before creating the `zip`.


suhsteve:
#207

PR #228, opened 2019-08-29
DataFrameReader.Load() calls different Scala overload depending on path count

AFFogarty:
In Scala, `DataFrameReader.load()` has 3 overloads:
 
* `load()`
* `load(path: String)`
* `load(paths: String*)`

Some popular Scala libraries only support particular overloads of this function.

Until now, .Net for Spark always calls `load(path: String*)`, even if no paths or a single path are provided.

This PR changes the C# implementation of `DataFrameReader.Load(params string[] paths)` to call the appropriate Scala overload of `load()` depending on the number of paths provided to the `paths` parameter.

This is related to Issue #229 .

### DIFF
--- a/src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
+++ b/src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
@@ -114,8 +114,19 @@ namespace Microsoft.Spark.Sql
         /// </remarks>
         /// <param name="paths">Input paths</param>
         /// <returns>DataFrame object</returns>
-        public DataFrame Load(params string[] paths) =>
-            new DataFrame((JvmObjectReference)_jvmObject.Invoke("load", (object)paths));
+        public DataFrame Load(params string[] paths)
+        {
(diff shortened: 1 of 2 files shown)


src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
```diff
@@ -114,8 +114,20 @@ public DataFrameReader Options(Dictionary<string, string> options)
         /// </remarks>
         /// <param name="paths">Input paths</param>
         /// <returns>DataFrame object</returns>
-        public DataFrame Load(params string[] paths) =>
-            new DataFrame((JvmObjectReference)_jvmObject.Invoke("load", (object)paths));
+        public DataFrame Load(params string[] paths)
```
imback82: Instead of doing this, let's expose, `Load()` and `Load(string path)`.

AFFogarty: Ok

PR #235, opened 2019-09-04
Implement MapType in ComplexTypes.cs

### LINKED ISSUE #185: [FEATURE REQUEST]: Implement MapType in ComplexTypes.cs.
The `MapType` in `ComplexTypes.cs` is not implemented: https://github.com/dotnet/spark/blob/master/src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs#L81

elvaliuliuliu:
This fixes #185.

### DIFF
--- a/src/csharp/Microsoft.Spark.UnitTest/Sql/TypesTests.cs
+++ b/src/csharp/Microsoft.Spark.UnitTest/Sql/TypesTests.cs
@@ -45,7 +45,25 @@ namespace Microsoft.Spark.UnitTest
             Assert.Equal("integer", arrayType.ElementType.TypeName);
             Assert.True(arrayType.ContainsNull);
         }
-
+        
+        [Fact]
+        public void TestMapType()
+        {
+            var schemaJson =
+                @"{
+                    ""type"":""map"",
+                    ""keyType"":""integer"",
+                    ""valueType"":""double"",
+                    ""valueContainsNull"":true
+                }";
+            var mapType = (MapType)DataType.ParseDataType(schemaJson);
+            Assert.Equal("map", mapType.TypeName);
+            Assert.Equal("map<integer,double>", mapType.SimpleString);
+            Assert.Equal("integer", mapType.KeyType.TypeName);
+            Assert.Equal("double", mapType.ValueType.TypeName);
+            Assert.True(mapType.ValueContainsNull);
+        }
+        
         [Fact]
         public void TestStructTypeAndStructFieldTypes()
         {
--- a/src/csharp/Microsoft.Spark.UnitTest/Sql/TypesTests.cs
+++ b/src/csharp/Microsoft.Spark.UnitTest/Sql/TypesTests.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Spark.UnitTest
             Assert.Equal("array", arrayType.TypeName);
             Assert.Equal("array<integer>", arrayType.SimpleString);
             Assert.Equal("integer", arrayType.ElementType.TypeName);
-            Assert.True(arrayType.ContainsNull);
+            Assert.False(arrayType.ContainsNull);
         }
         
         [Fact]
@@ -61,7 +61,7 @@ namespace Microsoft.Spark.UnitTest
             Assert.Equal("map<integer,double>", mapType.SimpleString);
             Assert.Equal("integer", mapType.KeyType.TypeName);
             Assert.Equal("double", mapType.ValueType.TypeName);
-            Assert.True(mapType.ValueContainsNull);
+            Assert.False(mapType.ValueContainsNull);
         }
         
         [Fact]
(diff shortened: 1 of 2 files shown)


src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -71,14 +71,72 @@ private DataType FromJson(JObject json)
...
     {
         /// <summary>
-        /// Initializes the <see cref="MapType"/> instance.
+        /// Constructor for MapType class.
         /// </summary>
-        public MapType() => throw new NotImplementedException();
+        /// <param name="keyType">The data type of keys in this map</param>
+        /// <param name="valueType">The data type of values in this map</param>
+        /// <param name="valueContainsNull">Indicates if values can be null</param>
+        public MapType(DataType keyType, DataType valueType, bool valueContainsNull = true)
```
imback82: any reason this defaults to `true`?

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -71,14 +71,72 @@ private DataType FromJson(JObject json)
...
+        /// </summary>
+        /// <param name="json">JSON object used to construct a MapType object</param>
+        /// <returns>MapType object</returns>
+        private DataType FromJson(JObject json)
+        {
+            KeyType = ParseDataType(json["keyType"]);
+            ValueType = ParseDataType(json["valueType"]);
+            ValueContainsNull = (bool)json["valueContainsNull"];
+            return this;
+        }
```
imback82: @suhsteve what's the strategy for unit tests for these classes? If there is no existing tests, please create an issue so we can track.

imback82: Looks like we already have one: TypesTests.cs. @elvaliuliuliu please add unit tests.

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
elvaliuliuliu: Spark sets up this defaults to `true`. Please refer to https://github.com/apache/spark/blob/2b24a71fec021755f43db99628a56bd4a01518eb/sql/catalyst/src/main/scala/org/apache/spark/sql/types/MapType.scala#L97

imback82: Got it. thanks.

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
elvaliuliuliu: @imback82 I have added maptype in unit tests.

src/csharp/Microsoft.Spark.UnitTest/Sql/TypesTests.cs
```diff
@@ -45,7 +45,25 @@ public void TestArrayType()
...
+        
+        [Fact]
+        public void TestMapType()
+        {
+            var schemaJson =
+                @"{
+                    ""type"":""map"",
+                    ""keyType"":""integer"",
+                    ""valueType"":""double"",
+                    ""valueContainsNull"":true
```
imback82: Since the default value of `valueContainsNull` is true, let's test with `false`.

src/csharp/Microsoft.Spark.UnitTest/Sql/TypesTests.cs
```diff
@@ -45,7 +45,25 @@ public void TestArrayType()
             Assert.Equal("integer", arrayType.ElementType.TypeName);
             Assert.True(arrayType.ContainsNull);
         }
-
+        
+        [Fact]
+        public void TestMapType()
+        {
+            var schemaJson =
```
imback82: Since we don't use `var` unless you are using `new` or casting, can we do `string schemaJson` here and fix other places in this file?

src/csharp/Microsoft.Spark.UnitTest/Sql/TypesTests.cs
imback82: Can you also update `containsNull` to use false in `TestArrayType` above?

elvaliuliuliu: I have changed both `containsNull` and `valueContainsNull` to use `false`

src/csharp/Microsoft.Spark.UnitTest/Sql/TypesTests.cs
elvaliuliuliu: Sure, I have changed `var schemaJson` to `string schemaJson` in this file.

PR #247, opened 2019-09-17
Enable IsPackable for Microsoft.Spark.Extensions.Delta 

AFFogarty:
This is a simple PR that sets `<IsPackable>true</IsPackable>` for the Delta extension.

### DIFF
--- a/src/csharp/Extensions/Microsoft.Spark.Extensions.Delta/Microsoft.Spark.Extensions.Delta.csproj
+++ b/src/csharp/Extensions/Microsoft.Spark.Extensions.Delta/Microsoft.Spark.Extensions.Delta.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <IsPackable>true</IsPackable>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


src/csharp/Extensions/Microsoft.Spark.Extensions.Delta/Microsoft.Spark.Extensions.Delta.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <IsPackable>true</IsPackable>
   </PropertyGroup>
```
imback82: Can you add more fields like you found here: https://github.com/dotnet/spark/blob/master/src/csharp/Microsoft.Spark.Experimental/Microsoft.Spark.Experimental.csproj?

AFFogarty: Added the rest of the fields.

imback82:
@AFFogarty can you fix the tests? Now that `<GenerateDocumentationFile>true</GenerateDocumentationFile>` is introduced, it's validating the comments.

AFFogarty:
Made a change to escape the `<` and `>` in Doc code examples.  Otherwise, the build is broken when `<GenerateDocumentationFile>true</GenerateDocumentationFile>` is enabled.

Issue #246, opened 2019-09-17
[FEATURE REQUEST]: Expose DataFrameReader.Schema() that takes in StructType.


imback82:
Expose DataFrameReader.Schema() that takes in StructType that will define the schema for the data you are reading from.

We want to achieve something similar to https://spark.apache.org/docs/2.3.0/sql-programming-guide.html#programmatically-specifying-the-schema.

imback82:
cc: @elvaliuliuliu

imback82:
Fixed by #248

PR #248, opened 2019-09-18
Expose Schema() that takes in StructType

elvaliuliuliu:
This pr is the StructType Schema support for DataStreamReader as well as DataFrameReader.
It specifies the input schema that takes in StructType. By specifying the schema here, the underlying data source can skip the schema inference step, and thus speed up data loading.

This PR will fix issue #246.


src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameReaderTests.cs
```diff
@@ -27,7 +27,14 @@ public void TestSignaturesV2_3_X()
             DataFrameReader dfr = _spark.Read();
 
             Assert.IsType<DataFrameReader>(dfr.Format("json"));
-
+            
+            Assert.Throws<NotSupportedException>(() => Assert.IsType<DataFrameReader>(
```
imback82: Actually, the goal of #246 is to implement it such that this exception is not thrown.

elvaliuliuliu: Okay, I see. I will change it accordingly. Thanks!

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameReaderTests.cs
```diff
@@ -27,7 +28,14 @@ public void TestSignaturesV2_3_X()
             DataFrameReader dfr = _spark.Read();
 
             Assert.IsType<DataFrameReader>(dfr.Format("json"));
-
+            
+            Assert.IsType<DataFrameReader>(
+                dfr.Schema(
+                    new StructType(new[]
+                    {
+                        new StructField("age", new IntegerType()),
```
imback82: How is this working without `StructField` having JVMReference?

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -235,6 +235,22 @@ public sealed class StructType : DataType
         /// </summary>
         public List<StructField> Fields { get; private set; }
 
+        /// <summary>
+        /// Constructor for StructType class.
+        /// </summary>
+        JvmObjectReference IJvmObjectReferenceProvider.Reference
```
imback82: One reason I am not sure if this is the right approach or not is that we have `public StructType(IEnumerable<StructField> fields)`. For example, if I create a `StructType` object, I can have two objects one in JVM and on in CLR. 

If you look at other class that derives `IJvmObjectReferenceProvider`, it's just a thin wrapper.

@suhsteve did we compare with PySpark's approach?

suhsteve: Yes I've pointed out two approaches that can possibly be taken.  The Jvm object route, or checking the pyspark implementation.

imback82: JVM object route doesn't seem to make sense, right?

suhsteve: Yes I don't think we need a JVM object for every structtype/structfield.  If we went that route, that would also mean that when we deserialize rows, we would be creating JVM StructType object(s ) for every batch.

elvaliuliuliu: Yes, when I was checking PySpark's approach, I did encounter some issues with `SparkSession` and `parseDataType`, then I tried Jvm object route. I will explore more of PySpark's approach. Thanks!

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameReaderTests.cs
elvaliuliuliu: I am not so sure why only `StructType` has JVMReference works. But I will explore more of PySpark's implementaion for now.

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
suhsteve: I think you should explore the [parseDatatType](https://github.com/apache/spark/blob/fff2e847c2fb58e4ef50063e70fa0053498a4109/python/pyspark/sql/readwriter.py#L94) route again.

src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
```diff
@@ -22,7 +23,7 @@ internal DataFrameReader(JvmObjectReference jvmObject)
         }
 
         JvmObjectReference IJvmObjectReferenceProvider.Reference => _jvmObject;
-
+
```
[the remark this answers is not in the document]
elvaliuliuliu: Removed

src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
```diff
@@ -34,7 +35,25 @@ public DataFrameReader Format(string source)
             _jvmObject.Invoke("format", source);
             return this;
         }
-
+        
+        /// <summary>
+        /// Specifies the schema by using StructType object.
```
suhsteve: Let's try to enrich the comments
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/cref-attribute

```suggestion
        /// Specifies the schema by using <see cref="StructType"/>.
```

src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
```diff
@@ -34,7 +35,25 @@ public DataFrameReader Format(string source)
...
+        /// </summary>
+        /// <remarks>
+        /// Some data sources (e.g. JSON) can infer the input schema automatically
+        /// from data. By specifying the schema here, the underlying data source can
+        /// skip the schema inference step, and thus speed up data loading.
+        /// </remarks>
+        /// <param name="schema">class:`Microsoft.Spark.Sql.Types.StructType` object</param>
```
suhsteve:
```suggestion
        /// <param name="schema">The input schema</param>
```

src/csharp/Microsoft.Spark/Sql/Streaming/DataStreamReader.cs
```diff
@@ -30,6 +31,24 @@ public DataStreamReader Format(string source)
             return this;
         }
 
+        /// <summary>
+        /// Specifies the schema by using StructType object.
```
suhsteve:
```suggestion
        /// Specifies the schema by using <see cref="StructType"/>.
```

src/csharp/Microsoft.Spark/Sql/Streaming/DataStreamReader.cs
```diff
@@ -30,6 +31,24 @@ public DataStreamReader Format(string source)
...
+        /// </summary>
+        /// <remarks>
+        /// Some data sources (e.g. JSON) can infer the input schema automatically
+        /// from data. By specifying the schema here, the underlying data source can
+        /// skip the schema inference step, and thus speed up data loading.
+        /// </remarks>
+        /// <param name="schema">class:`Microsoft.Spark.Sql.Types.StructType` object</param>
```
suhsteve:
```suggestion
        /// <param name="schema">The input schema</param>
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamReaderTests.cs
```diff
@@ -30,6 +31,12 @@ public void TestSignaturesV2_3_X()
 
             Assert.IsType<DataStreamReader>(dsr.Format("parquet"));
 
+            Assert.IsType<DataStreamReader>(
+                dsr.Schema(
+                    new StructType(new[]
+                    {
+                        new StructField("columnName", new IntegerType())
+                    })));
```
suhsteve: I think this can fit in one line.

src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
elvaliuliuliu: Sounds good. Thanks.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamReaderTests.cs
elvaliuliuliu: I am following the format in [dictionary](https://github.com/elvaliuliuliu/spark/blob/37be26f8fe816ea9556bd53247aa3e8f8b755562/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamReaderTests.cs#L47)

I think we can either change both to one line or keep separate to be consistent. Thanks.

src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
elvaliuliuliu: May I know why it's better to specify as `The input schema` than mention `StructType`?

suhsteve: It's more descriptive than reiterating the type.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamReaderTests.cs
suhsteve: Sure you can keep it the way it is.

elvaliuliuliu: Cool, sounds good.

src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
elvaliuliuliu: Sounds good. Thanks for the suggestion.

src/csharp/Microsoft.Spark/Sql/Streaming/DataStreamReader.cs
```diff
@@ -30,6 +31,24 @@ public DataStreamReader Format(string source)
...
+        /// <remarks>
+        /// Some data sources (e.g. JSON) can infer the input schema automatically
+        /// from data. By specifying the schema here, the underlying data source can
+        /// skip the schema inference step, and thus speed up data loading.
+        /// </remarks>
+        /// <param name="schema">The input schema</param>
+        /// <returns>This DataStreamReader object</returns>
+        public DataStreamReader Schema(StructType schema)
+        {
+            var jvmSchema = (JvmObjectReference)_jvmObject.Jvm.CallStaticJavaMethod(
```
imback82: How about we create:
```C#
internal static JvmObjectReference FromJson(IJvmBridge jvm, string json)
```
inside `DataType.cs`?

Can you also search for `org.apache.spark.sql.types.DataType`? I see two other places where `fromJson` is called.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamReaderTests.cs
```diff
@@ -30,6 +31,12 @@ public void TestSignaturesV2_3_X()
 
             Assert.IsType<DataStreamReader>(dsr.Format("parquet"));
 
+            Assert.IsType<DataStreamReader>(
+                dsr.Schema(
+                    new StructType(new[]
```
imback82: can we add an e2e test that tests `fromJson` behavior for all types similar to `TypesTests.cs` in `Microsoft.Spark.UnitTest`? You can do this as a follow up PR.

elvaliuliuliu: Sure, will write a follow up PR to add an e2e test of `fromJson`.

src/csharp/Microsoft.Spark/Sql/Streaming/DataStreamReader.cs
elvaliuliuliu: In `org.apache.spark.sql.types.DataType`, `fromJson` is called [here](https://github.com/apache/spark/blob/fe1f456b200075bef476b5d8eca2be505b486033/sql/catalyst/src/main/scala/org/apache/spark/sql/types/DataType.scala#L129). And there is another `fromJson` is called from `org.apache.spark.sql.types.Metadata` [here](https://github.com/apache/spark/blob/86b25c43506a0a858601bd1082e9f96bb0415eb8/sql/catalyst/src/main/scala/org/apache/spark/sql/types/Metadata.scala#L129). I think the rest `fromJson` methods are not related.

imback82: oh I meant from our C# repo.

imback82: Check `UserDefinedFunction.cs` and `UDFRegistration.cs`

elvaliuliuliu: oh I see. Yes, one is from [here](https://github.com/dotnet/spark/blob/815c2a9a3e4dbbf0004279d07672710948e76d32/src/csharp/Microsoft.Spark/Sql/Expressions/UserDefinedFunction.cs#L37) in `UserDefinedFunction.cs` and the other is from [here](https://github.com/dotnet/spark/blob/815c2a9a3e4dbbf0004279d07672710948e76d32/src/csharp/Microsoft.Spark/Sql/UDFRegistration.cs#L270) in `UDFRegistration.cs`.

elvaliuliuliu: Added `FromJson` inside `DataType.cs`. Thanks.

src/csharp/Microsoft.Spark/Sql/Types/DataType.cs
```diff
@@ -67,7 +67,21 @@ public string Json
         /// JSON value of this data type.
         /// </summary>
         internal virtual object JsonValue => TypeName;
-
+        
+        /// <summary>
+        /// Parses a JSON string to create a JvmObjectReference object.
```
imback82: , which references a StructType object on the JVM side.

imback82:
@AFFogarty can you follow up on the following TODO after this PR is merged?

https://github.com/dotnet/spark/blob/master/src/csharp/Extensions/Microsoft.Spark.Extensions.Delta.E2ETest/DeltaTableTests.cs#L217

elvaliuliuliu:
@imback82 Should I replace DDL with StructType in `Basic.cs` or add StructType?

src/csharp/Microsoft.Spark/Sql/Types/DataType.cs
elvaliuliuliu: Sure, added.

src/csharp/Microsoft.Spark/Sql/Types/DataType.cs
```diff
@@ -69,7 +69,7 @@ public string Json
         internal virtual object JsonValue => TypeName;
         
         /// <summary>
-        /// Parses a JSON string to create a JvmObjectReference object.
+        /// Parses a JSON string to create a JvmObjectReference object, which references a StructType object on the JVM side.
```
imback82: this will be over the number of charcacters limit

imback82:
> @imback82 Should I replace DDL with StructType in `Basic.cs` or add StructType?

I would replace but leave the DDL in a comment.

src/csharp/Microsoft.Spark/Sql/Streaming/DataStreamReader.cs
imback82: those two files should use `DataType.FromJson`.

src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
```diff
@@ -34,7 +35,24 @@ public DataFrameReader Format(string source)
...
+        /// Some data sources (e.g. JSON) can infer the input schema automatically
+        /// from data. By specifying the schema here, the underlying data source can
+        /// skip the schema inference step, and thus speed up data loading.
+        /// </remarks>
+        /// <param name="schema">The input schema</param>
+        /// <returns>This DataFrameReader object</returns>
+        public DataFrameReader Schema(StructType schema)
+        {
+            JvmObjectReference jvmSchema = DataType.FromJson(_jvmObject.Jvm, schema.Json);
+            _jvmObject.Invoke("schema", jvmSchema);
```
imback82: Let's just do:
```C#
_jvmObject.Invoke("schema", DataType.FromJson(_jvmObject.Jvm, schema.Json);
```

src/csharp/Microsoft.Spark/Sql/Streaming/DataStreamReader.cs
```diff
@@ -30,6 +31,23 @@ public DataStreamReader Format(string source)
...
+        /// Some data sources (e.g. JSON) can infer the input schema automatically
+        /// from data. By specifying the schema here, the underlying data source can
+        /// skip the schema inference step, and thus speed up data loading.
+        /// </remarks>
+        /// <param name="schema">The input schema</param>
+        /// <returns>This DataStreamReader object</returns>
+        public DataStreamReader Schema(StructType schema)
+        {
+            JvmObjectReference jvmSchema = DataType.FromJson(_jvmObject.Jvm, schema.Json);
+            _jvmObject.Invoke("schema", jvmSchema);
```
imback82: ditto

src/csharp/Microsoft.Spark/Sql/Types/DataType.cs
suhsteve: Can you add the following xml tags as well in your description ?
```xml
<see cref="JvmObjectReference"/>
<see cref="StructType"/>
```

src/csharp/Microsoft.Spark/Sql/Streaming/DataStreamReader.cs
elvaliuliuliu: Updated.

src/csharp/Microsoft.Spark/Sql/Types/DataType.cs
elvaliuliuliu: Updated. Thanks.

src/csharp/Microsoft.Spark/Sql/Streaming/DataStreamReader.cs
elvaliuliuliu: Changed.

src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
elvaliuliuliu: Changed

src/csharp/Microsoft.Spark/Sql/Expressions/UserDefinedFunction.cs
```diff
@@ -34,10 +35,7 @@ internal sealed class UserDefinedFunction : IJvmObjectReferenceProvider
         {
             var pythonFunction = UdfUtils.CreatePythonFunction(jvm, command);
 
-            var dataType = (JvmObjectReference)jvm.CallStaticJavaMethod(
-                "org.apache.spark.sql.types.DataType",
-                "fromJson",
-                $"{returnType}");
+            var dataType = DataType.FromJson(jvm, returnType);
```
imback82: I think you need "{}" to form a json.

src/csharp/Microsoft.Spark/Sql/UDFRegistration.cs
```diff
@@ -269,10 +270,7 @@ internal void Register<TResult>(string name, Delegate func, UdfUtils.PythonEvalT
 
         private JvmObjectReference GetDataType<T>()
         {
-            return (JvmObjectReference)_jvmObject.Jvm.CallStaticJavaMethod(
-                "org.apache.spark.sql.types.DataType",
-                "fromJson",
-                $"{UdfUtils.GetReturnType(typeof(T))}");
+            return DataType.FromJson(_jvmObject.Jvm, UdfUtils.GetReturnType(typeof(T)));
```
imback82: ditto

examples/Microsoft.Spark.CSharp.Examples/Sql/Batch/Basic.cs
```diff
@@ -31,7 +32,12 @@ public void Run(string[] args)
 
             // Need to explicitly specify the schema since pickling vs. arrow formatting
             // will return different types. Pickling will turn longs into ints if the values fit.
-            DataFrame df = spark.Read().Schema("age INT, name STRING").Json(args[0]);
+            var dfSchema = new StructType(new[]
```
imback82:
``` C#
// Same as the "age INT, name STRING" DDL.
var inputSchema = ...
```

src/csharp/Microsoft.Spark/Sql/Expressions/UserDefinedFunction.cs
imback82: it's not `var` anymore. Either `JvmObjectReference dataType` or just put it directly below. Can you also fix `var pythonFunction` right above?

elvaliuliuliu: Sure, fixed both

examples/Microsoft.Spark.CSharp.Examples/Sql/Batch/Basic.cs
elvaliuliuliu: Changed. Thanks.

PR #251, opened 2019-09-19
Normalize AssemblyName

suhsteve:
This PR addresses normalizing AssemblyNames containing characters that will cause issues in the following scenarios:

- Using the AssemblyName as part of the filename is illegal.  ie, Windows does not allow some characters in the filename.
- Cause issues when attempting to use SparkContext.AddFile(AssemblyName.dll).

### DIFF
--- a/src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
+++ b/src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
@@ -201,12 +201,9 @@ namespace Microsoft.Spark.Utils
             // Roslyn assembly name: "\u211B*4b31b71b-d4bd-4642-9f63-eef5f5d99197#1-14"
             // Normalized Roslyn assembly name: "4b31b71b-d4bd-4642-9f63-eef5f5d99197-1-14"
             Match match = s_roslynAssemblyNameRegex.Match(assemblyName);
-            if (match.Success)
-            {
-                return $"{match.Groups[1].Value}-{match.Groups[2].Value}";
-            }
-
-            return assemblyName;
+            return match.Success ?
+                $"{match.Groups[1].Value}-{match.Groups[2].Value}" :
+                assemblyName;
         }
     }
 }
--- a/src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
+++ b/src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
@@ -113,7 +113,7 @@ namespace Microsoft.Spark.Utils
                     return assembly;
                 }
 
-                throw new FileNotFoundException($"Assembly '{assemblyName}' file not found: '{assemblyFileName}'");
+                throw new FileNotFoundException($"Assembly '{assemblyName}' file not found '{assemblyFileName}' in '{string.Join(",", s_searchPaths)}'");
             }
         }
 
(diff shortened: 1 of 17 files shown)


src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -169,5 +180,8 @@ private static bool TryLoadAssembly(string assemblyFileName, ref Assembly assemb
 
             return false;
         }
+
+        private static string NormalizeAssemblyName(string assemblyName) =>
```
suhsteve: Should we make this public?  We would then be able to use this in our repl middleware.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -90,6 +95,11 @@ internal static class AssemblyLoader
         /// found.</exception>
         internal static Assembly LoadAssembly(string assemblyName, string assemblyFileName)
         {
+            if (string.IsNullOrWhiteSpace(assemblyFileName))
```
suhsteve: This condition is present when we serialize UDF in the repl.

imback82: Can you add the comments to the code?

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -79,6 +80,10 @@ internal static class AssemblyLoader
 
         private static readonly object s_cacheLock = new object();
 
+        // Illegal characters: #, *, :, <, >, ", |, ?, /, \
```
imback82: Illegal characters for ..?

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
imback82: This should be fine since this is more like a temporary solution.

Can you add a comment why need to normalize assembly name (due to roslyn running inside repl, etc...)

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -169,5 +182,16 @@ private static bool TryLoadAssembly(string assemblyFileName, ref Assembly assemb
...
+
+        /// <summary>
+        /// Normalizes the assemblyName by removing characters known to cause
+        /// issues. This is useful in situations where the assemblyName is
+        /// automatically generated, ie the Roslyn compiler used in the REPL
+        /// generates an assembly name that contains * and #.
+        /// </summary>
+        /// <param name="assemblyName">Assembly name</param>
+        /// <returns>Normalized assembly name</returns>
+        private static string NormalizeAssemblyName(string assemblyName) =>
```
imback82: do we need to normalize the assembly name even if assembly file name is not used?

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -193,7 +195,23 @@ private static bool TryLoadAssembly(string assemblyFileName, ref Assembly assemb
...
+            // Roslyn assembly name: "\u211B*4b31b71b-d4bd-4642-9f63-eef5f5d99197#1-14"
+            // Normalized Roslyn assembly name: "4b31b71b-d4bd-4642-9f63-eef5f5d99197-1-14"
+            if (assemblyName.StartsWith("\u211B*")) {
+                Match match = s_roslynAssemblyNameRegex.Match(assemblyName);
+                if (match.Success)
+                {
+                    return $"{match.Groups[1].Value}-{match.Groups[2].Value}";
+                }
+
+                throw new FormatException(
```
imback82: I would just match this and return the original string if it doesn't match.

PR #255, opened 2019-09-23
Support for SparkFiles

suhsteve:
Support [SparkFiles.scala](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/SparkFiles.scala) . This is useful in determining the path of the files added through `SparkContext.addFile()`

### DIFF
--- a/src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
+++ b/src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
@@ -166,8 +166,13 @@ namespace Microsoft.Spark.Utils
         private static JvmObjectReference CreateEnvVarsForPythonFunction(IJvmBridge jvm)
         {
             JvmObjectReference environmentVars = jvm.CallConstructor("java.util.Hashtable");
-            string assemblySearchPath = Environment.GetEnvironmentVariable(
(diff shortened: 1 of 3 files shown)


src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
```diff
@@ -166,8 +166,14 @@ internal static JvmObjectReference CreatePythonFunction(IJvmBridge jvm, byte[] c
         private static JvmObjectReference CreateEnvVarsForPythonFunction(IJvmBridge jvm)
         {
             JvmObjectReference environmentVars = jvm.CallConstructor("java.util.Hashtable");
-            string assemblySearchPath = Environment.GetEnvironmentVariable(
-                AssemblySearchPathResolver.AssemblySearchPathsEnvVarName);
+            string assemblySearchPath = string.Join(",",
+                new[]
+                {
+                    SparkFiles.GetRootDirectory(),
+                    Environment.GetEnvironmentVariable(
```
imback82: can we put this first?

suhsteve: reordered.

Issue #259, opened 2019-09-25
[FEATURE REQUEST]: Add support for spark.Read().Jdbc()


rapoth:
**Is your feature request related to a problem? Please describe.**
While we provide spark.Write().Jdbc(), there is no equivalent spark.Read().Jdbc(). This makes for an inconsistent experience between Scala and C#.

**Scala**
```scala
spark
  .read
  .jdbc(jdbcUrl, "tablename", connectionProperties)
  .load()
```

**C#**
```csharp
spark
  .Read()
  .Format("jdbc")
  .Options(new Dictionary<string, string> 
                { 
                    {"url", ""}, 
                    {"dbtable", ""}, 
                    {"user", ""}, {"password", ""}
                })
  .Load();
```

**Describe the solution you'd like**
Expose the JDBC API

rapoth:
Closing the issue. Thanks @suhsteve!

PR #262, opened 2019-09-25
Support for DataFrameReader.Jdbc

suhsteve:
This PR provides support for the [Jdbc methods in DataFrameReader.scala](https://github.com/apache/spark/blob/d8b0914c2e0fdee72a3b9abb2d65283e22b6e8ee/sql/core/src/main/scala/org/apache/spark/sql/DataFrameReader.scala#L249).  Addresses issue #259


src/csharp/Microsoft.Spark/Utils/JvmObjectUtils.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Collections.Generic;
+using Microsoft.Spark.Interop;
+using Microsoft.Spark.Interop.Ipc;
+
+namespace Microsoft.Spark.Utils
+{
+    /// <summary>
+    /// JvmObjectUtils provides functions related to JVM objects.
+    /// </summary>
+    internal static class JvmObjectUtils
```
imback82: I think this name could be too generic. What do you want to achieve thru this class in a long run? Will this be more like a factory?

src/csharp/Microsoft.Spark/Utils/JvmObjectUtils.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// Create a java.util.Properties JVM object and populate the entries
+        /// using <paramref name="properties"/>
+        /// </summary>
+        /// <param name="properties">Dictionary used to populate the
+        /// java.util.Properties JVM object</param>
+        /// <returns>java.util.Properties JVM object</returns>
+        internal static JvmObjectReference CreateProperties(Dictionary<string, string> properties)
+        {
+            JvmObjectReference javaProperties =
+                SparkEnvironment.JvmBridge.CallConstructor("java.util.Properties");
```
imback82: Let's remove direct dependency on SparkEnvironment.

src/csharp/Microsoft.Spark/Utils/JvmObjectUtils.cs
imback82: How about putting  `jvm.CallConstructor("java.util.ArrayList");` and `jvm.CallConstructor("java.util.Hashtable");` here as well?

PR #266, opened 2019-09-25
Expose session related APIs for SparkSession.

imback82:
Expose session related APIs for `SparkSession` such `Active()`, `SetActiveSession()`, etc.

### DIFF
--- a/src/csharp/Microsoft.Spark/Interop/Scala/Option.cs
+++ b/src/csharp/Microsoft.Spark/Interop/Scala/Option.cs
@@ -1,0 +1,44 @@
+﻿using Microsoft.Spark.Interop.Ipc;
+
+namespace Microsoft.Spark.Interop.Scala
+{
+    /// <summary>
+    /// Exposes subset of scala.Option[T] APIs.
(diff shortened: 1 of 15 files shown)


src/csharp/Microsoft.Spark/Interop/Scala/Option.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Spark.Interop.Ipc;
+
+namespace Microsoft.Spark.Interop.Scala
```
suhsteve: Was it supposed to be `Microsoft.Spark.Interop.Scala` or `Microsoft.Spark.Internal.Scala`

imback82: It should have been `Microsoft.Spark.Interop.Internal.*` I moved the existing ones as well.

PR #273, opened 2019-09-27
Update release-0.5.md to reference 0.5.0 of the jars

GoEddie:
The jar version wasn't incremented in the readme


GoEddie:
No probs, glad I could help :)

PR #275, opened 2019-09-28
Update SparkFiles to be callable by the Worker.

suhsteve:
The SparkFiles root path is sent to the Microsoft.Spark.Worker as part of its payload.  On the worker, use the payload to set and use this path as part of its assembly probing paths.

### DIFF
--- a/src/csharp/Microsoft.Spark.Worker/Processor/PayloadProcessor.cs
+++ b/src/csharp/Microsoft.Spark.Worker/Processor/PayloadProcessor.cs
@@ -56,6 +56,7 @@ namespace Microsoft.Spark.Worker.Processor
             payload.Version = SerDe.ReadString(stream);
             payload.TaskContext = new TaskContextProcessor(_version).Process(stream);
             payload.SparkFilesDir = SerDe.ReadString(stream);
+            SparkFiles.SetRootDirectory(payload.SparkFilesDir);
 
             if (Utils.SettingUtils.IsDatabricks)
             {
--- a/src/csharp/Microsoft.Spark/SparkFiles.cs
+++ b/src/csharp/Microsoft.Spark/SparkFiles.cs
@@ -2,33 +2,62 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
+using System.IO;
 using Microsoft.Spark.Interop;
 using Microsoft.Spark.Interop.Ipc;
 
 namespace Microsoft.Spark
 {
     /// <summary>
-    /// Resolves paths to files added through `SparkContext.addFile()`.
+    /// Resolves paths to files added through
+    /// <c>SparkContext.addFile()</c>.
     /// </summary>
     public static class SparkFiles
     {
         private static IJvmBridge Jvm { get; } = SparkEnvironment.JvmBridge;
         private static readonly string s_sparkFilesClassName = "org.apache.spark.SparkFiles";
 
-        /// <summary>
-        /// Get the absolute path of a file added through `SparkContext.addFile()`.
-        /// </summary>
-        /// <param name="fileName">The name of the file added through `SparkContext.addFile()`
-        /// </param>
-        /// <returns>The absolute path of the file.</returns>
-        public static string Get(string fileName) =>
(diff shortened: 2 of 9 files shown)


src/csharp/Microsoft.Spark.Worker/Processor/PayloadProcessor.cs
```diff
@@ -56,6 +56,7 @@ internal Payload Process(Stream stream)
             payload.Version = SerDe.ReadString(stream);
             payload.TaskContext = new TaskContextProcessor(_version).Process(stream);
             payload.SparkFilesDir = SerDe.ReadString(stream);
+            SparkFiles.SetRootDirectory(payload.SparkFilesDir);
```
imback82: Does this even work in cluster mode? There is no listener on the worker nodes. You should just leverage the Python env variables used in registering the UDF.

suhsteve: This was how python did it.  This should set the private variables on the worker.   When the worker calls SparkFIles.getRootDirectory, it will won't call the DotnetBackend to fulfill the request, but use the private variables that were set.

src/csharp/Microsoft.Spark/SparkFiles.cs
```diff
@@ -2,33 +2,62 @@
...
+using System;
+using System.IO;
 using Microsoft.Spark.Interop;
 using Microsoft.Spark.Interop.Ipc;
 
 namespace Microsoft.Spark
 {
     /// <summary>
-    /// Resolves paths to files added through `SparkContext.addFile()`.
+    /// Resolves paths to files added through
```
imback82: put these in a same line?

src/csharp/Microsoft.Spark/SparkFiles.cs
```diff
@@ -2,33 +2,62 @@
...
     public static class SparkFiles
     {
         private static IJvmBridge Jvm { get; } = SparkEnvironment.JvmBridge;
         private static readonly string s_sparkFilesClassName = "org.apache.spark.SparkFiles";
 
+        [ThreadStatic]
+        private static string s_rootDirectory;
+
+        [ThreadStatic]
+        private static bool s_isWorker;
```
imback82: s_isRunningOnWorker

src/csharp/Microsoft.Spark.Worker/Processor/PayloadProcessor.cs
imback82: Ah OK. This should be fine, but not sure which approach is cleaner. Now, you have to do more stuff to set up the tests, etc.

src/csharp/Microsoft.Spark/SparkFiles.cs
```diff
@@ -2,33 +2,62 @@
...
+
         /// <summary>
-        /// Get the absolute path of a file added through `SparkContext.addFile()`.
+        /// Get the absolute path of a file added through
+        /// <c>SparkContext.addFile()</c>.
```
imback82: any reason this is on a new line?

imback82: same in other places.

suhsteve: no significant reason, Just hit max line limit and thought it may look nicer if it wasn't too long.  Reverted.

src/csharp/Microsoft.Spark.Worker/Processor/PayloadProcessor.cs
suhsteve: I've tested on local mode.  I'll still have to test this out on cluster mode which may take some time.

imback82: Oh I meant changes in unit tests.

suhsteve: Meant to address your earlier comment regarding if this was tested in Cluster Mode.

suhsteve:
Let's revisit this when there's a need to call SparkFiles from the worker.  We'll keep the current implementation that's currently in `master` for now.

PR #277, opened 2019-10-01
VectorUdf with DataFrame + Arrow

pgovind:
Minor changes to the UDF API to pass in and return corefxlab DataFrames
Accompanying unit test changes 

Putting it up here to get initial thoughts. FxDataFrame's Arrow support means true zero copy exchange of data. 

@eerhardt @rapoth @imback82 . Not able to add reviewers for some reason :/

Can be reviewed now!


src/csharp/Microsoft.Spark.Worker.UnitTest/CommandExecutorTests.cs
```diff
@@ -522,14 +516,13 @@ Int64Array ConvertInt64s(Int64Array int64s)
...
-                    resultSchema,
-                    new IArrowArray[]
-                    {
-                        ConvertStrings((StringArray)batch.Column(0)),
-                        ConvertInt64s((Int64Array)batch.Column(1)),
-                    },
-                    batch.Length));
+                (dataFrame) =>
+                {
+                    StringArray strings = ConvertStrings(dataFrame.Column(0));
```
pgovind: This is because FxDataFrame.ArrowStringColumn is immutable.

src/csharp/Microsoft.Spark.Worker.UnitTest/CommandExecutorTests.cs
```diff
@@ -601,10 +594,10 @@ Int64Array ConvertInt64s(Int64Array int64s)
                     Assert.Equal($"udf: {i}", stringArray.GetString(i));
                 }
 
-                var longArray = (Int64Array)outputBatch.Column(1);
+                var doubleArray = (DoubleArray)outputBatch.Column(1);
```
pgovind: long -> double occurs in the line 523 above. `dataFrame.Column(1) + 100` returns a double column.

eerhardt:
Looks like a test is failing:

```
[xUnit.net 00:00:35.73]     Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestGroupedMapUdf [FAIL]
  X Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestGroupedMapUdf [1s 347ms]
  Error Message:
   Assert.Equal() Failure
Expected: 3
Actual:   5
```

examples/Microsoft.Spark.CSharp.Examples/Sql/VectorUdfs.cs
```diff
@@ -50,45 +53,28 @@ public void Run(string[] args)
...
-                string current = stringValues.GetString(i);
-                characterCount += current.Length;
+                characterCount += ((string)dataFrame[stringFieldName][i]).Length;
+                characterCountColumn[i] = characterCount;
             }
 
-            int groupFieldIndex = records.Schema.GetFieldIndex(groupFieldName);
-            Field groupField = records.Schema.GetFieldByIndex(groupFieldIndex);
-
-            // Return 1 record, if we were given any. 0, otherwise.
```
eerhardt: This isn't doing what it was originally doing.

Originally it would only return 1 record with 2 columns: "Age" and "name_CharCount".

However, you are now return N records, where N is the number of records passed in to the UDF.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -26,16 +26,14 @@
...
     <PackageReference Include="Apache.Arrow" Version="0.14.1" />
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
+    <PackageReference Include="Microsoft.Data.DataFrame" Version="0.1.1-e190930-1" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
     <PackageReference Include="Razorvine.Pyrolite" Version="4.26.0" />
     <PackageReference Include="System.Memory" Version="4.5.2" />
   </ItemGroup>
 
   <ItemGroup>
-    <Content Include="..\..\scala\microsoft-spark-*\target\microsoft-spark-*.jar"
```
eerhardt: Should revert this.

src/csharp/Microsoft.Spark.Worker.UnitTest/CommandExecutorTests.cs
eerhardt: That seems wrong. `PrimitiveColumn<long>` + `int` should return a `long`.  Just like how:

```C#
long x = 0;
var y = x + 5;

y.GetType()
```

returns `System.Int64`.

---
In reply to: [330316160](https://github.com/dotnet/spark/pull/277#discussion_r330316160) [](ancestors = 330316160)

examples/Microsoft.Spark.CSharp.Examples/Sql/VectorUdfs.cs
```diff
@@ -8,6 +8,9 @@
 using Microsoft.Spark.Sql;
 using Microsoft.Spark.Sql.Types;
 using StructType = Microsoft.Spark.Sql.Types.StructType;
+using FxDataFrame = Microsoft.Data.DataFrame;
```
eerhardt: @imback82 @rapoth - do either of you have any thoughts/opinions on how to make this better? Using 2 types named `DataFrame` in Spark seems unfortunate. Is there a better term we can call the corefx `DataFrame` class?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -534,18 +535,23 @@ internal class ArrowGroupedMapCommandExecutor : SqlCommandExecutor
             ArrowStreamWriter writer = null;
             foreach (RecordBatch input in GetInputIterator(inputStream))
             {
-                RecordBatch result = worker.Func(input);
+                FxDataFrame dataFrame = new FxDataFrame(input);
```
eerhardt: This looks like I'd expect. There should be zero-copying here, right?

I expect a similar change to the `ArrowSqlCommandExecutor` above.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -266,36 +269,18 @@ public void TestGroupedMapUdf()
...
-                string current = stringValues.GetString(i);
-                characterCount += current.Length;
+                characterCount += ((string)dataFrame["name"][i]).Length;
+                characterCountColumn[i] = characterCount;
             }
-
-            int groupFieldIndex = records.Schema.GetFieldIndex("age");
-            Field groupField = records.Schema.GetFieldByIndex(groupFieldIndex);
-
-            // Return 1 record, if we were given any. 0, otherwise.
```
eerhardt: Similar comment as the Examples VectorUdfs.cs file. (since this is duplicated code)

src/csharp/Microsoft.Spark.Worker.UnitTest/CommandExecutorTests.cs
pgovind: Would you consider this something we HAVE to change? For example, `long` + `float` results in a `float` because the compiler knows to make the conversions. The DataFrame doesn't attempt to convert because each column's `Memory` has to be cloned in the conversion. For a chain of operations, this seems wasteful? Instead it converts once to a `PrimitiveColumn<double>`(or `PrimitiveColumn<decimal>`) when there is a type mismatch so all subsequent operations can work without cloning. 

The other alternative I'd considered was converting only when the underlying `Memory` had to be changed. `PrimitiveColumn<long>` + `int` would involve no conversions for example. I didn't think we should duplicate all the conversion logic in the compiler however, so I defaulted to what we have now. Do you know of an easier/better way to do this?

eerhardt: I find it very surprising that adding, subtracting, or multiplying two integers would result in a floating-point number. IMO, I think that is something that should be changed.

But I'm not all customers, so maybe getting more thoughts/opinions here would be a good idea. But matching normal C# behavior seems like it is reasonable default behavior.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
pgovind: Done

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
pgovind: Yup, this is zero-copy.

benchmark/csharp/Tpch/VectorFunctions.cs
```diff
@@ -3,49 +3,58 @@
...
 
-            int length = price.Length;
-            var builder = new DoubleArray.Builder().Reserve(length);
-            ReadOnlySpan<double> prices = price.Values;
-            ReadOnlySpan<double> discounts = discount.Values;
-            ReadOnlySpan<double> taxes = tax.Values;
-            for (int i = 0; i < length; ++i)
+            PrimitiveColumn<double> ret = new PrimitiveColumn<double>("Prices", price.Length);
+
+            foreach (ReadOnlyMemory<double> readOnlyTax in tax.GetReadOnlyDataBuffers())
```
eerhardt: These nested foreach statements aren't really correct, are they? If there was more than 1 buffer, you wouldn't get the correct answer.

pgovind: Yup, this is wrong. Part of why I marked it WIP :). CI passes now, so I'm cleaning it all up.

examples/Microsoft.Spark.CSharp.Examples/Sql/VectorUdfs.cs
```diff
@@ -56,39 +58,27 @@ public void Run(string[] args)
...
-            int returnLength = records.Length > 0 ? 1 : 0;
-
-            return new RecordBatch(
-                new Schema.Builder()
-                    .Field(groupField)
-                    .Field(f => f.Name(stringFieldName + "_CharCount").DataType(Int32Type.Default))
-                    .Build(),
-                new IArrowArray[]
-                {
-                    records.Column(groupFieldIndex),
```
pgovind: RecordBatch doesn't allow variable length ArrowArrays right? `records.Column(groupFieldIndex)` could have a Length of more than 1 here?

eerhardt: RecordBatch allows it. It doesn't enforce that each array is of the same length. So you can create 2 columns:

1. Length 10
2. Length 5

And wrap a RecordBatch around them and say the RecordBatch's length is `5`. This allows you to reuse buffers/arrays without copying data just to truncate the length.

I'm not sure if this is great behavior or not, but it is how it works currently. I can look how the C++ and Java Arrow libraries work w.r.t. this.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -72,6 +74,35 @@ internal abstract class SqlCommandExecutor
...
+                    // in order to create and write back the result schema.
+
+                    int columnCount = reader.Schema.Fields.Count;
+                    var arrays = new IArrowArray[columnCount];
+                    for (int i = 0; i < columnCount; ++i)
+                    {
+                        IArrowType type = reader.Schema.GetFieldByIndex(i).DataType;
+                        arrays[i] = ArrowArrayHelpers.CreateEmptyArray(type);
+                    }
+                    yield return new RecordBatch(reader.Schema, arrays, 0);
```
suhsteve:
```suggestion

                    yield return new RecordBatch(reader.Schema, arrays, 0);
```

src/csharp/Microsoft.Spark/Sql/ArrowArrayHelpers.cs
```diff
@@ -38,110 +39,84 @@ internal static class ArrowArrayHelpers
...
-
-        private static ArrayData BuildEmptyArrayDataFromArrowType(IArrowType arrowType)
-        {
-            if (s_twoBufferArrowTypes.Contains(arrowType.TypeId))
+            BaseColumn ret;
+            if (typeof(T) == typeof(PrimitiveColumn<bool>))
             {
-                return new ArrayData(arrowType, 0,
-                    buffers: new[] { ArrowBuffer.Empty, ArrowBuffer.Empty });
+                ret = new PrimitiveColumn<bool>("Empty");
```
suhsteve: Is the `"Empty"` column name something that will be checked in the future to validate that this designates an empty column ?  If so should we pull this name out and define it as some kind of constant ?

benchmark/csharp/Tpch/VectorFunctions.cs
```diff
@@ -3,49 +3,70 @@
...
+
+            IEnumerator<ReadOnlyMemory<double>> pricesEnumerator = readOnlyPrices.GetEnumerator();
+            IEnumerator<ReadOnlyMemory<double>> discountsEnumerator = readOnlyDiscounts.GetEnumerator();
+
+            while (pricesEnumerator.MoveNext() && discountsEnumerator.MoveNext())
             {
-                builder.Append(prices[i] * (1 - discounts[i]));
+                ReadOnlySpan<double> prices = pricesEnumerator.Current.Span;
+                ReadOnlySpan<double> discounts = discountsEnumerator.Current.Span;
+                for (int i = 0; i < prices.Length; ++i)
```
imback82: Is it possible to do this one loop?
Ideally, 
```
while (prices.MoveNext() && discounts.MoveNext())
{
    ret[i] = (prices.Current() * (1 - discounts.Current()));
}
```
It may be less performant b/c of all the function calls, but it will be good to have for the user experience?

The most ideal user experience will be:
```C#
var ret = prices * (1 - discounts);
```
similar to Pandas. I am curious whether this can be done. :)
```

Comment (imback82) [inline benchmark/csharp/Tpch/VectorFunctions.cs]: Is it possible to do this one loop?
For example,
```
while (prices.MoveNext() && discounts.MoveNext())
{
    ret[i] = (prices.Current() * (1 - discounts.Current()));
}
```
It may be less performant b/c of all the function calls, but it will be good to have for the user experience?

The most ideal user experience will be:
```C#
var ret = prices * (1 - discounts);
```
similar to Pandas. I am curious whether this can be done. :)
```

Comment (pgovind) [reply benchmark/csharp/Tpch/VectorFunctions.cs]: > The most ideal user experience will be:

We can. I'll update the PR with that syntax

File: benchmark/csharp/Tpch/VectorFunctions.cs [hunk 13]
```diff
@@ -3,49 +3,70 @@
...
+
+            while (taxesEnumerator.MoveNext() && pricesEnumerator.MoveNext() && discountsEnumerator.MoveNext())
             {
-                builder.Append(prices[i] * (1 - discounts[i]) * (1 + taxes[i]));
+                ReadOnlySpan<double> taxes = taxesEnumerator.Current.Span;
+                ReadOnlySpan<double> prices = pricesEnumerator.Current.Span;
+                ReadOnlySpan<double> discounts = discountsEnumerator.Current.Span;
+                for (int i = 0; i < prices.Length; ++i)
+                {
+                    ret[i] = (prices[i] * (1 - discounts[i]) * (1 + taxes[i]));
```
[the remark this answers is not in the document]
pgovind: Same reply as to Terry. This exact syntax will work. I'll update the PR :)

src/csharp/Microsoft.Spark/Sql/ArrowArrayHelpers.cs
pgovind: No, `Empty` is not a special name. I just put it here to make it readable. Maybe I should change it to "Return" to avoid confusion

benchmark/csharp/Tpch/VectorFunctions.cs
pgovind: Updated

eng/Versions.props
```diff
@@ -6,6 +6,7 @@
     <RestoreSources>
       $(RestoreSources);
       https://api.nuget.org/v3/index.json;
+      https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json;
```
pgovind: This can be removed

examples/Microsoft.Spark.CSharp.Examples/Sql/VectorUdfs.cs
```diff
@@ -3,11 +3,13 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using Apache.Arrow;
-using Apache.Arrow.Types;
 using Microsoft.Spark.Sql;
 using Microsoft.Spark.Sql.Types;
 using StructType = Microsoft.Spark.Sql.Types.StructType;
+using FxDataFrame = Microsoft.Data.Analysis.DataFrame;
```
pgovind: Same question that @eerhardt posed: 

@imback82  @rapoth - do either of you have any thoughts/opinions on how to make this better? Using 2 types named DataFrame in Spark seems unfortunate. Is there a better term we can call the corefx DataFrame class?

pgovind: We should think about this sooner rather than later. For ex: In the last .NET Town Hall, Spark Notebooks was demo'd and Dan had a question of whether the `DataFrame` he saw in the sample was a Spark `DataFrame` or the corefx `DataFrame`.

pgovind:
Updated to address comments and use the Microsoft.Data.Analysis package on Nuget

benchmark/csharp/Tpch/VectorFunctions.cs
```diff
@@ -3,49 +3,31 @@
...
-            ReadOnlySpan<double> prices = price.Values;
-            ReadOnlySpan<double> discounts = discount.Values;
-            ReadOnlySpan<double> taxes = tax.Values;
-            for (int i = 0; i < length; ++i)
-            {
-                builder.Append(prices[i] * (1 - discounts[i]) * (1 + taxes[i]));
-            }
-
-            return builder.Build();
+            return (PrimitiveDataFrameColumn<double>)(price * (1 - discount) * (1 + tax));
```
eerhardt: Do we not have operator overloads for `PrimitiveDataFrameColumn<T>` that return the same type? I think that would make this code simpler so the user doesn't have to cast.

examples/Microsoft.Spark.CSharp.Examples/Sql/Batch/VectorUdfs.cs
```diff
@@ -56,39 +58,28 @@ public void Run(string[] args)
...
             int characterCount = 0;
 
-            for (int i = 0; i < stringValues.Length; ++i)
+            PrimitiveDataFrameColumn<int> characterCountColumn = new PrimitiveDataFrameColumn<int>(stringFieldName + "CharCount");
+            PrimitiveDataFrameColumn<int> ageColumn = new PrimitiveDataFrameColumn<int>(groupFieldName);
+            for (long i = 0; i < dataFrame.RowCount; i++)
             {
-                string current = stringValues.GetString(i);
-                characterCount += current.Length;
+                characterCount += ((string)dataFrame[stringFieldName][i]).Length;
```
eerhardt: Can you cache `dataFrame[stringFieldName]` into a variable, and then you don't need to cast in the loop.

eerhardt:
>                 // Arrow array types

Do we still need these?

---
Refers to: src/csharp/Microsoft.Spark/Utils/UdfUtils.cs:97 in 5295123. [](commit_id = 5295123347218bb898d594f9022c46c276456b98, deletion_comment = False)

src/csharp/Microsoft.Spark.Experimental/Sql/ExperimentalFunctions.cs
```diff
@@ -23,8 +23,8 @@ public static class ExperimentalFunctions
         /// A delegate that returns a <see cref="Column"/> for the result of the Vector UDF.
         /// </returns>
         public static Func<Column, Column> VectorUdf<T, TResult>(Func<T, TResult> udf)
-            where T : IArrowArray
-            where TResult : IArrowArray
+            where T : DataFrameColumn
```
imback82: @pgovind Is is possible to introduce this as new APIs instead of changing the existing ones? We have few customers using this and it would be an easier transition for them. What do you think? Would it be too much work?

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -26,6 +26,7 @@
   <ItemGroup>
     <PackageReference Include="Apache.Arrow" Version="0.14.1" />
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
+    <PackageReference Include="Microsoft.Data.Analysis" Version="0.1.0" />
```
imback82: Should we now update this to 0.2.0?

pgovind:
Just rebased on current master. Will address the comments soon

benchmark/csharp/Tpch/VectorFunctions.cs
pgovind: No unfortunately. I'll file a bug.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
pgovind: Yup. Updated

src/csharp/Microsoft.Spark.Experimental/Sql/RelationalGroupedDatasetExtensions.cs
```diff
@@ -36,7 +37,7 @@ public static class RelationalGroupedDatasetExtensions
         public static DataFrame Apply(
             this RelationalGroupedDataset dataset,
             StructType returnType, 
-            Func<RecordBatch, RecordBatch> func)
+            Func<FxDataFrame, FxDataFrame> func)
```
pgovind: Do we have concerns about breaking this API?

imback82: Yes, can we keep this side by side to give users some time for migration to new APIs?

src/csharp/Microsoft.Spark.Experimental/Sql/ExperimentalFunctions.cs
pgovind: I have this working on my machine now as new APIs. Cleaning it up now and putting it up for review in a bit

pgovind:
> Do we still need these?

There's 1 unit test that depends on those values being in the dictionary. Now that we have DataFrameUdfUtils.cs, we should be able to remove them from UdfUtils.cs. Once CI succeeds, I'll try removing them from UdfUtils.cs to see if I break anything.

examples/Microsoft.Spark.CSharp.Examples/Sql/Batch/VectorUdfs.cs
```diff
@@ -3,11 +3,13 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using Apache.Arrow;
-using Apache.Arrow.Types;
 using Microsoft.Spark.Sql;
 using Microsoft.Spark.Sql.Types;
 using StructType = Microsoft.Spark.Sql.Types.StructType;
+using FxDataFrame = Microsoft.Data.Analysis.DataFrame;
```
pgovind: Sort usings

src/csharp/Microsoft.Spark.Experimental/Sql/RelationalGroupedDatasetExtensions.cs
pgovind: Done.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -51,18 +53,51 @@ internal abstract class SqlCommandExecutor
                 throw new ArgumentException("Unexpected serialization mode found.");
             }
 
+            bool useDataFrameCommandExecutor = false;
+            foreach (SqlCommand command in commands)
+            {
+                WorkerFunction workerFunc = command.WorkerFunction;
+                DataFrameWorkerFunction dataFrameWorkerFunction = workerFunc as DataFrameWorkerFunction;
```
pgovind: I want to get feedback about this `if` pattern first. If it's acceptable, then I can address a bug where a user could create a `CommandPayload` defined with both `DataFrameWorkerFunction` and `ArrowSqlWorkerFunction`.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -26,6 +26,7 @@
   <ItemGroup>
     <PackageReference Include="Apache.Arrow" Version="0.14.1" />
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
+    <PackageReference Include="Microsoft.Data.Analysis" Version="0.2.0" />
```
pgovind: Is the version number important here? If I stop specifying it, will we always look for the latest version or do we stop looking if an older version is found in a local Nuget cache?

src/csharp/Microsoft.Spark/Sql/WorkerFunction.DataFrame.cs
```diff
@@ -0,0 +1,105 @@
...
+
+using System;
+using Apache.Arrow;
+using Microsoft.Data.Analysis;
+using FxDataFrame = Microsoft.Data.Analysis.DataFrame;
+
+namespace Microsoft.Spark.Sql
+{
+    ///// <summary>
+    ///// Function that will be executed in the worker.
```
pgovind: Remove commented out code

examples/Microsoft.Spark.CSharp.Examples/Sql/Batch/VectorUdfs.cs
```diff
@@ -56,39 +58,29 @@ public void Run(string[] args)
...
-                    .Field(groupField)
-                    .Field(f => f.Name(stringFieldName + "_CharCount").DataType(Int32Type.Default))
-                    .Build(),
-                new IArrowArray[]
-                {
-                    records.Column(groupFieldIndex),
-                    new Int32Array.Builder().Append(characterCount).Build()
-                },
-                returnLength);
+            FxDataFrame ret = new FxDataFrame(new List<DataFrameColumn> { ageColumn, characterCountColumn });
```
pgovind: Can pass the columns in without a list

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
pgovind: Similar pattern in `CommandProcessor.cs`

pgovind:
\azp run

pgovind:
I keep getting this error in CI(and locally):

Caused by: java.lang.AssertionError: assertion failed: Invalid schema from pandas_udf: expected ArrayType(StringType,true), got StringType

I can't seem to figure out why pandas_udf is getting `StringType` when the DataFrame being returned clearly has 3 rows in it. I tested that RecordBatch roundtripping works on the DataFrame side, and tested that ArrowStreamWriter/Reader roundtripping works in Arrow, so maybe the bug/my error is somewhere else? Hoping @eerhardt can help me with this.

pgovind:
Fixed the CI errors(thanks to @eerhardt ). This can be reviewed now.

benchmark/csharp/Tpch/TpchFunctionalQueries.cs
```diff
@@ -62,11 +65,19 @@ internal void Q1()
 
         internal void Q1v()
         {
+#if NETCOREAPP
```
eerhardt: Why `#if NETCOREAPP` here?

benchmark/csharp/Tpch/VectorFunctions.cs
```diff
@@ -3,49 +3,30 @@
...
```
eerhardt: Do we need to cast here? Doing match on a double column should return a double column, right?

examples/Microsoft.Spark.CSharp.Examples/Sql/Batch/VectorUdfs.cs
```diff
@@ -56,39 +58,29 @@ public void Run(string[] args)
...
         {
-            int stringFieldIndex = records.Schema.GetFieldIndex(stringFieldName);
-            StringArray stringValues = records.Column(stringFieldIndex) as StringArray;
-
             int characterCount = 0;
 
-            for (int i = 0; i < stringValues.Length; ++i)
+            PrimitiveDataFrameColumn<int> characterCountColumn = new PrimitiveDataFrameColumn<int>(stringFieldName + "CharCount");
+            PrimitiveDataFrameColumn<int> ageColumn = new PrimitiveDataFrameColumn<int>(groupFieldName);
+            ArrowStringDataFrameColumn fieldColumn = dataFrame[stringFieldName] as ArrowStringDataFrameColumn;
```
eerhardt: (nit) what does `fieldColumn` mean? Why that name?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -266,36 +346,26 @@ public void TestGroupedMapUdf()
             }
         }
 
-        private static RecordBatch CountCharacters(RecordBatch records)
+        private static FxDataFrame CountCharacters(FxDataFrame dataFrame, string groupFieldName, string stringFieldName)
```
eerhardt: Do we still want the RecordBatch test as well? Might as well test it, as long as it is supported.

src/csharp/Microsoft.Spark/Sql/ArrowUdfWrapper.DataFrame.cs
```diff
@@ -0,0 +1,463 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
eerhardt: What do you think of naming this file `DataFrameUdfWrapper.cs` instead of `ArrowUdfWrapper.DataFrame.cs`?

src/csharp/Microsoft.Spark.Worker/Processor/CommandProcessor.cs
```diff
@@ -157,18 +157,29 @@ private static RDDCommand ReadRDDCommand(Stream stream)
...
+                            }
+                            else if (obj is DataFrameWorkerFunction.ExecuteDelegate dataFrameWorkerFunctionDelegate)
+                            {
+                                var curWorkerFunction = new DataFrameWorkerFunction(dataFrameWorkerFunctionDelegate);
+                                command.WorkerFunction = (command.WorkerFunction == null) ?
+                                curWorkerFunction :
+                                DataFrameWorkerFunction.Chain(
+                                    (DataFrameWorkerFunction)command.WorkerFunction,
+                                    curWorkerFunction);
+                            }
```
eerhardt: else throw?  Here and in the grouped map section below.

src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
```diff
@@ -103,6 +104,15 @@ internal enum PythonEvalType
...
                 {typeof(BinaryArray), "binary"},
+
+                {typeof(PrimitiveDataFrameColumn<bool>), "boolean"},
+                {typeof(PrimitiveDataFrameColumn<byte>), "byte"},
+                {typeof(PrimitiveDataFrameColumn<short>), "short"},
+                {typeof(PrimitiveDataFrameColumn<int>), "integer"},
+                {typeof(PrimitiveDataFrameColumn<long>), "long"},
+                {typeof(PrimitiveDataFrameColumn<float>), "float"},
+                {typeof(PrimitiveDataFrameColumn<double>), "double"},
+                {typeof(ArrowStringDataFrameColumn), "string"},
```
eerhardt: what about `binary`?

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -294,10 +342,20 @@ private static Delegate CreateUdfWrapperDelegate<T>(Type type, object[] paramete
                 parameters,
                 null);
 
-            return Delegate.CreateDelegate(
-                typeof(T),
-                udfWrapper,
-                UdfWrapperMethodName);
+            if (type == typeof(DataFrameGroupedMapUdfWrapper))
```
eerhardt: Why is this needed? Why just this one type?

src/csharp/Microsoft.Spark.Experimental/Sql/RelationalGroupedDatasetExtensions.cs
```diff
@@ -33,6 +34,14 @@ public static class RelationalGroupedDatasetExtensions
...
         /// <returns>New DataFrame object with the UDF applied.</returns>
+        public static DataFrame Apply(
+            this RelationalGroupedDataset dataset,
+            StructType returnType, 
+            Func<FxDataFrame, FxDataFrame> func)
+        {
+            return dataset.Apply(returnType, func);
+        }
+
         public static DataFrame Apply(
```
eerhardt: Should we have xml docs comments for this method?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
eerhardt: I wonder if this wouldn't be better lower in the stack? For example, using the `ArrowSqlCommandExecutor` for both cases, and then in `ICommandRunner commandRunner = CreateCommandRunner(commands);` you can look at the command and decide whether to go the "data frame" route or the "arrow array" route.

Note: if you keep the current design, I would refactor this chunk out into a separate method.

benchmark/csharp/Tpch/VectorFunctions.cs
pgovind: Yea, we need a cast unfortunately. The math operators are defined on the base `DataFrameColumn` class and the operator overloads return `DataFrameColumn`. We have https://github.com/dotnet/corefxlab/issues/2822 filed to improve this. When I think about it now, maybe they should've been defined in `PrimitiveDataFrameColumn<T>` instead. No other column can do math anyway. We could roll with this for now and modify it once we fix 2822

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
pgovind: Yup, good catch. I must've missed it in my review. I'll add it.

benchmark/csharp/Tpch/TpchFunctionalQueries.cs
pgovind: Since this is benchmark code, I assumed we'd want to retain the vectorized methods for core? (VectorFunctions.Intrinsics.cs exposes a vectorized `ComputeTotal` method). Since `DataFrame` is not vectorized yet, I enabled the DataFrame APIs only for framework(which was using Arrow APIs before - see changes to VectorFunctions.cs).

examples/Microsoft.Spark.CSharp.Examples/Sql/Batch/VectorUdfs.cs
pgovind: Hmm, I don't remember why I chose that anymore :). Fixed

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
pgovind: Yup, good catch. I must've missed it in my review. Added

src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
pgovind: We actually don't support binary yet in DataFrame.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
pgovind: Not needed anymore now that `DeserializeArrowOrDataFrameUdf` is passing in the right types. Deleted.

src/csharp/Microsoft.Spark.Experimental/Sql/RelationalGroupedDatasetExtensions.cs
pgovind: Yup. Fixed.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -216,11 +216,51 @@ private sealed class UdfWrapperData
...
+            var udfWrapperData = GetUdfWrapperDataFromStream(stream, out serializerMode, out deserializerMode, out runMode);
+
+            var nodeIndex = 0;
+            var udfIndex = 0;
+            UdfWrapperNode node = udfWrapperData.UdfWrapperNodes[nodeIndex];
+            var nodeType = Type.GetType(node.TypeName);
+            Delegate udf = null;
+            if (nodeType == typeof(DataFrameGroupedMapUdfWrapper))
+            {
+                udf = (DataFrameGroupedMapWorkerFunction.ExecuteDelegate)DeserializeUdfs<DataFrameGroupedMapWorkerFunction.ExecuteDelegate>(udfWrapperData, ref nodeIndex, ref udfIndex);
```
imback82: Can you help making the repo consistent with max char limit by following https://github.com/dotnet/spark/blob/master/docs/coding-guidelines/csharp-coding-style.md? Thanks!

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -216,11 +216,51 @@ private sealed class UdfWrapperData
...
+            }
+            else if (nodeType == typeof(DataFrameWorkerFunction))
+            {
+                udf = (DataFrameWorkerFunction.ExecuteDelegate)DeserializeUdfs<DataFrameWorkerFunction.ExecuteDelegate>(udfWrapperData, ref nodeIndex, ref udfIndex);
+            }
+            else
+            {
+                if (nodeType.IsSubclassOf(typeof(DataFrameUdfWrapper)))
```
imback82:
```C#
else if ((nodeType == typeof(DataFrameWorkerFunction) || (nodeType.IsSubclassOf(typeof(DataFrameUdfWrapper))))
```

imback82:
@suhsteve can you also help reviewing this change when you get a chance? Thanks! (Note that we want to support both Arrow APIs and Fx.DataFrame APIs for the time being.)

src/csharp/Microsoft.Spark/Sql/Functions.cs
```diff
@@ -4071,7 +4071,7 @@ private static UserDefinedFunction CreateUdf<TResult>(string name, Delegate exec
             return CreateUdf<TResult>(name, execute, UdfUtils.PythonEvalType.SQL_BATCHED_UDF);
         }
 
-        private static UserDefinedFunction CreateVectorUdf<TResult>(string name, Delegate execute)
+        internal static UserDefinedFunction CreateVectorUdf<TResult>(string name, Delegate execute)
```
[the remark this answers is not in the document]
pgovind: ExperimentalDataFrameFunctions calls this in ExperimentalFunctions.DataFrame.cs

src/csharp/Microsoft.Spark/Utils/UdfUtils.DataFrame.cs
```diff
@@ -0,0 +1,155 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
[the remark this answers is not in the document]
pgovind: Nope. Renamed to `DataFrameUdfUtils.cs`

src/csharp/Microsoft.Spark/Sql/ArrowArrayHelpers.cs
```diff
@@ -38,10 +39,61 @@ internal static class ArrowArrayHelpers
...
 
-        public static IArrowArray CreateEmptyArray<T>()
+        public static DataFrameColumn CreateEmptyColumn<T>()
         {
-            ArrayData data = BuildEmptyArrayDataFromArrayType<T>();
-            return ArrowArrayFactory.BuildArray(data);
+            DataFrameColumn ret;
+            if (typeof(T) == typeof(PrimitiveDataFrameColumn<bool>))
+            {
+                ret = new PrimitiveDataFrameColumn<bool>("Empty");
```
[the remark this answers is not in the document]
pgovind: You mean you'd like to see `return PrimitiveDataFrameColumn<bool>("Empty")` directly here instead of setting `ret` and returning it at the end of the method? I'm not opposed to it, but it doesn't change the behavior right?

src/csharp/Microsoft.Spark/Sql/ArrowArrayHelpers.cs
```diff
@@ -38,10 +39,61 @@ internal static class ArrowArrayHelpers
...
+            }
+            else if (typeof(T) == typeof(PrimitiveDataFrameColumn<sbyte>))
+            {
+                ret = new PrimitiveDataFrameColumn<sbyte>("Empty");
+            }
+            else if (typeof(T) == typeof(PrimitiveDataFrameColumn<byte>))
```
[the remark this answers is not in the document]
pgovind: Cached `typeof(T)`

The only way I've gotten the switch statement to compile is with syntax such 
```case boolColumn when (condition)```. The compiler internally converts them to `if` statements anyway, so I'd rather just use the `if` statements.

File: examples/Microsoft.Spark.CSharp.Examples/Sql/Batch/VectorUdfs.cs [hunk 41]
```diff
@@ -56,39 +58,29 @@ public void Run(string[] args)
             spark.Stop();
         }
 
-        private static RecordBatch CountCharacters(
```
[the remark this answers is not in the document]
pgovind: Yup. Just moved the DataFrame version of this example to its own `VectorDataFrameUdfs.cs` file

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -216,11 +216,51 @@ private sealed class UdfWrapperData
...
             out SerializedMode serializerMode,
             out SerializedMode deserializerMode,
-            out string runMode) where T : Delegate
+            out string runMode)
+        {
+            var udfWrapperData = GetUdfWrapperDataFromStream(stream, out serializerMode, out deserializerMode, out runMode);
+
+            var nodeIndex = 0;
+            var udfIndex = 0;
+            UdfWrapperNode node = udfWrapperData.UdfWrapperNodes[nodeIndex];
```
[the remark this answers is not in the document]
pgovind: I don't quite follow what you mean. Do you mean `udfWrapperData.UdfWrapperNodes` will always be empty?

pgovind:
I haven't addressed all the comments yet. I'll do those tomorrow and this should be pretty much ready to merge at that point.

benchmark/csharp/Tpch/VectorFunctions.cs
```diff
@@ -3,49 +3,30 @@
...
 
-        internal static DoubleArray ComputeDiscountPrice(DoubleArray price, DoubleArray discount)
+        internal static PrimitiveDataFrameColumn<double> ComputeDiscountPrice(PrimitiveDataFrameColumn<double> price, PrimitiveDataFrameColumn<double> discount)
         {
             if (price.Length != discount.Length)
             {
                 throw new ArgumentException("Arrays need to be the same length");
             }
 
-            int length = price.Length;
```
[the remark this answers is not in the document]
pgovind: Yup. Done.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
pgovind: Done. I made a new `ArrowOrDataFrameSqlCommandExecutor` class that internally chooses which SqlCommandExecutor to call based on the type of the `workerFunction`.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -72,6 +107,36 @@ internal abstract class SqlCommandExecutor
             return executor.ExecuteCore(inputStream, outputStream, commands);
         }
 
+        protected IEnumerable<RecordBatch> GetInputIterator(Stream inputStream)
+        {
+            using (var reader = new ArrowStreamReader(inputStream, leaveOpen: true))
```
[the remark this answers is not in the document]
pgovind: Yup, made a new `ArrowBasedCommandExecutor` that implements `GetInputIterator`

benchmark/csharp/Tpch/TpchFunctionalQueries.cs
pgovind: Yup. Introduced `Q1d` and `Q8d` and removed the `#if NETCOREAPP`

pgovind:
Addressed the remaining comments and rebased on current upstream/master

pgovind:
Just a note for myself: I repro'd the CI failures locally. It looks like the type arguments in `CreateDelegate` are not correct. I'll fix this when I'm back on Wednesday.

pgovind:
I believe this is ready for review again now. All the CI errors have been fixed.

benchmark/csharp/Tpch/VectorDataFrameFunctions.cs
```diff
@@ -0,0 +1,32 @@
...
+    internal static class VectorDataFrameFunctions
+    {
+        internal static PrimitiveDataFrameColumn<double> ComputeTotal(PrimitiveDataFrameColumn<double> price, PrimitiveDataFrameColumn<double> discount, PrimitiveDataFrameColumn<double> tax)
+        {
+            if ((price.Length != discount.Length) || (price.Length != tax.Length))
+            {
+                throw new ArgumentException("Arrays need to be the same length");
+            }
+
+            return (PrimitiveDataFrameColumn<double>)(price * (1 - discount) * (1 + tax));
```
eerhardt: This looks so much better than the Arrow implementation.

When we get strongly typed operators, we can cut out the cast and it will look even better.

examples/Microsoft.Spark.CSharp.Examples/Sql/Batch/VectorDataFrameUdfs.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        private static FxDataFrame CountCharacters(
+            FxDataFrame dataFrame,
+            string groupFieldName,
+            string stringFieldName)
+        {
+            int characterCount = 0;
+
+            PrimitiveDataFrameColumn<int> characterCountColumn = new PrimitiveDataFrameColumn<int>(stringFieldName + "CharCount");
+            PrimitiveDataFrameColumn<int> ageColumn = new PrimitiveDataFrameColumn<int>(groupFieldName);
```
eerhardt: If we are going to call this `ageColumn`, we might as well not pass the `groupFieldName` and `stringFieldName` into this function, and just hard-code those names here.

```C#
PrimitiveDataFrameColumn<int> ageColumn = new PrimitiveDataFrameColumn<int>("age");
ArrowStringDataFrameColumn nameColumn = dataFrame["name"] as ArrowStringDataFrameColumn;
```

src/csharp/Microsoft.Spark/Sql/ArrowUdfWrapper.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Spark.Sql
     /// <typeparam name="T">Specifies the type of the first argument to the UDF.</typeparam>
     /// <typeparam name="TResult">Specifies the return type of the UDF.</typeparam>
     [UdfWrapper]
-    internal sealed class ArrowUdfWrapper<T, TResult>
+    internal sealed class ArrowUdfWrapper<T, TResult>
```
eerhardt: It looks like this is all unnecessary white space changes. This whole file can be reverted.

src/csharp/Microsoft.Spark.Experimental/Sql/RelationalGroupedDatasetExtensions.cs
```diff
@@ -13,6 +14,34 @@ namespace Microsoft.Spark.Sql
     /// </summary>
     public static class RelationalGroupedDatasetExtensions
     {
+        /// <summary>
+        /// Maps each group of the current DataFrame using a UDF and
+        /// returns the result as a DataFrame.
+        /// 
+        /// The user-defined function should take an FxDataFrame
```
eerhardt: > should take an FxDataFrame

The readers of this doc don't know what a `FxDataFrame` is. Maybe use a `<see cref` to point to the concrete type.

src/csharp/Microsoft.Spark.Experimental/Sql/UdfRegistrationExtensions.DataFrame.cs
```diff
@@ -0,0 +1,271 @@
...
+using System;
+using Microsoft.Data.Analysis;
+using Microsoft.Spark.Utils;
+
+namespace Microsoft.Spark.Sql
+{
+    /// <summary>
+    /// Extension methods for <see cref="UdfRegistration"/>.
+    /// </summary>
+    public static class DataFrameUdfRegistrationExtensions
```
eerhardt: The type name and the file name should match. The file should be named `DataFrameUdfRegistrationExtensions.cs`.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -294,13 +296,87 @@ public object Run(int splitId, object input)
         }
     }
 
+    internal class ArrowBasedCommandExecutor : SqlCommandExecutor
```
eerhardt: `abstract`?

Then you don't need to override `ExecuteCore`.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -294,13 +296,87 @@ public object Run(int splitId, object input)
...
+            }
+            if (useDataFrameCommandExecutor && useArrowSqlCommandExecutor)
+            {
+                // Mixed mode. Not supported
+                throw new NotSupportedException("Combined Arrow and DataFrame style commands are not supported");
+            }
+            SqlCommandExecutor executor = null;
+            if (useDataFrameCommandExecutor)
+            {
+                executor = new DataFrameSqlCommandExecutor();
```
eerhardt: Do we really need to create a new object here? Can we accomplish the same thing with just calling different methods on the current class?

eerhardt:
This is looking really good. Once the last few comments I have are addressed, I'll be able to sign off.

src/csharp/Microsoft.Spark.Experimental/Sql/RelationalGroupedDatasetExtensions.cs
pgovind: Good idea

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
pgovind: Yup.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
pgovind: Testing this.

pgovind: Yup, good idea. I was going to move all the common methods to the base class, but this is a step further and better.

benchmark/csharp/Tpch/VectorDataFrameFunctions.cs
imback82: Yea this looks really good. Also, it should be as performant right? (We will run a separate benchmark as well)

examples/Microsoft.Spark.CSharp.Examples/README.md
```diff
@@ -29,6 +29,9 @@ an open source and cross-platform machine learning framework.
   <tr>
     <td align="middle"><a href="Sql/Batch/VectorUdfs.cs"><b>VectorUdfs.cs</a></b><br>Example using vectorized UDFs to improve query performance.<br></td>
   </tr>
+  <tr>
+    <td align="middle"><a href="Sql/Batch/VectorDataFrameUdfs.cs"><b>VectorDataFrameUdfs.cs</a></b><br>Example using vectorized UDFs and convenience APIs from Microsft.Data.Analysis to improve query performance.<br></td>
```
imback82: this should go into `<td>` since we have two cols (or use `colspan="2"`)? You can see the output by looking at the rich diff.

examples/Microsoft.Spark.CSharp.Examples/Sql/Batch/VectorDataFrameUdfs.cs
```diff
@@ -0,0 +1,84 @@
...
+                characterCount += nameColumn[i].Length;
+            }
+
+            if (dataFrame.Rows.Count > 0)
+            {
+                characterCountColumn.Append(characterCount);
+                ageColumn.Append((int?)dataFrame["age"][0]);
+            }
+
+            FxDataFrame ret = new FxDataFrame(ageColumn, characterCountColumn);
```
imback82: just `return new FxDataFrame(ageColumn, characterCountColumn);`

examples/Microsoft.Spark.CSharp.Examples/Sql/Batch/VectorDataFrameUdfs.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            spark.Stop();
+        }
+
+        private static FxDataFrame CountCharacters(
+            FxDataFrame dataFrame)
+        {
+            int characterCount = 0;
+
+            PrimitiveDataFrameColumn<int> characterCountColumn = new PrimitiveDataFrameColumn<int>("name" + "CharCount");
```
imback82: nit: Let's use `var characterCountColumn`, `var ageColumn`, and `var nameColumn` per https://github.com/dotnet/spark/blob/master/docs/coding-guidelines/csharp-coding-style.md

benchmark/csharp/Tpch/VectorDataFrameFunctions.cs
eerhardt: The Microsoft.Data.Analysis library doesn't use SIMD operations yet. But eventually the goal is to use them in these math operations.

For now, my assumption would be this would be on the same order as looping over the rows and doing the calculations 1-by-one. Getting a measurement on this would be great.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -219,6 +223,82 @@ public void TestVectorUdf()
             }
         }
 
+        [Fact]
+        public void TestDataFrameVectorUdf()
+        {
+            Func<PrimitiveDataFrameColumn<int>, ArrowStringDataFrameColumn, ArrowStringDataFrameColumn> udf1Func =
+                (ages, names) =>
+                {
+                    StringArray stringArray = (StringArray)ToArrowArray(
```
imback82: nit: var

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -219,6 +223,82 @@ public void TestVectorUdf()
...
         }
 
+        [Fact]
+        public void TestDataFrameVectorUdf()
+        {
+            Func<PrimitiveDataFrameColumn<int>, ArrowStringDataFrameColumn, ArrowStringDataFrameColumn> udf1Func =
+                (ages, names) =>
+                {
+                    StringArray stringArray = (StringArray)ToArrowArray(
+                    Enumerable.Range(0, (int)names.Length)
```
imback82: indentation looks off?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -219,6 +223,82 @@ public void TestVectorUdf()
...
+                Assert.Equal("Michael is 0", rows[0].GetAs<string>(0));
+                Assert.Equal("Andy is 30", rows[1].GetAs<string>(0));
+                Assert.Equal("Justin is 19", rows[2].GetAs<string>(0));
+            }
+
+            // Chained UDFs.
+            Func<Column, Column> udf2 = ExperimentalDataFrameFunctions.VectorUdf<ArrowStringDataFrameColumn, ArrowStringDataFrameColumn>(
+                (strings) =>
+                {
+                    StringArray stringArray = (StringArray)ToArrowArray(
```
imback82: So do we need to first change to arrow array then back to dataframe column? No way to go from `ArrowStringDataFrameColumn` to `ArrowStringDataFrameColumn`? cc: @eerhardt

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -298,6 +378,76 @@ private static RecordBatch CountCharacters(RecordBatch records)
...
+                        throw new Exception($"Unexpected age: {age}.");
+                }
+            }
+        }
+
+        private static FxDataFrame CountCharacters(FxDataFrame dataFrame)
+        {
+            int characterCount = 0;
+
+            PrimitiveDataFrameColumn<int> characterCountColumn = new PrimitiveDataFrameColumn<int>("name" + "CharCount");
```
imback82: ditto on `var` usage.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -298,6 +378,76 @@ private static RecordBatch CountCharacters(RecordBatch records)
...
+                characterCount += fieldColumn[i].Length;
+            }
+
+            if (dataFrame.Rows.Count > 0)
+            {
+                characterCountColumn.Append(characterCount);
+                ageColumn.Append((int?)dataFrame["age"][0]);
+            }
+
+            FxDataFrame ret = new FxDataFrame(ageColumn, characterCountColumn);
```
imback82: ditto

src/csharp/Microsoft.Spark.UnitTest/CommandSerDeTests.cs
```diff
@@ -80,6 +81,48 @@ public void TestCommandSerDeForSqlArrow()
             }
         }
 
+        [Fact]
+        public void TestCommandSerDeForSqlArrowDataFrame()
+        {
+            var udfWrapper = new Sql.DataFrameUdfWrapper<ArrowStringDataFrameColumn, ArrowStringDataFrameColumn>(
+                (strings) =>
+                {
+                    StringArray stringColumn = (StringArray)ToArrowArray(
```
imback82: ditto for `var` and indentation.

src/csharp/Microsoft.Spark.UnitTest/CommandSerDeTests.cs
```diff
@@ -80,6 +81,48 @@ public void TestCommandSerDeForSqlArrow()
...
+                    Enumerable.Range(0, (int)strings.Length)
+                        .Select(i => $"hello {strings[i]}")
+                        .ToArray());
+                    return ToArrowStringDataFrameColumn(stringColumn);
+                });
+
+            var workerFunction = new DataFrameWorkerFunction(udfWrapper.Execute);
+
+            var serializedCommand = Utils.CommandSerDe.Serialize(
```
imback82: nit: `byte[] serializedCommand`

src/csharp/Microsoft.Spark.Experimental/Sql/RelationalGroupedDatasetExtensions.cs
```diff
@@ -13,6 +14,34 @@ namespace Microsoft.Spark.Sql
...
+        /// the returned FxDataFrame are combined as a DataFrame.
+        ///
+        /// The returned <see cref="FxDataFrame"/> can be of arbitrary length and its schema must match
+        /// <paramref name="returnType"/>.
+        /// </summary>
+        /// <param name="dataset">
+        /// The <see cref="RelationalGroupedDataset"/> containing grouped data.
+        /// </param>
+        /// <param name="returnType">
+        /// The <see cref="StructType"/> that represents the shape of the return data set.
```
imback82: that represents the `schema` of

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
eerhardt: I thought we had an `Apply` function that could be used here.
One of the biggest issues is going from the UTF8 arrow string format, to the UTF16 .NET string format, then back to UTF8 arrow string format.

---
In reply to: [387794908](https://github.com/dotnet/spark/pull/277#discussion_r387794908) [](ancestors = 387794908)

src/csharp/Microsoft.Spark.UnitTest/WorkerFunctionTests.cs
```diff
@@ -82,6 +84,21 @@ public void TestArrowWorkerFunction()
...
 
+        [Fact]
+        public void TestDataFrameWorkerFunction()
+        {
+            var func = new DataFrameWorkerFunction(
+                new DataFrameUdfWrapper<ArrowStringDataFrameColumn, ArrowStringDataFrameColumn>(
+                    (str) => str).Execute);
+
+            string[] input = { "arg1" };
+            StringArray column = (StringArray)ToArrowArray(input);
```
imback82: nit: `var column`

src/csharp/Microsoft.Spark.UnitTest/WorkerFunctionTests.cs
```diff
@@ -109,6 +126,41 @@ public void TestArrowWorkerFunctionForBool()
...
+        /// for both input and output.
+        /// </summary>
+        [Fact]
+        public void TestDataFrameWorkerFunctionForBool()
+        {
+            var func = new DataFrameWorkerFunction(
+                new DataFrameUdfWrapper<ArrowStringDataFrameColumn, PrimitiveDataFrameColumn<bool>, PrimitiveDataFrameColumn<bool>>(
+                    (strings, flags) =>
+                    {
+                        for (long i = 0; i < strings.Length; i++)
```
imback82: nit: `++i`

src/csharp/Microsoft.Spark.UnitTest/WorkerFunctionTests.cs
```diff
@@ -109,6 +126,41 @@ public void TestArrowWorkerFunctionForBool()
...
+                        {
+                            flags[i] = flags[i].Value || strings[i].Contains("true");
+                        }
+                        return flags;
+                    }).Execute);
+
+            StringArray stringColumn = (StringArray)ToArrowArray(new[] { "arg1_true", "arg1_true", "arg1_false", "arg1_false" });
```
imback82: nit: `var stringColumn`, `var boolColumn`, `var input`

src/csharp/Microsoft.Spark.UnitTest/WorkerFunctionTests.cs
```diff
@@ -109,6 +126,41 @@ public void TestArrowWorkerFunctionForBool()
...
+
+            ArrowStringDataFrameColumn ArrowStringDataFrameColumn = ToArrowStringDataFrameColumn(stringColumn);
+            PrimitiveDataFrameColumn<bool> boolColumn = new PrimitiveDataFrameColumn<bool>("Bool", Enumerable.Range(0, 4).Select(x => x % 2 == 0 ? true : false));
```
imback82: do you need `? true : false`?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
imback82: Yea, it will be really nice if the user doesn't have to go thru Arrow APIs but lives in FxDataFrame APIs.

pgovind: We don't have it in 0.2.0 yet. However, we should release a 0.3.0 soon which will have this API. How about we accept this change for now and I put up a new PR once 0.3.0 is out?

eerhardt: That sounds like a good plan to me.

pgovind: Filed https://github.com/dotnet/corefxlab/issues/2860

imback82: Sounds good to me as well.

src/csharp/Microsoft.Spark.UnitTest/TestUtils/ArrowTestUtils.cs
```diff
@@ -100,6 +108,16 @@ public static IArrowType GetArrowType<T>()
             throw new NotSupportedException($"Unknown type: {typeof(T)}");
         }
 
+        public static ArrowStringDataFrameColumn ToArrowStringDataFrameColumn(StringArray array)
+        {
+            return new ArrowStringDataFrameColumn("String",
```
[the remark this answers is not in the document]
pgovind: Actually we can't. Arrow doesn't allow empty or null values for the column name. We encounter this when we go back from DataFrame->Arrow Record Batches

pgovind:
Addressed comments and did a search/replace for styling changes(I think I got all of them).

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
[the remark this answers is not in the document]
imback82: n/m I read the code wrong. :)

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -217,11 +217,60 @@ private sealed class UdfWrapperData
...
+            var nodeType = Type.GetType(node.TypeName);
```
imback82: no `var` here.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -217,11 +217,60 @@ private sealed class UdfWrapperData
...
+            Delegate udf = null;
+            if (nodeType == typeof(DataFrameGroupedMapUdfWrapper))
+            {
+                udf = (DataFrameGroupedMapWorkerFunction.ExecuteDelegate)DeserializeUdfs<DataFrameGroupedMapWorkerFunction.ExecuteDelegate>(
+                        udfWrapperData,
+                        ref nodeIndex,
+                        ref udfIndex);
```
imback82: indentation looks off. I will also do
```
                udf = (DataFrameGroupedMapWorkerFunction.ExecuteDelegate)
                    DeserializeUdfs<DataFrameGroupedMapWorkerFunction.ExecuteDelegate>(
                        udfWrapperData,
                        ref nodeIndex,
                        ref udfIndex);
```

same for below else ifs.

imback82:
> Addressed comments and did a search/replace for styling changes(I think I got all of them).

I was probably reviewing while you were making changes. If you already fixed them, just resolve my comments. Thanks!

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -567,33 +750,49 @@ internal class ArrowGroupedMapCommandExecutor : SqlCommandExecutor
...
+                FxDataFrame resultDataFrame = worker.Func(dataFrame);
+                IEnumerable<RecordBatch> recordBatches = resultDataFrame.ToArrowRecordBatches();
 
-                if (!returnedResult)
+                foreach (RecordBatch result in recordBatches)
                 {
-                    // When no input batches were received, return an empty RecordBatch
-                    // in order to create and write back the result schema.
+                    int numEntries = result.Length;
+                    stat.NumEntriesProcessed += numEntries;
```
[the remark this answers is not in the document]
pgovind: Yup. Remnants of debug code

pgovind:
> I see. So when we get ArrowStringDataFrameColumn in the UDF, is it already named? If so, the name is from the Arrow sent from JVM, right?

Yup, it is already named with the Arrow data from the JVM. The Arrow schema defines a `Field` for a `StringColumn` that provides a `Name`.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -217,11 +217,65 @@ private sealed class UdfWrapperData
...
 
-        internal static T Deserialize<T>(
+        internal static object DeserializeArrowOrDataFrameUdf(
             Stream stream,
             out SerializedMode serializerMode,
             out SerializedMode deserializerMode,
-            out string runMode) where T : Delegate
+            out string runMode)
+        {
+            var udfWrapperData = GetUdfWrapperDataFromStream(
```
imback82: Looks like I missed `var` usage in this file. :) Can you fix `udfWrapperData`, `nodeIndex`, `udfIndex` in this file? Thanks!

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -240,10 +294,18 @@ private sealed class UdfWrapperData
...
 
-            int nodeIndex = 0;
-            int udfIndex = 0;
+        internal static T Deserialize<T>(
+            Stream stream,
+            out SerializedMode serializerMode,
+            out SerializedMode deserializerMode,
+            out string runMode) where T : Delegate
+        {
+            var udfWrapperData = GetUdfWrapperDataFromStream(stream, out serializerMode, out deserializerMode, out runMode);
```
imback82: sorry, I meant this file.

pgovind:
Thanks for the review @imback82. I know it was a time consuming PR

PR #291, opened 2019-10-12
Support .NET Core 3.1

imback82:
Now that .NET Core 3.0 is officially released, we are moving worker binaries from `netcoreapp2.1` to `netcoreapp3.1`.


imback82:
I will also update docs that are related to building from source.

imback82:
cc: @eerhardt, can you check if I am missing anything migrating to netcore 3.0? Thanks!

docs/building/windows-instructions.md
```diff
@@ -20,7 +20,7 @@ Building Spark .NET on Windows
 
 If you already have all the pre-requisites, skip to the [build](windows-instructions.md#building) steps below.
 
-  1. Download and install the **[.NET Core SDK](https://dotnet.microsoft.com/download/dotnet-core/2.1)** - installing the SDK will add the `dotnet` toolchain to your path. .NET Core 2.1, 2.2 and 3.0 preview are supported.
+  1. Download and install the **[.NET Core 3.0 SDK](https://dotnet.microsoft.com/download/dotnet-core/3.0)** - installing the SDK will add the `dotnet` toolchain to your path.
   2. Install any edition of **[Visual Studio 2019](https://www.visualstudio.com/downloads/)** or [Visual Studio 2017](https://www.visualstudio.com/downloads/). The Community version is completely free. When configuring your installation, include these components at minimum:
```
eerhardt: Since we are now building for `netcoreapp3.0`, we now require VS 2019 Update 3. So this should be updated.

src/csharp/Extensions/Microsoft.Spark.Extensions.Delta/Microsoft.Spark.Extensions.Delta.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
```
eerhardt: You don't need to make this change. You can just target `netstandard2.0` and everything will work find on .NET Core 3.0.

The only reason you would multi-target for standard 2.0 and 2.1 is if you wanted to use a new API in 2.1 that wasn't available in 2.0, which it doesn't appear that we are doing.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
```
eerhardt: You don't need to make this change. You can just target netstandard2.0 and everything will work find on .NET Core 3.0.

The only reason you would multi-target for standard 2.0 and 2.1 is if you wanted to use a new API in 2.1 that wasn't available in 2.0, which it doesn't appear that we are doing.

src/csharp/Microsoft.Spark.UnitTest/Microsoft.Spark.UnitTest.csproj
```diff
@@ -10,7 +10,7 @@
     <PackageReference Include="System.Memory" Version="4.5.2" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' != 'netcoreapp2.1' ">
+  <ItemGroup Condition=" '$(TargetFramework)' != 'netcoreapp3.0' ">
```
eerhardt: This can just get removed. There is never a case where TargetFramework != netcoreapp3.0.

imback82:
> `3.0` is the `Current` release of .NET Core. While `2.1` is still the LTS (long-term support) version. `3.1` will be the next LTS version shipping next month. Which means `3.0` will go out of support in about 4 months. So we will need to upgrade to `3.1` when it comes out. This shouldn't block us from upgrading to `3.0` today, but I just wanted to ensure you were aware of the policy.

Thanks @eerhardt. In this case, I will leave this PR open and update it to 3.1 when it's released next month.

src/csharp/Extensions/Microsoft.Spark.Extensions.Delta/Microsoft.Spark.Extensions.Delta.csproj
AeroXuk: For consideration with targeting `netstandard2.1`:
> .NET Framework won't support .NET Standard 2.1 or later versions. For more details, see the [announcement of .NET Standard 2.1](https://devblogs.microsoft.com/dotnet/announcing-net-standard-2-1/).

Source: https://docs.microsoft.com/en-us/dotnet/standard/net-standard

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
AeroXuk: For consideration with targeting `netstandard2.1`:
> .NET Framework won't support .NET Standard 2.1 or later versions. For more details, see the [announcement of .NET Standard 2.1](https://devblogs.microsoft.com/dotnet/announcing-net-standard-2-1/).

Source: https://docs.microsoft.com/en-us/dotnet/standard/net-standard

src/csharp/Microsoft.Spark.Experimental/Microsoft.Spark.Experimental.csproj
```diff
@@ -1,7 +1,7 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
```
AeroXuk: See other `netstandard2.1` comments on this PR.

src/csharp/Extensions/Microsoft.Spark.Extensions.Delta/Microsoft.Spark.Extensions.Delta.csproj
eerhardt: That's why it is targeting both `netstandard2.0` and `netstandard2.1`. .NET Framework will use the `netstandard2.0` compatible version.

imback82:
Now that .NET Core 3.1 is released, I will resume this work.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
imback82: reverted

docs/building/windows-instructions.md
imback82: updated to install 16.4 or higher. thanks!

src/csharp/Extensions/Microsoft.Spark.Extensions.Delta/Microsoft.Spark.Extensions.Delta.csproj
imback82: reverted.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netcoreapp2.1;netstandard2.1</TargetFrameworks>
```
imback82: @eerhardt Should I add `netcoreapp3.1` instead of replacing `netcoreapp2.1`?

eerhardt: For libraries (like `Microsoft.Spark.csproj`), you don't need to target all the frameworks you support running on. You just need to target the lowest/earliest frameworks that you support.

For example, if you just target `netstandard2.0`, your library will run on any desktop .NET Framework 4.6.1 or higher. And it will run on .NET Core 2.0 or higher.

The reason you would multi-target a library is so you can have different code that will execute when running on different frameworks. For example, .NET Core 3.1 adds a new API you want to use, so you multi-target to `netstandard2.0;netcoreapp3.1`. And now you can use `#if` conditions in your code to compile different code when building for `netstandard2.0`, and use the new API when building for `netcoreapp3.1`.

Since you don't have `netcoreapp3.1` specific code in this library, it isn't necessary to put `netcoreapp3.1` in this list at all.

The only framework-specific code I see in this project is:

https://github.com/dotnet/spark/blob/8dc789fa6cc5d60eecb5d787cc04ea3a366e2035/src/csharp/Microsoft.Spark/Sql/Types/DataType.cs#L208-L219

So in all actuality, if you want to drop 1st class support for .NET Core 2.1, you would really only need:

`<TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>` on this line. (Because .NET Core 3.1 is compatible with `netstandard2.1`, so the `netstandard2.1` compiled assemblies will be used when running on .NET Core 3.1.)

imback82: Thanks @eerhardt. `netcoreapp2.1` was added from this discussion: https://github.com/dotnet/spark/pull/364#discussion_r354627266. I will remove `netcoreapp2.1` for now and will add it if it is absolutely needed.

eerhardt: The only affect this will cause is if someone is on .NET Core 2.1 and uses the `Microsoft.Spark` library, the above code will use the slower Substring/ToLower code since the `netstandard2.0` built version will be used. That doesn't seem too bad, especially if you aren't supporting .NET Core 2.1 on the worker anymore.

imback82: Yea, and this code is driver side and not used by worker anyway. Thanks @eerhardt!

PR #296, opened 2019-10-16
DataStreamReader.Load(path) and E2E test for DeltaTable streaming

AFFogarty:
This PR does two things:

1. Adds an overload to `DataStreamReader.Load()` that accepts a `path` parameter.  This new overload matches the Spark API documented [here](https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/streaming/DataStreamReader.html#load-java.lang.String-).  Without this overload, DeltaTables can only be read as a stream source by calling `.Option("path", path).Load()` which is verbose and does not match the Scala syntax found in the Delta documentation found [here](https://docs.delta.io/latest/delta-streaming.html#delta-table-as-a-stream-source).
2. Adds an E2E test for basic `DeltaTable` streaming functionality.

### DIFF
--- a/src/csharp/Extensions/Microsoft.Spark.Extensions.Delta.E2ETest/DeltaTableTests.cs
+++ b/src/csharp/Extensions/Microsoft.Spark.Extensions.Delta.E2ETest/DeltaTableTests.cs
@@ -116,11 +116,12 @@ namespace Microsoft.Spark.Extensions.Delta.E2ETest
                 string sourcePath = Path.Combine(tempDirectory.Path, "source-delta-table");
                 _spark.Range(0, 5).Write().Format("delta").Save(sourcePath);
 
-                DataFrame source = _spark.ReadStream().Format("delta").Load(sourcePath);
-
                 // Create a stream from the source DeltaTable to the sink DeltaTable.
                 string sinkPath = Path.Combine(tempDirectory.Path, "sink-delta-table");
-                StreamingQuery stream = source
+                StreamingQuery stream = _spark
+                    .ReadStream()
+                    .Format("delta")
+                    .Load(sourcePath)
                     .WriteStream()
                     .Format("delta")
                     .OutputMode("append")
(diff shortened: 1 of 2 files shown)


src/csharp/Extensions/Microsoft.Spark.Extensions.Delta.E2ETest/DeltaTableTests.cs
```diff
@@ -102,6 +104,51 @@ public void TestTutorialScenario()
...
+                string sinkPath = Path.Combine(tempDirectory.Path, "sink-delta-table");
+                StreamingQuery stream = source
+                    .WriteStream()
+                    .Format("delta")
+                    .OutputMode("append")
+                    .Option("checkpointLocation", Path.Combine(tempDirectory.Path, "checkpoints"))
+                    .Start(sinkPath);
+
+                // Take a short pause so that the stream has time to write to the sink.
+                Thread.Sleep(TimeSpan.FromSeconds(5));
```
imback82: This is a reason the test was not added initially. Also, this could cause random failure if the sink doesn't happen in 5 seconds. Is there another approach?

AFFogarty: Ok, I will change the test to use a "One-time micro-batch" trigger so that everything runs synchronously.

AFFogarty: Made the change.  Everything should now be synchronous and deterministic.

src/csharp/Extensions/Microsoft.Spark.Extensions.Delta.E2ETest/DeltaTableTests.cs
```diff
@@ -170,6 +213,10 @@ public void TestSignatures()
                 table.Delete("id > 10");
                 table.Delete(Functions.Expr("id > 5"));
                 table.Delete();
+
+                // Load the table as a streaming source.
+                Assert.IsType<DataFrame>(_spark.ReadStream().Format("delta").Option("path", path).Load());
```
suhsteve: over 100 chars.

AFFogarty: Fixed.

AFFogarty:
Fixed the line length issues.

There are a couple lines left that look long but are actually 100 chars long.

PR #301, opened 2019-10-22
Release 0.6.0

imback82:
Release prep for 0.6.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@
     <tbody align="center">
         <tr>
             <td >2.3.*</td>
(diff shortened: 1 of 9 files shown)


suhsteve:
I'm wondering when the Microsoft.Spark.Extensions.Delta `0.5.1` nuget is built, will it automatically have Microsoft.Spark `0.5.1` as part if its dependencies?

imback82:
Thanks @suhsteve. We will prob. do 0.6.0 release to avoid confusion.

PR #320, opened 2019-10-29
Samples Revamp: ML - Batch Sentiment Analysis

bamurtaugh:
-Add new batch sentiment analysis sample app & readme
-Add Machine Learning folder, and readme folder within it
-Add readme for Machine Learning folder
-Add reference to ML.NET v1.3.1 in C# csproj file


examples/Microsoft.Spark.CSharp.Examples/Sql/MachineLearning/SentimentAnalysis.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// </summary>
+    internal sealed class SentimentAnalysis : IExample
+    {
+        public void Run(string[] args)
+        {
+            // Change this flag once you've set up
+            // the ML.NET dependencies described in the README:
+            // Update url to location of MLModel.zip (line 90)
+            // Copy ML.NET dlls to your project's folder
+            int dependenciesDone = 0;
```
suhsteve: Is this really necessary ?

examples/Microsoft.Spark.CSharp.Examples/Sql/MachineLearning/SentimentAnalysis.cs
```diff
@@ -0,0 +1,122 @@
...
+                "MLudf",
+                (text) => Sentiment(text));
+
+            df.CreateOrReplaceTempView("Reviews");
+            DataFrame sqlDf = spark.Sql("SELECT Column1, MLudf(Column1) FROM Reviews");
+            sqlDf.Show();
+
+            // Print out first 20 rows of data
+            // Prevents data getting cut off (as it is when we print a DF)
+            IEnumerable<Row> rows = sqlDf.Collect();
```
suhsteve: You can use the [`Limit`](https://github.com/dotnet/spark/blob/67451fa1179e9d9667d96c40fc9ae5b222dcea8d/src/csharp/Microsoft.Spark/Sql/DataFrame.cs#L483) method

examples/Microsoft.Spark.CSharp.Examples/Sql/MachineLearning/SentimentAnalysis.cs
```diff
@@ -0,0 +1,122 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.ML;
+using Microsoft.ML.Data;
+using Microsoft.Spark.Sql;
+
+namespace Microsoft.Spark.Examples.Sql.MachineLearning
```
suhsteve: Should these be in `Examples.Sql.MachineLearning` or just `Examples.MachineLearning` ?  I think the latter, mostly because we are trying to showcase MachineLearning with these examples.

examples/Microsoft.Spark.CSharp.Examples/Sql/MachineLearning/SentimentAnalysis.cs
```diff
@@ -0,0 +1,122 @@
...
+{
+    /// <summary>
+    /// Example of using ML.NET + .NET for Apache Spark
+    /// for sentiment analysis.
+    /// </summary>
+    internal sealed class SentimentAnalysis : IExample
```
suhsteve: If we want to do away with `IExample` interface everywhere, we can have separate folders and `.csproj` for every example (creating separate executables for every example).  This way you can make your ML examples and README a bit simpler I think  You can include some folder  `Examples\MachineLearning\Sentiment` and in this folder you can have a subfolder `SentimentML.Model` with a `SentimentML.Model.csproj` along with its source code.  In the root folder you can have Sentiment.csproj and add a `ProjectReference` to the `SentimentML.Model.csproj`.   This same comment would also be relevant to the other ML and original SQL/Streaming examples.

examples/Microsoft.Spark.CSharp.Examples/Sql/MachineLearning/SentimentAnalysis.cs
bamurtaugh: Unnecessary now.

examples/Microsoft.Spark.CSharp.Examples/Sql/MachineLearning/SentimentAnalysis.cs
bamurtaugh: Used `Take(20)` instead (so that it would return an `IEnumerable<Row>`)

examples/Microsoft.Spark.CSharp.Examples/Sql/MachineLearning/SentimentAnalysis.cs
bamurtaugh: I believe we addressed this feedback through our collaboration yesterday. Thank you again for all your help!

examples/Microsoft.Spark.CSharp.Examples/Sql/MachineLearning/SentimentAnalysis.cs
bamurtaugh: Adjusted to `Examples.MachineLearning.Sentiment` to reflect newest folder structure.

examples/Microsoft.Spark.CSharp.Examples/Microsoft.Spark.CSharp.Examples.csproj
```diff
@@ -13,4 +13,9 @@
     <ProjectReference Include="..\..\src\csharp\Microsoft.Spark\Microsoft.Spark.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ML" Version="1.3.1" />
```
suhsteve: this shouldn't be required.   `Microsoft.ML.FastTree` brings in this library as a dependency.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+namespace Microsoft.Spark.Examples.MachineLearning.Sentiment
+{
+    /// <summary>
+    /// Example of using ML.NET + .NET for Apache Spark
+    /// for sentiment analysis.
+    /// </summary>
+    //public class Program
+    internal sealed class Program : IExample
+    {
+        //public static void Main(string[] args)
```
suhsteve: remove comment

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+                .AppName(".NET for Apache Spark Sentiment Analysis")
+                .GetOrCreate();
+
+            // Read in and display Yelp reviews
+            DataFrame df = spark
+                .Read()
+                .Option("header", true)
+                .Option("inferSchema", true)
+                .Csv(args[0] + "yelp.csv");
+                //.Csv(@"Resources\yelp.csv");
```
suhsteve: remove comment

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+            DataFrame sqlDf = spark.Sql("SELECT Column1, MLudf(Column1) FROM Reviews");
+            sqlDf.Show();
+
+            // Print out first 20 rows of data
+            // Prevents data getting cut off (as it is when we print a DF)
+            IEnumerable<Row> rows = sqlDf.Take(20);
+            foreach (Row row in rows)
+            {
+                Console.WriteLine(row);
+            }
```
suhsteve: `DataFrame.Show()` takes in parameters to adjust how many rows to show.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+        // Method to call ML.NET code for sentiment analysis
+        // Code primarily comes from ML.NET Model Builder
+        public static bool Sentiment(string text, string modelPath)
+        {
+            MLContext mlContext = new MLContext();
+
+            ITransformer mlModel = mlContext
+                .Model
+                .Load(modelPath, out var modelInputSchema);
+                //.Load(@"Resources\MLModel.zip", out var modelInputSchema);
```
suhsteve: remove comment

examples/Microsoft.Spark.CSharp.Examples/Microsoft.Spark.CSharp.Examples.csproj
bamurtaugh: Oh that's right! Just changed.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
bamurtaugh: I wanted to display each row completely (rather than getting cutoff as it does with a DataFrame), thus I chose to convert to an IEnumerable and then display each row object. Unless there's a way to `DataFrame.Show()` without cutting off any of the output?

suhsteve: I believe you can set the `truncate` parameter to 0.

bamurtaugh: Did not know about this parameter, changed now. Thanks!

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Problem
+
+Our goal here is to determine if online reviews are positive or negative. We'll be using .NET for Apache Spark to read in a dataset of reviews and ML.NET to perform **binary classification** since categorizing reviews involves choosing one of two groups: positive or negative. You can read more about the problem through the [ML.NET documentation](https://docs.microsoft.com/en-us/dotnet/machine-learning/tutorials/sentiment-analysis).
+
+## Dataset
+
+We'll be using a set of Amazon reviews to train our model and a set of Yelp reviews for testing in our Spark + ML app. You can [download the original data](https://archive.ics.uci.edu/ml/machine-learning-databases/00331/sentiment%20labelled%20sentences.zip) from the [UCI Sentiment Labeled Sentences Dataset]( https://archive.ics.uci.edu/ml/datasets/Sentiment+Labelled+Sentences).
+
+For the specific ML training/predictions in this app (i.e. when using Model Builder), it helps to have a header for the data. Versions of the Amazon and Yelp datasets with headers can be found in the [Resources](./Resources) folder.
```
suhsteve: I looked at the `.csv` files with the headers.   Would it be better to have more meaningful names for the columns besides `Column1` `Column2` ?

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+![ML.NET Zip and Files](https://github.com/bamurtaugh/spark/blob/SparkMLNet/examples/Microsoft.Spark.CSharp.Examples/MachineLearning/images/modelbuilder5proj.PNG)
+
+You'll also generate C# code you can use to consume your model in other .NET apps, like the Spark app we'll be creating.
+
+![Generated Code](https://github.com/bamurtaugh/spark/blob/SparkMLNet/examples/Microsoft.Spark.CSharp.Examples/MachineLearning/images/modelbuilder5code.PNG)
+
+### 4. Add ML.NET to .NET for Apache Spark App
+
+You have a few options to start creating a .NET for Apache Spark app that uses this ML.NET code and trained model. Make sure that in any app you develop, you've downloaded the [Microsoft.ML NuGet Package](https://www.nuget.org/packages/Microsoft.ML).
```
suhsteve:
```suggestion
You have a few options to start creating a .NET for Apache Spark app that uses this ML.NET code and trained model. Make sure that in any app you develop, you've added the [Microsoft.ML.FastTree NuGet Package](https://www.nuget.org/packages/Microsoft.ML.FastTree).
```

suhsteve: Do you know why we need FastTree?   Does that mean the `Next Steps` in the screenshot in section 3 is incorrect ?

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
```diff
@@ -0,0 +1,188 @@
...
+* A console app that allows you to build/train/test the model (**myMLAppML.ConsoleApp**)
+* A .NET Standard class library that contains model input/output and your trained model in a zip file (**myMLAppML.Model**)
+
+You would begin writing your .NET for Apache Spark code (and paste in the code generated from Model Builder, shown in step 3 above) in the original app **myMLApp.**
+
+![Model Builder Result](https://dotnet.microsoft.com/static/images/model-builder-generated-code.png?v=iC-r8k3zpKUwQVoNOH34D903IhXhIb4CsX003484s7c)
+
+#### Option 2: Create a new console app (shown in this repo)
+
+Rather than working with the projects/files produced by Model Builder's Add Projects, you can create a new, separate C# console app. You just need to copy over your model's zip file to a directory your new console app can access. In this repo, a trained model **MLModel.zip** has already been included for you in the *Resources* folder.
```
suhsteve:
```suggestion
Rather than working with the projects/files produced by Model Builder's Add Projects, you can create a new, separate C# console app. You just need to copy over your model's zip file to a directory your new console app can access. In this repo, a trained model **MLModel.zip** has already been included for you in the [Resources](./Resources) folder.
```

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+As we create the logic for our Spark app, we'll paste in the code generated from Model Builder and include some other class definitions.
+
+## .NET for Spark
+
+Now that we've trained an ML.NET model for sentiment analysis, we can begin writing the .NET for Spark code that will read in our Yelp data, pass each review to our ML.NET model, and predict whether reviews are positive or negative.
```
suhsteve: Just a general ML question.  Is it okay to mix different types of reviews like this ?  The MLModel trained using the Amazon dataset, which are reviews on the quality of a product.   Yelp reviews would be reviews on food/decor/service.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+### 1. Create a Spark Session
+
+In any Spark application, we need to establish a new SparkSession, which is the entry point to programming Spark with the Dataset and 
+DataFrame API.
+
+```CSharp
+SparkSession spark = SparkSession
+       .Builder()
```
suhsteve:
```suggestion
    .Builder()
```
fix spacing with below code as well.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
```diff
@@ -0,0 +1,188 @@
...
+spark.Udf().Register<string, bool>("MLudf", (text) => Sentiment(text));
+```
+
+*Sentiment* is where we'll call our ML.NET code that was generated from the final step of Model Builder.
+
+```CSharp
+MLContext mlContext = new MLContext();
+ITransformer mlModel = mlContext.Model.Load(@"Resources\MLModel.zip", out var modelInputSchema);
+var predEngine = mlContext.Model.CreatePredictionEngine<Review, ReviewPrediction>(mlModel);
+```
```
suhsteve: Slightly different than the code, but I assume it's to make it easier to understand ?

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
```diff
@@ -0,0 +1,188 @@
...
+      // Represents the input review text
+      [LoadColumn(0)]
+      public string Column1;
+}
+```
+
+```CSharp
+public class ReviewPrediction : Review
+{
+
```
suhsteve:
```suggestion
```

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
```diff
@@ -0,0 +1,188 @@
...
+       public float Score { get; set; }
+} 
+```
+
+### 4. Spark SQL and Running Your Code
+
+Now that you've read in your data and incorporated ML, use Spark SQL to call the UDF that will run sentiment analysis on each row of your DataFrame:
+
+```CSharp
+DataFrame sqlDf = spark.Sql("SELECT _c0, MLudf(_c0) FROM Reviews");
```
suhsteve: This is different than what's in the code.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
```diff
@@ -0,0 +1,101 @@
...
+        // Code primarily comes from ML.NET Model Builder
+        public static bool Sentiment(string text, string modelPath)
+        {
+            MLContext mlContext = new MLContext();
+
+            ITransformer mlModel = mlContext
+                .Model
+                .Load(modelPath, out var modelInputSchema);
+
+            var predEngine = mlContext
```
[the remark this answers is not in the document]
bamurtaugh: Replaced with `PredictionEngine<Review, ReviewPrediction>`

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
```diff
@@ -0,0 +1,101 @@
...
+               .Model
+               .CreatePredictionEngine<Review, ReviewPrediction>(mlModel);
+
+            var result = predEngine.Predict(
```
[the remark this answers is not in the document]
bamurtaugh: Replaced with `ReviewPrediction`

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
bamurtaugh: I agree, I had used those column names while originally writing/testing the app, and now I'll adapt to have better names. Switching column names to `ReviewText` and `Sentiment`.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
bamurtaugh: So I just retrained the ML model with Model Builder (since I needed to train it with the new column names), and the recommended algorithm changed. I'm thinking that depending upon which algorithm is chosen and thus used within the trained model, that affects which nugets the user needs to download.

I referenced the new MLModel.zip I generated, and tried removing the reference to `Microsoft.ML.FastTree` and just including `Microsoft.ML` instead, and it worked. So I think it depends on the model.

I'll add an explanation about this to the readme.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
bamurtaugh: I believe it should be acceptable in this use case. Both datasets seem general enough that the model can learn what a user says when they're positive vs when they're negative about something (whether it's product quality or food/service). 

i.e. it'll catch that terms like "great" "very good" and "ideal" are positive, while "unacceptable" "waste" and "unacceptable" are negative, and both datasets contain terms like these.

bamurtaugh: Also, I found these datasets through the [ML.NET sentiment analysis tutorial](https://docs.microsoft.com/en-us/dotnet/machine-learning/tutorials/sentiment-analysis), so I also took that as a sign that it should be fine.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
bamurtaugh: Updated it now.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
bamurtaugh: I think it was initially to make it easier, but it just seems like now it'd be easier to include and explain the actual code. I've updated it to reflect the current code in `Program.cs`

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
bamurtaugh: Added an explanation, let me know what you think.

suhsteve: Sounds good!

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
suhsteve: Generally in the ML world if there is a small dataset, then they split it into a training set and a test set.   From the link you gave, they only use the yelp dataset.  The ML library looks it reads the yelp dataset and will createa two subsets, a training set (to create the model) and a test set (to do testing and validations).

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
bamurtaugh: Great, thanks! Marking this as resolved then (but let me know if you have any other feedback about it).

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
bamurtaugh: Yep I see where they have divided Yelp dataset into training and testing. Do you think trying to integrate something like that would be preferable/not too confusing rather than using the 2 separate datasets as the sample stands now?

suhsteve: I've done a little dabbling with ML, but my only concern is that I'm just unsure how legitimate this example would seem to an ML expert.

bamurtaugh: Yeah I understand what you're saying. I feel like with sentiment analysis it could be okay since we're just focusing on simple text, but to improve accuracy/legitimacy, I'm going to divide the Yelp dataset in half and train with one half of it and test with the other half.

bamurtaugh: Just updated, thanks for the feedback

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// <summary>
+    /// Example of using ML.NET + .NET for Apache Spark
+    /// for sentiment analysis.
+    /// </summary>
+    //public class Program
+    internal sealed class Program : IExample
+    {
+        public void Run(string[] args)
+        {
+            if (args.Length != 1)
```
suhsteve: Instead of a path to the Resources folder, can we instead have the user enter in the full path to the ML Model zip and also the full path to the dataset to run the ML model against ?

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+You may notice the use of *Review* and *ReviewPrediction.* These are classes we define in our project to represent the review data we're evaluating:
+
+```CSharp
+public class Review
+{
+      // Represents the input review text
+      [LoadColumn(0)]
+      public string Column1;
```
suhsteve:
```suggestion
      public string ReviewText;
```

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
bamurtaugh: Sure, I can make that change. I had written it this way because I thought it might make it easier to have fewer command line arguments. Is your thought here that it'd be better to have the user enter the two paths so that they aren't restricted to the Resources folder/requiring both resources to be in the same folder?

suhsteve: If you do it how it is, then you'll have to use `Path.Combine()` for yelptest.csv and MLModel.zip

bamurtaugh: Makes sense. Changed, thanks!

suhsteve: Can you update the readme too

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.ML.Data;
+using Microsoft.Spark.Sql;
+
+namespace Microsoft.Spark.Examples.MachineLearning.Sentiment
+{
+    /// <summary>
+    /// Example of using ML.NET + .NET for Apache Spark
+    /// for sentiment analysis.
+    /// </summary>
+    //public class Program
```
imback82: can you remove this?

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
```diff
@@ -0,0 +1,100 @@
...
+        public static bool Sentiment(string text, string modelPath)
+        {
+            MLContext mlContext = new MLContext();
+
+            ITransformer mlModel = mlContext
+                .Model
+                .Load(modelPath, out var modelInputSchema);
+
+            PredictionEngine<Review, ReviewPrediction> predEngine = mlContext
+               .Model
```
imback82: indentation seems off?

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+![Generated Code](https://github.com/bamurtaugh/spark/blob/SparkMLNet/examples/Microsoft.Spark.CSharp.Examples/MachineLearning/images/modelbuilder5code.PNG)
+
+### 4. Add ML.NET to .NET for Apache Spark App
+
+You have a few options to start creating a .NET for Apache Spark app that uses this ML.NET code and trained model. Make sure that in any app you develop, you've added the [Microsoft.ML NuGet Package](https://www.nuget.org/packages/Microsoft.ML).
+
+Depending upon the algorithm Model Builder chooses for your ML model, you may need to add an additional nuget reference in [Microsoft.Spark.CSharp.Examples.csproj](../../Microsoft.Spark.CSharp.Examples.csproj). For instance, if you get an error message that Microsoft.ML.FastTree cannot be found when running your Spark app, you need to add that nuget to your csproj file:
+
+`<PackageReference Include="Microsoft.ML.FastTree" Version="1.3.1" />`
```
imback82: this seems different from what we are referencing in our csproj?

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
```diff
@@ -0,0 +1,100 @@
...
+            sqlDf.Show(20, 0, false);
+
+            spark.Stop();
+        }
+
+        // Method to call ML.NET code for sentiment analysis
+        // Code primarily comes from ML.NET Model Builder
+        public static bool Sentiment(string text, string modelPath)
+        {
+            MLContext mlContext = new MLContext();
```
imback82: nit: we prefer`var` in this case.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
bamurtaugh: Missed this, removed.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
bamurtaugh: It was somehow a few spaces off, fixed now.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/README.md
bamurtaugh: Steve and I learned that depending on what algorithm is used in the model, you may have to add a different reference in your .csproj. Originally we needed `FastTree`, but once I changed to a new model that used a different algorithm, it no longer used `FastTree` and thus we could just reference `Microsoft.ML`. In case users train and use their own model and then run into issues, we wanted to add this explanation.

examples/Microsoft.Spark.CSharp.Examples/MachineLearning/Sentiment/Program.cs
bamurtaugh: Done!

PR #331, opened 2019-11-09
[Scala] Use Spark 3.0.0-preview instead of 3.0.0-SNAPSHOT

imback82:
Consume 3.0.0-preview jars from the Maven repository instead of the SNAPSHOT repository.


imback82:
merging to master

PR #332, opened 2019-11-10
Remove SparkSession.*ActiveSession APIs

### LINKED ISSUE #278: [BUG]: Fix flaky tests in SparkSessionTests.cs
Tests in `SparkSessionTests.cs` have been flaky (mostly due to clearing/setting the active session. This needs an investigation.

imback82:
`SparkSession.*ActiveSession` APIs uses thread-local variable to store the session. Since the netty server on the JVM side is multi-threaded, the behavior is unexpected.

This also fixes #278.

### DIFF
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
@@ -32,10 +32,6 @@ namespace Microsoft.Spark.E2ETest.IpcTests
 
             Assert.IsType<Builder>(SparkSession.Builder());
 
-            SparkSession.ClearActiveSession();
-            SparkSession.SetActiveSession(_spark);
-            Assert.IsType<SparkSession>(SparkSession.GetActiveSession());
-
             SparkSession.ClearDefaultSession();
             SparkSession.SetDefaultSession(_spark);
             Assert.IsType<SparkSession>(SparkSession.GetDefaultSession());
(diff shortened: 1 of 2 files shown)


imback82:
@rapoth @AFFogarty I also see `SparkSession.getActiveSession` is being used in `DeltaTable.isDeltaTable` and `DeltaTable.forPath`. Have you experienced any failures with `Could not find active SparkSession`?

imback82:
Another solution is to create a netty server with one working thread. The only concern is the perf, but it may not be too bad. I will also explore that route.

Regardless, I will merge this PR when it's ready to fix the flakiness of the E2E tests, and create an issue to follow up.

cc: @elvaliuliuliu @suhsteve @Niharikadutta

imback82:
I verified that the following would eventually fail:
```C#
        static void Main(string[] args)
        {
            SparkSession.Builder().GetOrCreate();
            for (int i = 0; i < 1000000; ++i)
            {
                DeltaTable.IsDeltaTable(".");
            }
        }
```
with the following message as expected:
```
java.lang.IllegalArgumentException: Could not find active SparkSession
        at io.delta.tables.DeltaTable$$anonfun$2.apply(DeltaTable.scala:665)
        at io.delta.tables.DeltaTable$$anonfun$2.apply(DeltaTable.scala:665)
        at scala.Option.getOrElse(Option.scala:121)
        at io.delta.tables.DeltaTable$.isDeltaTable(DeltaTable.scala:664)
        at io.delta.tables.DeltaTable.isDeltaTable(DeltaTable.scala)
```

PR #336, opened 2019-11-15
Use new List() in StructType constructor

AFFogarty:
Our current implementation of `StructType` has a constructor that looks like this:

``` 
public StructType(IEnumerable<StructField> fields)
{
    Fields = fields.ToList();
}
```

I checked-out what `ToList()` looks like under the hood and found that it just constructs a new `List` and adds each enumerable element one-by-one.  You can see the source [here](https://github.com/microsoft/referencesource/blob/17b97365645da62cf8a49444d979f94a59bbb155/System.Core/System/Linq/Enumerable.cs#L329).

Conversely, the [implementation](https://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,74) of `new List(IEnumerable<T> collection)` has an optimization that invokes native `Array.Copy()` when `collection` is an `ICollection`.

This PR changes the `StructType` constructor to use `Fields = new List<StructField>(fields);` so that we can take advantage of that optimization when `fields` is an `ICollection` (such as `List`, etc.)

### Measurement
I tested this with the `collection` param having 21474836 elements.  For each measurement, I did 300 iterations on .Net Core 3.0.

| Implementation  | `long[] collection`| `List<long> collection`|
| ------------- | ------------- |------------- |
| `ToList(collection)` (old) | 33331 ms | 66793 ms  |
| `new List(collection)` (new)  | 33493 ms  |33366 ms  |

This optimization consistently 2x faster when the `collection` param is a `List`.  For arrays, there is no difference. 

----------------------


suhsteve:
Good job doing the analysis.  But I have a question.  Why is the ToList() called in [your link](https://github.com/microsoft/referencesource/blob/17b97365645da62cf8a49444d979f94a59bbb155/System.Core/System/Linq/Enumerable.cs#L329) vs the extension method defined [here](https://github.com/microsoft/referencesource/blob/17b97365645da62cf8a49444d979f94a59bbb155/System.Core/System/Linq/Enumerable.cs#L947) ?

imback82:
For `StructType`, it will have 10s of fields. (Even thousand columns will be rare).

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -241,7 +240,7 @@ public sealed class StructType : DataType
         /// <param name="fields">A collection of StructFieldType objects</param>
         public StructType(IEnumerable<StructField> fields)
         {
-            Fields = fields.ToList();
+            Fields = new List<StructField>(fields);
```
imback82: @AFFogarty I also have the same question as @suhsteve. I see that the extension method is called, which is doing `return new List<TSource>(source);` Any thought?

PR #342, opened 2019-11-20
E2E Tests of DataType.FromJson for all data types

elvaliuliuliu:
This is a follow up PR of #248. When we expose `Schema()` to take in `StructType`, `DataType.FromJson` is exposed as well. 
In this PR, we aim to test `DataType.FromJson` behavior for all data types to make sure JMV side also works for all types.


src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        [InlineData("timestamp")]
+        [InlineData("double")]
+        [InlineData("float")]
+        [InlineData("byte")]
+        [InlineData("integer")]
+        [InlineData("long")]
+        [InlineData("short")]
+        public void TestSimpleTypes(string typeName)
+        {
+            JvmObjectReference atomicType = DataType.FromJson(Jvm, $@"""{typeName}""");
```
suhsteve: instead of passing a string, can you construct the DataType and call the .Json method instead ?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            var structType = new StructType(new[]
+            {
+                new StructField("age", new IntegerType()),
+                new StructField("name", new StringType())
+            });
+
+            JvmObjectReference jvmStructType = DataType.FromJson(Jvm, structType.Json);
+            Assert.IsType<JvmObjectReference>(jvmStructType);
+
+            // The following structField test failed for DataType.FromJson.
```
imback82: we can just say `StructField` is not supported.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            Assert.IsType<JvmObjectReference>(atomicType);
+        }
+
+        [Fact]
+        public void TestArrayType()
+        {
+            var arrayType = new ArrayType(new IntegerType());
+
+            JvmObjectReference jvmArrayType = DataType.FromJson(Jvm, arrayType.Json);
+            Assert.IsType<JvmObjectReference>(jvmArrayType);
```
imback82: We can just do
```C#
Assert.IsType<JvmObjectReference>(DataType.FromJson(Jvm, arrayType.Json))
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
imback82: yes, let's do that.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
elvaliuliuliu: Changed. Thanks.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
elvaliuliuliu: Changed. Thanks.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
elvaliuliuliu: I have changed to `ClassData`, so that we can pass `SimpleTypes` as `DataType`.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
```diff
@@ -14,23 +16,35 @@ public class TypesTests
...
-        [InlineData("integer")]
-        [InlineData("long")]
-        [InlineData("short")]
-        public void TestSimpleTypes(string typeName)
+        [ClassData(typeof(SimpleTypesGenerator))]
+        public void TestSimpleTypes(DataType simpleTypes)
         {
-            JvmObjectReference atomicType = DataType.FromJson(Jvm, $@"""{typeName}""");
-            Assert.IsType<JvmObjectReference>(atomicType);
+            Assert.IsType<JvmObjectReference>(DataType.FromJson(Jvm, simpleTypes.Json));
```
imback82: How about just create
```C#
private void Validate(DataType dataType)
{
    Assert.IsType<JvmObjectReference>(DataType.FromJson(Jvm, dataType.Json));
}
```

and just
```C#
Validate(new NullType());
Validate(new BinaryType());
```

imback82: You can reuse this in `TestArrayType`, etc.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        public void TestMapType()
+        {
+            var mapType = new MapType(new IntegerType(), new StringType());
+            Assert.IsType<JvmObjectReference>(DataType.FromJson(Jvm, mapType.Json));
+        }
+
+        [Fact]
+        public void TestStructType()
+        {
+            // Please note that StructField is not supported.
```
imback82: It's a bit confusing to put this comment here. You can put it at the end (outside function) and just say
```
// StructField is not tested because it cannot be converted from JSON by itself.
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
suhsteve: Do we need separate `[Fact]`s for ArrayType and MapType?  Can the type tests just be contained in one single test method ?

imback82: yea we discussed this earlier and these will be merged into one method.

elvaliuliuliu: Thanks for the suggestion! I have changed accordingly.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
elvaliuliuliu: Sure, added in the end. Thanks.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
imback82: sorry, now that we have one function, we can put at the end of the function:
```C#
            // The following validates for all ComplexTypes.
            Validate(new ArrayType(new IntegerType()));
            Validate(new MapType(new IntegerType(), new StringType()));
            Validate(new StructType(new[]
            {
                new StructField("age", new IntegerType()),
                new StructField("name", new StringType())
            }));

            // StructField is not tested because it cannot be converted from JSON by itself.
        }
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
```diff
@@ -0,0 +1,51 @@
...
+using Xunit;
+
+namespace Microsoft.Spark.E2ETest.IpcTests
+{
+    [Collection("Spark E2E Tests")]
+    public class TypesTests
+    {
+        private static IJvmBridge Jvm { get; } = SparkEnvironment.JvmBridge;
+
+        private void Validate(DataType dataType)
```
suhsteve: Since it's only one line, you can make this method an expression.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
```diff
@@ -0,0 +1,51 @@
...
+            // The following validates for all ComplexTypes.
+            Validate(new ArrayType(new IntegerType()));
+            Validate(new MapType(new IntegerType(), new StringType()));
+            Validate(new StructType(new[]
+            {
+                new StructField("age", new IntegerType()),
+                new StructField("name", new StringType())
+            }));
+        }
+        // StructField is not tested because it cannot be converted from JSON by itself.
```
suhsteve: this comment should be in the TestDataTypes() method shouldn't it ?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
elvaliuliuliu: Sure, moved inside the function. Thanks.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
elvaliuliuliu: Sure, changed. Thanks.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
elvaliuliuliu: Sure. moved inside. Thanks.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
```diff
@@ -14,10 +14,8 @@ public class TypesTests
     {
         private static IJvmBridge Jvm { get; } = SparkEnvironment.JvmBridge;
 
-        private void Validate(DataType dataType)
-        {
-            Assert.IsType<JvmObjectReference>(DataType.FromJson(Jvm, dataType.Json));
-        }
+        private void Validate(DataType dataType) => Assert.IsType<JvmObjectReference>(
+            DataType.FromJson(Jvm, dataType.Json));
```
suhsteve:
```csharp
        private void Validate(DataType dataType) =>
            Assert.IsType<JvmObjectReference>(DataType.FromJson(Jvm, dataType.Json));
```

elvaliuliuliu: Sure, changed. Thanks!

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/TypesTests.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.Spark.Interop.Ipc;
+using Microsoft.Spark.Sql.Types;
+using Xunit;
+
+namespace Microsoft.Spark.E2ETest.IpcTests
+{
+    [Collection("Spark E2E Tests")]
+    public class TypesTests
+    {
+        private static IJvmBridge Jvm { get; } = SparkEnvironment.JvmBridge;
```
suhsteve: @imback82 would it be better to use the SparkEnvironment to get the JvmBridge or the SparkFixture ?

imback82: you mean define `Jvm` in `SparkFixutre` first then retrieve?

suhsteve: Something like
```csharp
public TypesTests(SparkFixture fixture)
{
    _jvm = ((IJvmObjectReferenceProvider)fixture.Spark).Reference.Jvm;
    ...
}

suhsteve: Or do the above to define it in the `SparkFixture`.

imback82: If we are doing this, we want to define it in the `SparkFixture`.

elvaliuliuliu: So would you suggest to use SparkEnvironment or SparkFigure? Thanks.

suhsteve: I would say go ahead and use SparkFixture, adding the JVM property to it.

elvaliuliuliu: Sure, I have made the changes. Let me know if that makes sense to you. Thanks.

imback82:
Thanks @elvaliuliuliu!

PR #344, opened 2019-11-20
fix back end spelling

mairaw:
I noticed in the docs that back end was misspelled. It's either two words or hyphened when used as an adjective: https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/b/back-end

I've only changed text and comments, but not variable names.


imback82:
Can't `backend`,`back-end` and `back end` be used interchangeably? https://en.wikipedia.org/wiki/Back_end

mairaw:
I pointed to the Microsoft style guide above - that's our official guideline. Also, if you look at the examples under computing in the Wikipedia article, they follow the same rules as I described. But it's up to you whether you wanna take this change or not.

imback82 (approved):
LGTM

src/scala/microsoft-spark-2.4.x/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
```diff
@@ -147,7 +147,7 @@ object DotnetRunner extends Logging {
       } else {
         // scalastyle:off println
         println("***********************************************************************")
```
imback82: Can you update https://github.com/dotnet/spark/blob/master/src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala as well? That's causing the test failures.

imback82 (changes requested):

mairaw:
I rolled back some of the changes and kept only the ones in code comments and Markdown.

mairaw:
Closing this old PR. Doesn't look it will be taken.

PR #358, opened 2019-12-02
Use System.Text.Json for Json serialisation

### LINKED ISSUE #355: [FEATURE REQUEST]: Use System.Text.Json for Json serialisation
**Is your feature request related to a problem? Please describe.**
Currently this library uses Newtonsoft.Json for Json serialisation in IPC. [System.Text.Json](https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-apis/) was released recently and is faster and more memory efficient.

**Describe the solution you'd like**
For the Json serialisation to be done with System.Text.Json.

**Describe alternatives you've considered**
-

**Additional context**
I'd be happy to take this on if this seems like a good idea.

marcusturewicz:
This changes Json serialisation to use System.Text.Json rather than Newtonsoft.Json to hopefully improve CPU, memory and serialisation/deserialisation time.

4.6.0 is the first stable version of System.Text.Json and it requires at least version 4.5.3 of System.Memory (currently at 4.5.2) so I have increased the version for each project it is referenced in.

For the most part, the changes are just mapping from Newtonsoft.Json types to System.Text.Json types e.g. `JObject` -> `JsonElement` and so forth.

The most difficult part is in the `SortProperties()` method in `JsonSerDe`. I have also added a test for this method as I could not find one.

All tests seem to be passing locally.

There are a couple of items I'm not sure about, which I will comment on the line of code in the PR.

Fixes #355


src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,69 +2,90 @@
...
+                    {                     
+                        if (item.ValueKind == JsonValueKind.Object)
+                        {
+                            item.SortProperties().WriteTo(writer);
+                        }
+                        else if (item.ValueKind == JsonValueKind.Array)
+                        {
+                            item.SortArrayProperties().WriteTo(writer);
+                        }
+                        // TODO: is this required?
```
marcusturewicz: Is this required? It wasn't there before but what about other JsonValueKinds?

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -254,18 +260,21 @@ public StructType(IEnumerable<StructField> fields)
...
         /// <summary>
         /// Constructs a StructType object from a JSON object
         /// </summary>
         /// <param name="json">JSON object used to construct a StructType object</param>
         /// <returns>A StuructType object</returns>
-        private DataType FromJson(JObject json)
+        private DataType FromJson(JsonElement json)
         {
-            IEnumerable<JObject> fieldsJObjects = json["fields"].Select(f => (JObject)f);
+            // TODO: this might be incorrect, as it might need to be an object, not an array...?
```
marcusturewicz: I'm not sure about this one. It should probably be:
`.EnumerateObject`, but that returns `JsonProperty`'s. Any ideas?

src/csharp/Microsoft.Spark/Sql/Types/DataType.cs
```diff
@@ -56,10 +55,10 @@ public string Json
         {
             get
             {
-                object jObject = (JsonValue is JObject) ?
-                    ((JObject)JsonValue).SortProperties() :
+                object jObject = (JsonValue is JsonElement) ?
```
eerhardt: This can be re-written so you only have a single cast:

```
object jObject = (JsonValue is JsonElement jsonElement) ?
    jsonElement.SortProperties() :
    JsonValue;
```

eerhardt:
>         /// Parses a JToken object to construct a DataType.

Comments should be updated to not use Newtonsoft types.

---
Refers to: src/csharp/Microsoft.Spark/Sql/Types/DataType.cs:125 in 4e6465a. [](commit_id = 4e6465aced12d88f6579b179c77518b80982b8cb, deletion_comment = False)

src/csharp/Microsoft.Spark/Sql/Types/DataType.cs
eerhardt: Looking at this some more, `JsonValue` will never be a `JsonElement`. If you look at the complex types definition of JsonValue, they now all return anonymous types, which will never be `JsonElement`.

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,69 +2,90 @@
...
         {
-            if (jArray.Count == 0)
+            if (jArray.GetArrayLength() == 0)
             {
                 return jArray;
             }
 
-            var sortedJArray = new JArray();
-            foreach (JToken item in jArray)
+            using (var stream = new MemoryStream())
```
eerhardt: @ahsonkhan - is there a better recommendation for doing this (and `SortArrayProperties` below) rather than creating a new MemoryStream and writer for every level in the Json graph?

marcusturewicz: Yeah I didn't like that I used stream here. Investigating further, I think `ArrayBufferWriter<byte>` could be used instead, but it's only available from .NET Core 3.0. 

https://docs.microsoft.com/en-us/dotnet/api/system.buffers.arraybufferwriter-1?view=netcore-3.0

ahsonkhan: > is there a better recommendation for doing this (and `SortArrayProperties` below) rather than creating a new MemoryStream and writer for every level in the Json graph?

Depends on the goals here. How important is perf here and to what extent can we change the APIs (or implementation)?

You could try to implement a helper workhorse method which accepts a utf8jsonwriter and reset it after every flush.  I will work on providing a sample.

> Yeah I didn't like that I used stream here. Investigating further, I think `ArrayBufferWriter<byte>` could be used instead, but it's only available from .NET Core 3.0.

I understand it may not be ideal, but you could consider copying the specific implementation of the `IBufferWriter<T>` or provide your own. It doesn't use anything that is .NET Core 3.0 specific.

https://github.com/dotnet/runtime/blob/46da95eb75c052c828d67d14e46083e7ac2286b1/src/libraries/Common/src/System/Buffers/ArrayBufferWriter.cs#L17

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -51,21 +51,23 @@ public ArrayType(DataType elementType, bool containsNull = true)
         /// <summary>
         /// Returns JSON object describing this type.
         /// </summary>
-        internal override object JsonValue =>
-            new JObject(
-                new JProperty("type", TypeName),
-                new JProperty("elementType", ElementType.JsonValue),
-                new JProperty("containsNull", ContainsNull));
+        internal override object JsonValue => JsonDocument.Parse(JsonSerializer.Serialize(
```
ahsonkhan: For perf, consider using the `Utf8JsonWriter` rather than `JsonSerializer.Serialize` to create the JSON before passing it to `JsonDocument.Parse`.

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -155,24 +159,24 @@ public sealed class StructField
...
             DataType dataType,
             bool isNullable = true,
-            JObject metadata = null)
+            JsonElement? metadata = null)
         {
             Name = name;
             DataType = dataType;
             IsNullable = isNullable;
-            Metadata = metadata ?? new JObject();
+            Metadata = metadata ?? new JsonElement();
```
ahsonkhan: Keep in mind that `new JsonElement` is not the same as `new JObject()`.

```C#
JsonElement elem = new JsonElement();
Console.WriteLine(elem.ValueKind); // undefined
Console.WriteLine(elem.ToString()); // empty string

var elem = new JObject();
Console.WriteLine(elem.Type); // object
Console.WriteLine(elem.ToString()); // {}
```

I am not sure in what context this is used yet (or what it might break), but if it's important, you likely want to create a standalone `JsonElement` from "{}" and return that.

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -155,24 +159,24 @@ public sealed class StructField
             string name,
             DataType dataType,
             bool isNullable = true,
-            JObject metadata = null)
+            JsonElement? metadata = null)
```
ahsonkhan: Why is a nullable `JsonElement` being used? `JsonElement` is a struct. I would use `JsonElement metadata = default` instead.

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -203,12 +207,14 @@ internal StructField(JObject json)
         /// <summary>
         /// Returns JSON object describing this type.
         /// </summary>
-        internal object JsonValue =>
-            new JObject(
-                new JProperty("name", Name),
-                new JProperty("type", DataType.JsonValue),
-                new JProperty("nullable", IsNullable),
-                new JProperty("metadata", Metadata));
+        internal object JsonValue => JsonDocument.Parse(JsonSerializer.Serialize(
```
ahsonkhan: You need to dispose the `JsonDocument` and clone the `RootElement` to detach from the arraypool semantics of `JsonDocument`.

See https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.clone?view=netcore-3.0#System_Text_Json_JsonElement_Clone

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -253,19 +259,21 @@ public StructType(IEnumerable<StructField> fields)
         /// <summary>
         /// Returns JSON object describing this type.
         /// </summary>
-        internal override object JsonValue =>
-            new JObject(
-                new JProperty("type", TypeName),
-                new JProperty("fields", Fields.Select(f => f.JsonValue).ToArray()));
+        internal override object JsonValue => JsonDocument.Parse(JsonSerializer.Serialize(
```
ahsonkhan: Same here/elsewhere (dispose and clone).

src/csharp/Microsoft.Spark/Sql/Types/DataType.cs
```diff
@@ -56,10 +55,10 @@ public string Json
         {
             get
             {
-                object jObject = (JsonValue is JObject) ?
-                    ((JObject)JsonValue).SortProperties() :
+                object jObject = (JsonValue is JsonElement jsonElement) ?
+                    jsonElement.SortProperties() :
                     JsonValue;
-                return JsonConvert.SerializeObject(jObject, Formatting.None);
+                return JsonSerializer.Serialize(jObject);
```
ahsonkhan: Making `JsonElement` go through a sort phase and then a serialize phase seems inefficient since you are doing the work twice.

When will `JsonValue` not be a `JsonElement`? Maybe only call the serializer for that case and otherwise, just have a SortProperties helper that returns the string itself.

src/csharp/Microsoft.Spark/Sql/Types/DataType.cs
```diff
@@ -123,16 +122,15 @@ public override bool Equals(object obj)
...
+        /// Parses a JsonElement object to construct a DataType.
         /// </summary>
-        /// <param name="json">JToken object to parse</param>
+        /// <param name="json">JsonElement object to parse</param>
         /// <returns>The new DataType instance from the JSON string</returns>
-        internal static DataType ParseDataType(JToken json)
+        internal static DataType ParseDataType(JsonElement json)
         {
-            if (json.Type == JTokenType.Object)
+            if (json.ValueKind == JsonValueKind.Object)
```
ahsonkhan: What happens if `json` is `JsonValueKind.Array`?

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
ahsonkhan: > is there a better recommendation for doing this (and `SortArrayProperties` below) rather than creating a new MemoryStream and writer for every level in the Json graph?

Depends on the goals here. How important is perf here and to what extent can we change the APIs (or implementation)?

You could try to implement a helper workhorse method which accepts a utf8jsonwriter and reset it after every flush.  I will work on providing a sample.

> Yeah I didn't like that I used stream here. Investigating further, I think `ArrayBufferWriter<byte>` could be used instead, but it's only available from .NET Core 3.0.

I understand it may not be ideal, but you could consider copying the specific implementation of the `IBufferWriter<T>` or provide your own. It doesn't use anything that is .NET Core 3.0 specific.

https://github.com/dotnet/runtime/blob/46da95eb75c052c828d67d14e46083e7ac2286b1/src/libraries/Common/src/System/Buffers/ArrayBufferWriter.cs#L17

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,69 +2,85 @@
...
                 {
-                    sortedJArray.Add(arrayElem.SortProperties());
+                    writer.WriteStartArray();
+                    foreach (JsonElement item in jArray.EnumerateArray())
+                    {                     
+                        if (item.ValueKind == JsonValueKind.Object)
+                        {
+                            item.SortProperties().WriteTo(writer);
+                        }
+                        else if (item.ValueKind == JsonValueKind.Array)
```
ahsonkhan: What happens if a JSON array contains primitives like a list of strings? Looks like sort is ignoring them. Is that intentional?

```C#
string json = "{\"a\":1, \"c\": [1, 3, 5], \"b\": 3}";
using var doc = JsonDocument.Parse(json);
Console.WriteLine(doc.RootElement.SortProperties().ToString()); // {"a":1,"b":3,"c":[]}
```

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
ahsonkhan: This is an example of a more performant approach (and then we have other approaches that would be even better for perf, including using `ArrayBufferWriter<byte>` and avoiding Linq):
```C#
public static string SortProperties(this JsonElement jsonElement)
{
    using (var stream = new MemoryStream())
    {
        using (var writer = new Utf8JsonWriter(stream))
        {
            jsonElement.SortPropertiesCore(writer);
        }
        return Encoding.UTF8.GetString(stream.ToArray());
    }
}

private static void SortPropertiesCore(this JsonElement jsonElement, Utf8JsonWriter writer)
{
    switch (jsonElement.ValueKind)
    {
        case JsonValueKind.Undefined:
            throw new InvalidOperationException();
        case JsonValueKind.Object:
            jsonElement.SortObjectProperties(writer);
            break;
        case JsonValueKind.Array:
            jsonElement.SortArrayProperties(writer);
            break;
        case JsonValueKind.String:
        case JsonValueKind.Number:
        case JsonValueKind.True:
        case JsonValueKind.False:
        case JsonValueKind.Null:
            jsonElement.WriteTo(writer);
            break;
    };
}

private static void SortObjectProperties(this JsonElement jObject, Utf8JsonWriter writer)
{
    Debug.Assert(jObject.ValueKind == JsonValueKind.Object);

    writer.WriteStartObject();
    foreach (JsonProperty prop in jObject.EnumerateObject().OrderBy(p => p.Name))
    {
        writer.WritePropertyName(prop.Name);
        prop.Value.WriteElementHelper(writer);
    }
    writer.WriteEndObject();
}

private static void SortArrayProperties(this JsonElement jArray, Utf8JsonWriter writer)
{
    Debug.Assert(jArray.ValueKind == JsonValueKind.Array);

    writer.WriteStartArray();
    foreach (JsonElement item in jArray.EnumerateArray())
    {
        item.WriteElementHelper(writer);
    }
    writer.WriteEndArray();
}

private static void WriteElementHelper(this JsonElement item, Utf8JsonWriter writer)
{
    Debug.Assert(item.ValueKind != JsonValueKind.Undefined);

    if (item.ValueKind == JsonValueKind.Object)
    {
        item.SortObjectProperties(writer);
    }
    else if (item.ValueKind == JsonValueKind.Array)
    {
        item.SortArrayProperties(writer);
    }
    else
    {
        item.WriteTo(writer);
    }
}
```

``` ini

BenchmarkDotNet=v0.11.5.1159-nightly, OS=Windows 10.0.18362
Intel Core i7-6700 CPU 3.40GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.100-preview1-014459
  [Host]     : .NET Core 3.0.0 (CoreCLR 4.700.19.46205, CoreFX 4.700.19.46214), X64 RyuJIT
  Job-XVUECM : .NET Core 3.0.0 (CoreCLR 4.700.19.46205, CoreFX 4.700.19.46214), X64 RyuJIT

PowerPlanMode=00000000-0000-0000-0000-000000000000  IterationTime=250.0000 ms  MaxIterationCount=10  
MinIterationCount=5  WarmupCount=1  

```
|                  Method |     Mean |     Error |    StdDev |   Median |      Min |      Max | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------ |---------:|----------:|----------:|---------:|---------:|---------:|------:|--------:|-------:|------:|------:|----------:|
|                 SortOld | 3.171 us | 0.0593 us | 0.0154 us | 3.168 us | 3.150 us | 3.191 us |  1.00 |    0.00 | 0.5425 |     - |     - |    2304 B |
| SortOldWithoutSerialize | 2.499 us | 0.0618 us | 0.0409 us | 2.483 us | 2.453 us | 2.555 us |  0.79 |    0.02 | 0.5023 |     - |     - |    2120 B |
|                 SortNew | 1.746 us | 0.0468 us | 0.0279 us | 1.742 us | 1.708 us | 1.802 us |  0.55 |    0.01 | 0.4470 |     - |     - |    1888 B |
|              SortNewIBW | 1.656 us | 0.0488 us | 0.0323 us | 1.652 us | 1.614 us | 1.715 us |  0.52 |    0.01 | 0.3420 |     - |     - |    1456 B |
|        SortNewIBWCached | 1.584 us | 0.0389 us | 0.0231 us | 1.575 us | 1.557 us | 1.613 us |  0.50 |    0.01 | 0.2731 |     - |     - |    1144 B |
| SortNewIBWCached_Custom | 1.467 us | 0.0332 us | 0.0220 us | 1.462 us | 1.434 us | 1.505 us |  0.46 |    0.01 | 0.1103 |     - |     - |     464 B |

```C#
public class TestSortingProperties
{
    private JsonDocument _doc;
    private ArrayBufferWriter<byte> _output;

    [GlobalSetup]
    public void Setup()
    {
        string json = "{\"propertyA\": \"first\", \"propertyC\": 3, \"propertyB\": \"second\"}";

        _doc = JsonDocument.Parse(json);
        _output = new ArrayBufferWriter<byte>();
    }

    [GlobalCleanup]
    public void Cleanup()
    {
        _doc.Dispose();
    }

    [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
    [Benchmark(Baseline = true)]
    public string SortOld()
    {
        return JsonSerializer.Serialize(_doc.RootElement.SortProperties_old());
    }

    [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
    [Benchmark]
    public string SortOldWithoutSerialize()
    {
        return _doc.RootElement.SortProperties_old().ToString();
    }

    [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
    [Benchmark]
    public string SortNew()
    {
        return _doc.RootElement.SortProperties();
    }

    [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
    [Benchmark]
    public string SortNewIBW()
    {
        return _doc.RootElement.SortPropertiesIBW();
    }

    [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
    [Benchmark]
    public string SortNewIBWCached()
    {
        string str = _doc.RootElement.SortPropertiesIBW(_output);
        _output.Clear();
        return str;
    }

    [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
    [Benchmark]
    public string SortNewIBWCached_Custom()
    {
        string str = _doc.RootElement.SortPropertiesIBW_Custom(_output);
        _output.Clear();
        return str;
    }
}
```

It's also possible that a custom written sort might perform better than using linq (but that probably depends on the size of the payload and how deep the object graph gets and hence requires more careful perf analysis):
```C#
        [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
        [Benchmark(Baseline = true)]
        public int LinqSort()
        {
            int count = 0;
            foreach (JsonProperty prop in _doc.RootElement.EnumerateObject().OrderBy(p => p.Name))
            {
                count++;
            }

            return count;
        }

        [BenchmarkCategory(Categories.CoreFX, Categories.JSON)]
        [Benchmark]
        public int CustomSort()
        {
            int count = 0;
            var propertyNames = new List<string>();
            foreach (JsonProperty prop in _doc.RootElement.EnumerateObject())
            {
                propertyNames.Add(prop.Name);
            }
            propertyNames.Sort();

            foreach (string name in propertyNames)
            {
                count++;
            }

            return count;
        }
```

|     Method |     Mean |    Error |   StdDev |   Median |      Min |      Max | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------- |---------:|---------:|---------:|---------:|---------:|---------:|------:|--------:|-------:|------:|------:|----------:|
|   LinqSort | 868.7 ns | 24.30 ns | 14.46 ns | 868.3 ns | 847.0 ns | 887.4 ns |  1.00 |    0.00 | 0.1827 |     - |     - |     768 B |
| CustomSort | 656.4 ns | 52.15 ns | 34.50 ns | 648.8 ns | 616.7 ns | 729.9 ns |  0.76 |    0.04 | 0.0476 |     - |     - |     208 B |

ahsonkhan: And here are the results for the JSON string used in `TestSortProperties` from `GetTestJson`:

``` ini

BenchmarkDotNet=v0.11.5.1159-nightly, OS=Windows 10.0.18362
Intel Core i7-6700 CPU 3.40GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.100-preview1-014459
  [Host]     : .NET Core 3.0.0 (CoreCLR 4.700.19.46205, CoreFX 4.700.19.46214), X64 RyuJIT
  Job-JSWMCW : .NET Core 3.0.0 (CoreCLR 4.700.19.46205, CoreFX 4.700.19.46214), X64 RyuJIT

PowerPlanMode=00000000-0000-0000-0000-000000000000  IterationTime=250.0000 ms  MaxIterationCount=10  
MinIterationCount=5  WarmupCount=1  

```
|                  Method |     Mean |    Error |   StdDev |   Median |      Min |      Max | Ratio | RatioSD |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------------------ |---------:|---------:|---------:|---------:|---------:|---------:|------:|--------:|-------:|-------:|------:|----------:|
|                 SortOld | 65.20 us | 5.995 us | 3.135 us | 64.19 us | 61.62 us | 69.87 us |  1.00 |    0.00 | 9.2975 |      - |     - |  38.29 KB |
| SortOldWithoutSerialize | 57.35 us | 3.392 us | 2.018 us | 56.61 us | 55.18 us | 61.86 us |  0.88 |    0.05 | 9.4112 |      - |     - |  38.11 KB |
|                 SortNew | 21.85 us | 0.959 us | 0.570 us | 21.84 us | 21.21 us | 23.01 us |  0.34 |    0.01 | 4.4894 | 0.0880 |     - |  18.52 KB |
|              SortNewIBW | 21.41 us | 0.415 us | 0.108 us | 21.43 us | 21.24 us | 21.52 us |  0.33 |    0.02 | 4.0999 |      - |     - |   16.8 KB |
|        SortNewIBWCached | 21.93 us | 1.169 us | 0.612 us | 21.60 us | 21.35 us | 22.82 us |  0.34 |    0.02 | 2.9297 |      - |     - |  12.22 KB |
| SortNewIBWCached_Custom | 19.79 us | 0.406 us | 0.268 us | 19.85 us | 19.37 us | 20.13 us |  0.31 |    0.02 | 0.9728 |      - |     - |   4.03 KB |

marcusturewicz: Wow, thanks @ahsonkhan! SortNewIBWCached_Custom looks like the ticket! I will get this is in when I address the other changes.

ahsonkhan: Regarding `SortNewIBWCached_Custom`, I am not sure how well creating your own list scales for larger objects, or larger property strings (compared to using Linq to do the sorting), but here's what I had for that, in case you want to use that as a starting point.

Keep in mind, it still uses unbounded recursion, which is something we should consider changing by turning it into an iterative implementation using our own stack. Unless, there are existing guarantees that the `JsonElement` will not point to arbitrary depth payloads (like `[[[[....]]]`), the current impl could potentially result in stack overflow. That said, the `Utf8JsonWriter` currently can't write a `depth > 1_000` (which is a limitation of its implementation detail), which is why you likely won't hit a stackoverflow in common cases.

```C#
public static string SortPropertiesIBW_Custom(this JsonElement jsonElement, ArrayBufferWriter<byte> output)
{
    using (var writer = new Utf8JsonWriter(output))
    {
        jsonElement.SortPropertiesCore_Custom(writer);
    }
    return Encoding.UTF8.GetString(output.WrittenSpan);
}

private static void SortPropertiesCore_Custom(this JsonElement jsonElement, Utf8JsonWriter writer)
{
    switch (jsonElement.ValueKind)
    {
        case JsonValueKind.Undefined:
            throw new InvalidOperationException();
        case JsonValueKind.Object:
            jsonElement.SortObjectProperties_Custom(writer);
            break;
        case JsonValueKind.Array:
            jsonElement.SortArrayProperties_Custom(writer);
            break;
        case JsonValueKind.String:
        case JsonValueKind.Number:
        case JsonValueKind.True:
        case JsonValueKind.False:
        case JsonValueKind.Null:
            jsonElement.WriteTo(writer);
            break;
    };
}

private static void SortObjectProperties_Custom(this JsonElement jObject, Utf8JsonWriter writer)
{
    Debug.Assert(jObject.ValueKind == JsonValueKind.Object);

    var propertyNames = new List<string>();
    foreach (JsonProperty prop in jObject.EnumerateObject())
    {
        propertyNames.Add(prop.Name);
    }
    propertyNames.Sort();

    writer.WriteStartObject();
    foreach (string name in propertyNames)
    {
        writer.WritePropertyName(name);
        jObject.GetProperty(name).WriteElementHelper_Custom(writer);
    }
    writer.WriteEndObject();
}

private static void SortArrayProperties_Custom(this JsonElement jArray, Utf8JsonWriter writer)
{
    Debug.Assert(jArray.ValueKind == JsonValueKind.Array);

    writer.WriteStartArray();
    foreach (JsonElement item in jArray.EnumerateArray())
    {
        item.WriteElementHelper_Custom(writer);
    }
    writer.WriteEndArray();
}

private static void WriteElementHelper_Custom(this JsonElement item, Utf8JsonWriter writer)
{
    Debug.Assert(item.ValueKind != JsonValueKind.Undefined);

    if (item.ValueKind == JsonValueKind.Object)
    {
        item.SortObjectProperties_Custom(writer);
    }
    else if (item.ValueKind == JsonValueKind.Array)
    {
        item.SortArrayProperties_Custom(writer);
    }
    else
    {
        item.WriteTo(writer);
    }
}
```

marcusturewicz: Thanks. I'll try this for now. 

However, I'm having a problem with `Encoding.UTF8.GetString(ReadOnlySpan<byte>)`, the compiler is telling me it can't be found, but it is available in .NET Core 2.1.

image

https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.getstring?view=netcore-2.1#System_Text_Encoding_GetString_System_ReadOnlySpan_System_Byte__

ahsonkhan: Yes. The `ReadOnlySpan<byte>` overload was added in .NET Core 2.1 (and netstandard2.1), but the spark project targets netstandard2.0, so that API won't be available, unfortunately.

This means you either have to use the `byte[]` overload (by calling `ToArray()` on the span), or cross-compile and add #if/def to the code (similar to what I did here: https://github.com/dotnet/spark/pull/364/files#diff-f88fa8d0fd61f21045adfd6353902b8fR4)

That said, I don't know if that solution is viable/allowed and would defer to the maintainers of this project.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -26,9 +26,9 @@
   <ItemGroup>
     <PackageReference Include="Apache.Arrow" Version="0.14.1" />
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="System.Text.Json" Version="4.6.0" />
     <PackageReference Include="Razorvine.Pyrolite" Version="4.26.0" />
-    <PackageReference Include="System.Memory" Version="4.5.2" />
+    <PackageReference Include="System.Memory" Version="4.5.3" />
```
suhsteve: This can be removed since `System.Text.Json` pulls it in.

examples/Microsoft.Spark.FSharp.Examples/Microsoft.Spark.FSharp.Examples.fsproj
```diff
@@ -20,6 +20,10 @@
     <Compile Include="Program.fs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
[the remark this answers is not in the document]
suhsteve: Maybe we can add `NU1605` to the `NoWarn` of this file.

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,139 @@
...
+        }           
+
+        /// <summary>
+        /// Helper method to Parse an object into a JsonElement.
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <returns></returns>
+        public static JsonElement Parse(object obj)
+        {
+            return Parse(JsonSerializer.Serialize(obj));
```
marcusturewicz: Is there a better way than this? Seems quite inefficient. I thought of passing a `Dictionary<string, (JsonValueKind, object)>` in and then building the JsonElement with `Utf8JsonWriter` and writing the appropriate type but it seems a little verbose. Any ideas?

src/csharp/Microsoft.Spark/Interop/Ipc/ArrayBufferWriter.cs
```diff
@@ -0,0 +1,185 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+
+namespace System.Buffers
+{
+    /// <summary>
+    /// Represents a heap-based, array-backed output sink into which <typeparam name="T"/> data can be written.
+    /// </summary>
+    public class ArrayBufferWriter<T> : IBufferWriter<T>
```
ahsonkhan: nit: consider adding a comment for where this was copied from (in case some sync-ing is required in the future).

marcusturewicz:
.NET build/tests pass but the first E2E test fails. Not sure if that related to this change or not. 

https://dev.azure.com/dnceng/public/_build/results?buildId=456057

elvaliuliuliu:
Can you try to run the pipeline again to see if this will repro?

ahsonkhan:
Looks like a real test failure around `null`:
```text
[xUnit.net 00:00:45.64]     Microsoft.Spark.E2ETest.IpcTests.SparkSessionTests.TestSignaturesV2_4_X [SKIP]
  ! Microsoft.Spark.E2ETest.IpcTests.FunctionsTests.TestSignaturesV2_4_X [1ms]
  ! Microsoft.Spark.E2ETest.IpcTests.RuntimeConfigTests.TestSignaturesV2_4_X [1ms]
  ! Microsoft.Spark.E2ETest.IpcTests.SparkSessionTests.TestSignaturesV2_4_X [1ms]
[2019-12-13T02:45:15.5088912Z] [fv-az314] [Error] [JvmBridge] JVM method execution failed: Static method fromJson failed for class org.apache.spark.sql.types.DataType when called with 1 arguments ([Index=1, Type=String, Value=null], )
[2019-12-13T02:45:15.5089283Z] [fv-az314] [Error] [JvmBridge] java.lang.IllegalArgumentException: Failed to convert the JSON string 'null' to a data type.
```

```text
[2019-12-13T02:45:15.5090916Z] [fv-az314] [Exception] [JvmBridge] JVM method execution failed: Static method fromJson failed for class org.apache.spark.sql.types.DataType when called with 1 arguments ([Index=1, Type=String, Value=null], )
   at Microsoft.Spark.Interop.Ipc.JvmBridge.CallJavaMethod(Boolean isStatic, Object classNameOrJvmObjectReference, String methodName, Object[] args) in D:\a\1\s\src\csharp\Microsoft.Spark\Interop\Ipc\JvmBridge.cs:line 192
[xUnit.net 00:00:45.78]     Microsoft.Spark.E2ETest.IpcTests.TypesTests.TestDataTypes [FAIL]
  X Microsoft.Spark.E2ETest.IpcTests.TypesTests.TestDataTypes [59ms]
  Error Message:
   System.Exception : JVM method execution failed: Static method fromJson failed for class org.apache.spark.sql.types.DataType when called with 1 arguments ([Index=1, Type=String, Value=null], )
  Stack Trace:
     at Microsoft.Spark.Interop.Ipc.JvmBridge.CallJavaMethod(Boolean isStatic, Object classNameOrJvmObjectReference, String methodName, Object[] args) in D:\a\1\s\src\csharp\Microsoft.Spark\Interop\Ipc\JvmBridge.cs:line 192
   at Microsoft.Spark.Interop.Ipc.JvmBridge.CallJavaMethod(Boolean isStatic, Object classNameOrJvmObjectReference, String methodName, Object arg0) in D:\a\1\s\src\csharp\Microsoft.Spark\Interop\Ipc\JvmBridge.cs:line 113
   at Microsoft.Spark.Interop.Ipc.JvmBridge.CallStaticJavaMethod(String className, String methodName, Object arg0) in D:\a\1\s\src\csharp\Microsoft.Spark\Interop\Ipc\JvmBridge.cs:line 71
   at Microsoft.Spark.Sql.Types.DataType.FromJson(IJvmBridge jvm, String json) in D:\a\1\s\src\csharp\Microsoft.Spark\Sql\Types\DataType.cs:line 90
   at Microsoft.Spark.E2ETest.IpcTests.TypesTests.Validate(DataType dataType) in D:\a\1\s\src\csharp\Microsoft.Spark.E2ETest\IpcTests\Sql\TypesTests.cs:line 20
   at Microsoft.Spark.E2ETest.IpcTests.TypesTests.TestDataTypes() in D:\a\1\s\src\csharp\Microsoft.Spark.E2ETest\IpcTests\Sql\TypesTests.cs:line 27
```

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -10,6 +10,8 @@
     <Description>.NET for Apache Spark</Description>
     <PackageReleaseNotes>https://github.com/dotnet/spark/tree/master/docs/release-notes</PackageReleaseNotes>
     <PackageTags>spark;dotnet;csharp</PackageTags>
+    <!-- workaround due to including ArrayBufferWriter<T> source whilst TFM includes netstandard2.0 -->
+    <NoWarn>CS0436;$(NoWarn)</NoWarn>
```
ahsonkhan: Why are we getting this warning? I believe it is because of the netstandard2.1 TFM (ABW exists on netstandard2.1). If so, I would only include the file on older TFMs (conditionally include the file for specific TFMs where its needed) so that we don't have to workaround this warning.

I don't think we should be ignoring such warnings.

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
ahsonkhan: Is this `Parse` method necessary (i.e. one that takes in an `object`)? Can all the call-sites instead write to a writer rather than creating an object?

If that is too inconvenient, one way to make things faster is to have a utf-8 based implementation (which avoids the unnecessary transcoding cost from utf-16 string to utf-8 on both sides (the serialize and the parse):
```C#
return Parse(JsonSerializer.SerializeToUtf8Bytes(obj));

private static JsonElement Parse(byte[] utf8Json)
{
   using (var document = JsonDocument.Parse(utf8Json))
      return document.RootElement.Clone();
}
```

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,139 @@
...
-                }
+                item.SortObjectProperties(writer);
+            }
+            else if (item.ValueKind == JsonValueKind.Array)
+            {
+                item.SortArrayProperties(writer);
+            }
+            else
+            {
+                item.WriteTo(writer);
```
ahsonkhan: @suhsteve, @imback82 - can you chime in here?

The **previous** code was ignoring elements within an array. Why was that the case? Is it fine to continue ignoring it or should this be fixed?
```C#
string json = "{\"a\":1, \"c\": [1, 3, 5], \"b\": 3}";
using var doc = JsonDocument.Parse(json);
Console.WriteLine(doc.RootElement.SortProperties().ToString()); // {"a":1,"b":3,"c":[]}
```

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,139 @@
...
-                }
-                else if (item is JArray arrayElem)
-                {
-                    sortedJArray.Add(arrayElem.SortProperties());
-                }
+                item.SortObjectProperties(writer);
+            }
+            else if (item.ValueKind == JsonValueKind.Array)
+            {
+                item.SortArrayProperties(writer);
```
ahsonkhan: Also @suhsteve, @imback82

> Are we fine with these sorts of unbounded recursive calls, generally?

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
imback82: We can keep the previous behavior and do a follow up PR if needed; since this was only used in parsing datatype json, it wasn't needed.

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
imback82: It should be OK.

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,148 @@
...
+        }
+
+        /// <summary>
+        /// Helper method to Parse a byte array in UTF8 to a JsonElement.
+        /// </summary>
+        private static JsonElement Parse(byte[] utf8Json)
         {
-            var sortedJObject = new JObject();
-            foreach (JProperty property in jObject.Properties().OrderBy(p => p.Name))
+        using (var document = JsonDocument.Parse(utf8Json))
```
imback82: Please fix indentation.

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,148 @@
...
+            return document.RootElement.Clone();
```
imback82: https://github.com/dotnet/spark/blob/master/docs/coding-guidelines/csharp-coding-style.md#c-coding-style
`A single line statement block must go with braces.`

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -6,12 +6,15 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <IsPackable>true</IsPackable>
-
     <Description>.NET for Apache Spark</Description>
     <PackageReleaseNotes>https://github.com/dotnet/spark/tree/master/docs/release-notes</PackageReleaseNotes>
     <PackageTags>spark;dotnet;csharp</PackageTags>
   </PropertyGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.1' ">
```
imback82: @ahsonkhan @marcusturewicz Thanks for doing this. However, JSON parsing is not in a critical path, and do you think it's worthwhile to add an extra complexity for perf? How about getting the first version in with minimal changes and iterate on it if necessary? What do you guys think?

examples/Microsoft.Spark.FSharp.Examples/Microsoft.Spark.FSharp.Examples.fsproj
```diff
@@ -9,6 +9,8 @@
     <PublicSign>false</PublicSign>
     <!-- workaround https://github.com/dotnet/fsharp/issues/4822 to support VS 2017 -->
     <NoWarn>2003;$(NoWarn)</NoWarn>
+    <!-- workaround as suggested in https://github.com/dotnet/spark/pull/358/files#r355646101 -->
```
imback82: This link doesn't seem to work?

src/csharp/Microsoft.Spark.UnitTest/JsonSerDeTests.cs
```diff
@@ -0,0 +1,185 @@
...
+using Xunit;
+
+namespace Microsoft.Spark.UnitTest
+{
+    public class JsonSerDeTests
+    {
+        [Fact]
+        public void TestSortProperties()
+        {
+            // Arrange
```
imback82: I don't think `// Arrange`, `// Act`, and `//Assert` comments are necessary.

src/csharp/Microsoft.Spark.UnitTest/JsonSerDeTests.cs
```diff
@@ -0,0 +1,185 @@
...
+            var testJson = GetTestJson();
+            using (var document = JsonDocument.Parse(testJson))
+            {
+                var testElement = document.RootElement;
+                var expectedJson = GetExpectedSortedJson();
```
imback82: How about creating a test json and expected json in a tuple so that it's easier to compare them?

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
ahsonkhan: I don't have perf insights around usage of the component being updated here, so I would defer to the judgement of folks like yourself who own this space.

I would imagine cross-compiling would be important to take advantage of the latest improvements within the framework, at least for areas that are perf critical, so this helps set the precedent for it. If this is introducing extra complexity without clear benefit, then it makes sense to defer such changes.

> How about getting the first version in with minimal changes and iterate on it if necessary?

That seems like a viable way to make progress and iterate where we separate the porting from perf improvements. It makes sense to me.

> However, JSON parsing is not in a critical path, and do you think it's worthwhile to add an extra complexity for perf?

Mainly out of curiosity, can you expand on why JSON parsing is not perf critical for spark scenarios (for my edification)?

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,154 @@
...
         /// Note: Scala side uses JSortedObject when parsing JSON, so the properties
-        /// in JObject need to be sorted.
+        /// in JsonElement need to be sorted.
         /// <summary>
         /// Extension method to sort items in a JSON object by keys.
         /// </summary>
-        /// <param name="jObject"></param>
+        /// <param name="jsonElement"></param>
+        /// <returns></returns>
+        public static string SortProperties(this JsonElement jsonElement)
```
ahsonkhan: nit: Since this method is public (granted on an internal class), it might be worth noting as a remark in the xml docs that this uses recursion and shouldn't be used with untrusted/unbounded input data. 

Mainly to avoid folks calling `SortProperties` on arbitrary `JsonElement` in a different context (as is) than what it is currently used for (which is a closed set/bounded input).

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,155 @@
...
+            {
+                item.SortObjectProperties(writer);
+            }
+            else if (item.ValueKind == JsonValueKind.Array)
+            {
+                item.SortArrayProperties(writer);
+            }
+            else
+            {
+                item.WriteTo(writer);
```
ahsonkhan: Given what @imback82 said here (https://github.com/dotnet/spark/pull/358/files#r358410589), we should remove the else branch and not write anything here (for values within an array) in this PR.

That was the previous behavior. Maybe put a `Debug.Assert(item.ValueKind == JsonValueKind.Object || item.ValueKind == JsonValueKind.Array);`

If the scenario requires it, we should make this change in a separate PR.

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,155 @@
...
     /// </summary>
     internal static class JsonSerDe
     {
+
         /// Note: Scala side uses JSortedObject when parsing JSON, so the properties
-        /// in JObject need to be sorted.
+        /// in JsonElement need to be sorted.
+        /// <summary>
+        /// Extension method to sort items in a JSON object by keys. Please note that this 
+        /// method uses recursion and shouldn't be used with untrusted/unbounded input data.
```
ahsonkhan: nit: Instead of putting this in the `summary` tag, move it to a `remark` tag, leaving the `summary` tag as is.

```C#
/// <remarks>
/// Please note that this method uses recursion and shouldn't be used with 
/// non-trusted/unbounded input data.
/// </remarks>
```

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
marcusturewicz: Removing that line breaks the functionality, as it's required for sorting objects. So, I've kept it and added a flag to toggle its use when sorting objects vs arrays. I've therefore left the Debug Assert as is since other types can come through.

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
marcusturewicz: Ah yes, forgot about `<remarks>`, cheers

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,158 @@
...
-                if (item is JObject elem)
-                {
-                    sortedJArray.Add(elem.SortProperties());
-                }
-                else if (item is JArray arrayElem)
-                {
-                    sortedJArray.Add(arrayElem.SortProperties());
-                }
+                writer.WritePropertyName(name);
+                jObject.GetProperty(name).WriteElementHelper(writer, true);
```
ahsonkhan: nit: Use named parameters
```suggestion
                jObject.GetProperty(name).WriteElementHelper(writer, writePrimitives: true);
```

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,158 @@
...
+                item.WriteElementHelper(writer, false);
+            }
+            writer.WriteEndArray();
         }
-    }
 
+        /// <summary>
+        /// Helper to appropriately write Json types based on their kind.
+        /// </summary>
+        private static void WriteElementHelper(this JsonElement item, Utf8JsonWriter writer, bool writePrimities)
```
ahsonkhan: nit: typo
```suggestion
        private static void WriteElementHelper(this JsonElement item, Utf8JsonWriter writer, bool writePrimitives)
```

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,158 @@
...
+        /// </summary>
+        private static void SortArrayProperties(this JsonElement jArray, Utf8JsonWriter writer)
+        {
+            Debug.Assert(jArray.ValueKind == JsonValueKind.Array);
 
-            return sortedJArray;
+            writer.WriteStartArray();
+            foreach (JsonElement item in jArray.EnumerateArray())
+            {
+                item.WriteElementHelper(writer, false);
```
ahsonkhan: same here:
```suggestion
                item.WriteElementHelper(writer, writePrimitives: false);
```

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
ahsonkhan: Good point. That works.

src/csharp/Microsoft.Spark/Interop/Ipc/JsonSerDe.cs
```diff
@@ -2,70 +2,158 @@
...
+
+            if (item.ValueKind == JsonValueKind.Object)
+            {
+                item.SortObjectProperties(writer);
+            }
+            else if (item.ValueKind == JsonValueKind.Array)
+            {
+                item.SortArrayProperties(writer);
+            }
+            else if (writePrimities)
```
ahsonkhan:
```suggestion
            else if (writePrimitives)
```

Issue #351, opened 2019-11-25
[FEATURE REQUEST]: Expose SparkContext.setLogLevel API


imback82:
Expose `SparkContext.setLogLevel` API, which helps changing log levels programatically.

imback82:
closed by #360

PR #360, opened 2019-12-03
Expose SparkContext.setLogLevel API

suhsteve:
This PR addresses #351 , exposing `SparkContext.setLogLevel` API, which helps change log levels programmatically.

### DIFF
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/SparkContextTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/SparkContextTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using Microsoft.Spark.E2ETest.Utils;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


src/csharp/Microsoft.Spark.E2ETest/IpcTests/SparkContextTests.cs
```diff
@@ -21,11 +21,20 @@ public void TestSignaturesV2_3_X()
...
             SparkContext sc = SparkContext.GetOrCreate(new SparkConf());
 
-            _ = sc.GetConf();
-            _ = sc.DefaultParallelism;
+            Assert.IsType<SparkConf>(sc.GetConf());
+            Assert.IsType<int>(sc.DefaultParallelism);
 
             sc.SetJobDescription("job description");
 
+            sc.SetLogLevel(SparkContext.LogLevel.ALL);
```
imback82: actually, not sure if this experience is good. can we just use string?

suhsteve: Sure, changed to use string

imback82:
Thanks @suhsteve!

PR #362, opened 2019-12-03
Add default spark conf for Spark .NET applications

suhsteve:
Add default spark conf for Spark .NET applications.


imback82:
Thanks @suhsteve!

PR #363, opened 2019-12-04
Set the microsoft-spark version only in the parent pom.xml

suhsteve:
To help reduce the number of touched files during version updates, the individual microsoft-spark-<version> pom.xmls will inherit the version defined in the parent pom.xml


imback82:
merging to master, thanks @suhsteve!

PR #376, opened 2019-12-19
Support UDF that returns RowType

### LINKED ISSUE #117: [FEATURE REQUEST]: UDF to return custom business objects
I have a C# component which is presently running in Azure Data lake and i am planning to move to Spark and reuse the same component.
My example scenario

C# takes an input of Manager Dataset like

| mgrId | name |
| ------- | ------ |
| 11 | ABC |
| 22 | DEF |

C# component returns a List of Reportee, where Reportee is Defined as
`Class {
public int EmpId;
public string Name;
public string Role;
public int MgrId;
}`

Reportee dataset

| empId | name | role | mgrId |
| ----- |  ----- | ----- | ----- | 
| 100  | pqr |  admin | 11 |
| 200 | stu  |  reader | 11 |
| 300 | wxy  |  reader | 22 |

intended UDF 
`var udf = Udf<int, List<Reportee>>((mgrId) => return component.Execute(mgrId); });`

for each row in my Manager dataset, i have to call UDF to get final result in spark as

| mgrId | mgrname | empname | empid | Role |
| --- | --- | ---- | ---- | ---- |
| 11 | ABC | pqr | 100 | admin |
| 11 | ABC | stu | 200 | reader |
| 22 | DEF | wxy | 300 | reader |

elvaliuliuliu:
This PR exposes new API to support udf that returns RowType. It will fix #117.

Please see below as an example:
```C#
var df = spark.Range(0, 5);

Func<Column, Column> udf = Udf<int>(
    r => new GenericRow(new object[] { r + 100 }),
    new StructType(new[]
    {
        new StructField("id", new IntegerType())
    }));
          
// Assume that we have a df that has:
//+---+
// | id |
//+---+
// |  0 |
// |  1 |
// |  2 |
// |  3 |
// |  4 |
//+---+

var udfDf = df.Select(udf(df["id"]).As("udf_col"));

// PrintSchema() prints:
// root
// |-- udf_col: struct (nullable = true)
// |      |--  id: integer (nullable = true)
udfDf.PrintSchema();

// Show() prints:
// +--------+
// | udf_col|
// +--------+
// |  [100] |
// |  [101] |
// |  [102] |
// |  [103] |
// |  [104] |
// +--------+
udfDf.Show();
```
`Udf` that returns `GenericRow` looks like the following:
```C#
public static Func<Column, Column> Udf<T>(Func<T, GenericRow> udf, StructType returnType)
{
    return CreateUdf<GenericRow>(udf.Method.ToString(), UdfUtils.CreateUdfWrapper(udf), returnType).Apply1;
}
```


elvaliuliuliu:
@imback82 @suhsteve: For this WIP PR, I got the following error where `RunSimpleWorker()` has finished successfully, but I think error happened when sending bytes to C#(`java.lang.Object cannot be cast to net.razorvine.pickle.IObjectConstructor`). Any idea where could cause this error? Thanks!

image

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -158,25 +158,15 @@ public void TestUdfWithRowType()
...
-                }));
+            var schema = new StructType(new[]
+            {
+                new StructField("name", new StringType())
+            });
+            Func<Column, Column> udf = Udf<string>(
+                str => new GenericRow(new object[] { "Hello" + str }), schema);
+
+            Row[] rows = _df.Select(udf(_df["name"])).Collect().ToArray();
+            Assert.Equal(3, rows.Length);
```
AFFogarty: Can we check that `rows` contains the correct values?

src/csharp/Microsoft.Spark/Sql/Functions.cs
```diff
@@ -3797,6 +3798,188 @@ public static Func<Column> Udf<TResult>(Func<TResult> udf)
             return CreateUdf<TResult>(udf.Method.ToString(), UdfUtils.CreateUdfWrapper(udf)).Apply10;
         }
 
+        /// <summary>Creates a UDF from the specified delegate.</summary>
+        /// <param name="udf">The UDF function implementation.</param>
+        /// <param name="returnType">Schema associated with this row</param>
```
AFFogarty: I wonder if `returnType` isn't the best name for this thing.  The description "Schema associated with this row" wasn't super clear to me, so I had to dig into the code to understand that it's the schema of the row returned by the UDF.

src/csharp/Microsoft.Spark/Sql/Functions.cs
```diff
@@ -4088,7 +4277,7 @@ private static UserDefinedFunction CreateVectorUdf<TResult>(string name, Delegat
                     CommandSerDe.SerializedMode.Row,
                     CommandSerDe.SerializedMode.Row),
                 evalType,
-                UdfUtils.GetReturnType(typeof(TResult)));
+                returnType.Json ?? UdfUtils.GetReturnType(typeof(TResult)));
```
AFFogarty: Won't evaluating `returnType.Json` throw an exception if `returnType` is null?

src/csharp/Microsoft.Spark/Sql/Functions.cs
elvaliuliuliu: I was following `returnType` naming [here](https://github.com/dotnet/spark/blob/7fd0a0315d3935523633fc3e7a258f5e2e63d0eb/src/csharp/Microsoft.Spark/Sql/Expressions/UserDefinedFunction.cs#L20). Do you think `schema` would be better?

src/csharp/Microsoft.Spark/Sql/Functions.cs
elvaliuliuliu: If `returnType` is null, it will use `UdfUtils.GetReturnType(typeof(TResult)))` instead.

AFFogarty: Did you test this?  My understanding is that `??` will evaluate `returnType.Json` which will throw `NullReferenceException` if `returnType` is null.  But I might be wrong.

elvaliuliuliu: Yes, tests will pass when `returnType.Json` is null. `??` operator will return left-hand value which is `returnType.Json` here if it's not null, otherwise will evaluate and return right-hand value. Please refer to [this](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator).

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Sure, will add the test.

src/csharp/Microsoft.Spark/Sql/Functions.cs
AFFogarty: I mean when `returnType` is null, not `returnType.Json`.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -67,5 +68,57 @@ internal static object[] GetUnpickledObjects(Stream stream, int messageLength)
...
+                {
+                    SerDe.Write(stream, Opcodes.GLOBAL);
+                    SerDe.Write(stream, Encoding.UTF8.GetBytes(
+                        $"{_module}\n_create_row_inbound_converter\n"));
+                }
+                else
+                {
+                    if (!(o is GenericRow genericRow))
+                    {
+                        throw new InvalidOperationException("A GenericRow object is expected.");
```
Niharikadutta: In this exception can we add message for what is actually being passed instead of GenericRow?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -67,5 +68,57 @@ internal static object[] GetUnpickledObjects(Stream stream, int messageLength)
...
+                    SerDe.Write(stream, Opcodes.TUPLE);
+                    SerDe.Write(stream, Opcodes.REDUCE);
+                }
+            }
+        }
+
+        internal static Pickler CreatePickler()
+        {
+            new GenericRowPickler().Register();
+            return new Pickler();
```
Niharikadutta: Do we need to add unpickling on C# side for the returned Row object?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -67,5 +68,57 @@ internal static object[] GetUnpickledObjects(Stream stream, int messageLength)
...
+            {
+                if (o.Equals(this))
+                {
+                    SerDe.Write(stream, Opcodes.GLOBAL);
+                    SerDe.Write(stream, Encoding.UTF8.GetBytes(
+                        $"{_module}\n_create_row_inbound_converter\n"));
+                }
+                else
+                {
+                    if (!(o is GenericRow genericRow))
```
Niharikadutta: Can we use GetType() instead? Also do we need 'genericRow' variable name here?

elvaliuliuliu: Thanks for your comments! I have removed this part as per Terry's suggestion about pickling.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: Removed. Thanks!

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: Removed. Thanks!

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -165,6 +165,11 @@ private void WriteOutput(Stream stream, IEnumerable<object> rows, int sizeHint)
             if (s_outputBuffer == null)
                 s_outputBuffer = new byte[sizeHint];
 
+            if (rows.FirstOrDefault() is GenericRow)
+            {
+                rows = rows.Select(r => (object)(r as GenericRow).Values).AsEnumerable();
```
imback82: Can we double check we are not creating a separate list? cc: @suhsteve

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -158,25 +158,15 @@ public void TestUdfWithRowType()
...
-                    {
-                        new StructField("name", new StringType()),
-                    };
-                    var schema = new StructType(structFields);
-                    var row = new Row(new object[] { str }, schema);
-                    return row;
-                }));
+            var schema = new StructType(new[]
+            {
+                new StructField("name", new StringType())
```
imback82: can we extend this test to support more types?

src/csharp/Microsoft.Spark/Sql/Functions.cs
```diff
@@ -4088,7 +4277,7 @@ private static UserDefinedFunction CreateVectorUdf<TResult>(string name, Delegat
                     CommandSerDe.SerializedMode.Row,
                     CommandSerDe.SerializedMode.Row),
                 evalType,
-                UdfUtils.GetReturnType(typeof(TResult)));
+                returnType?.Json ?? UdfUtils.GetReturnType(typeof(TResult)));
```
imback82: Instead of mixing these two, we should have a separate function.

elvaliuliuliu: Sure, I can create another function maybe called `CreateUdfWithReturnType`. Does the naming of this function make sense?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
elvaliuliuliu: From what I observed is that `rows` become `Null` after this statement. `.Show()` will work but not `.Collect()`

imback82: did you look into why?

elvaliuliuliu: UdfDf.Collect() returns Null, will look more into GetRows function.

imback82: ah ok. then it's not this code here. Can you print out the schema? We may need to update pickler, and we should do it as a separate PR since it's covering an independent issue.

Can you try to repro this without udf? You can look at the schema where it fails and try to repro it with json data file.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Sure, I have added `IntegerType` as well.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
elvaliuliuliu: I can repro this without udf via reading json data file.

I have captured the exception and changed the test accordingly. Thanks!

src/csharp/Microsoft.Spark/Sql/Functions.cs
```diff
@@ -4091,6 +4279,22 @@ private static UserDefinedFunction CreateVectorUdf<TResult>(string name, Delegat
...
+            return UserDefinedFunction.Create(
+                name,
+                CommandSerDe.Serialize(
+                    execute,
+                    CommandSerDe.SerializedMode.Row,
+                    CommandSerDe.SerializedMode.Row),
+                evalType,
+                returnType.Json);
+        }
+
```
suhsteve: How about combining to something like:
```csharp
        private static UserDefinedFunction CreateUdf<TResult>(
            string name,
            Delegate execute,
            UdfUtils.PythonEvalType evalType) =>
            CreateUdf(name, execute, evalType, UdfUtils.GetReturnType(typeof(TResult)));

        private static UserDefinedFunction CreateUdf(
            string name,
            Delegate execute,
            UdfUtils.PythonEvalType evalType,
            StructType returnType) =>
            CreateUdf(name, execute, evalType, returnType.Json);

        private static UserDefinedFunction CreateUdf(
            string name,
            Delegate execute,
            UdfUtils.PythonEvalType evalType,
            string returnType) => UserDefinedFunction.Create(name,
                CommandSerDe.Serialize(
                    execute,
                    CommandSerDe.SerializedMode.Row,
                    CommandSerDe.SerializedMode.Row),
                evalType, returnType);
```

Then you can also change
- 1
    ```csharp
            public static Func<Column> Udf(Func<GenericRow> udf, StructType returnType)
            {
                return CreateUdf<GenericRow>(udf.Method.ToString(), UdfUtils.CreateUdfWrapper(udf), returnType).Apply0;
            }
    ```
    to
    ```csharp
            public static Func<Column> Udf(Func<Row> udf, StructType returnType)
            {
                return CreateUdf(udf.Method.ToString(), UdfUtils.CreateUdfWrapper(udf), returnType).Apply0;
            }
    ```

and
- 2 
    ```csharp
            private static UserDefinedFunction CreateUdf<TResult>(string name, Delegate execute, StructType returnType)
            {
                return CreateUdfWithReturnType<TResult>(name, execute, UdfUtils.PythonEvalType.SQL_BATCHED_UDF, returnType);
            }
    ```
    to
    ```csharp
            private static UserDefinedFunction CreateUdf(string name, Delegate execute, StructType returnType)
            {
                return CreateUdf(name, execute, UdfUtils.PythonEvalType.SQL_BATCHED_UDF, returnType);
            }
    ```

elvaliuliuliu: > Then you can also change
> 
> * 1
>   ```cs
>           public static Func<Column> Udf(Func<GenericRow> udf, StructType returnType)
>           {
>               return CreateUdf<GenericRow>(udf.Method.ToString(), UdfUtils.CreateUdfWrapper(udf), returnType).Apply0;
>           }
>   ```
>   
>   
>   to
>   ```cs
>           public static Func<Column> Udf(Func<Row> udf, StructType returnType)
>           {
>               return CreateUdf(udf.Method.ToString(), UdfUtils.CreateUdfWrapper(udf), returnType).Apply0;
>           }
>   ```
> 
Thanks for your suggestion. In the to function part, do you mean `Func<Row>` or Func<GenericRow> as below?
```cs
public static Func<Column> Udf(Func<GenericRow> udf, StructType returnType)
{
    return CreateUdf(udf.Method.ToString(), UdfUtils.CreateUdfWrapper(udf), returnType).Apply0;
}
 ```

Comment (suhsteve) [reply src/csharp/Microsoft.Spark/Sql/Functions.cs]: Func<GenericRow>

Comment (elvaliuliuliu) [reply src/csharp/Microsoft.Spark/Sql/Functions.cs]: Change. Thanks for your suggestion!

File: src/csharp/Microsoft.Spark/Sql/Functions.cs [hunk 11]
```diff
@@ -3797,6 +3798,188 @@ public static Func<Column> Udf<TResult>(Func<TResult> udf)
...
+        /// <summary>Creates a UDF from the specified delegate.</summary>
+        /// <typeparam name="T1">Specifies the type of the first argument to the UDF.</typeparam>
+        /// <typeparam name="T2">Specifies the type of the second argument to the UDF.</typeparam>
+        /// <param name="udf">The UDF function implementation.</param>
+        /// <param name="returnType">Schema associated with this row</param>
+        /// <returns>
+        /// A delegate that returns a <see cref="Column"/> for the result of the UDF.
+        /// </returns>
+        public static Func<Column, Column, Column> Udf<T1, T2>(Func<T1, T2, GenericRow> udf, StructType returnType)
```
suhsteve: line is too long, can you add a new line after the `(`

src/csharp/Microsoft.Spark/Sql/Functions.cs
```diff
@@ -4079,16 +4267,29 @@ private static UserDefinedFunction CreateVectorUdf<TResult>(string name, Delegat
...
         {
             return UserDefinedFunction.Create(
                 name,
                 CommandSerDe.Serialize(
                     execute,
                     CommandSerDe.SerializedMode.Row,
                     CommandSerDe.SerializedMode.Row),
-                evalType,
-                UdfUtils.GetReturnType(typeof(TResult)));
+                evalType, returnType);
```
suhsteve:
```suggestion
                evalType,
                returnType);
```

src/csharp/Microsoft.Spark/Sql/Functions.cs
elvaliuliuliu: Done.

src/csharp/Microsoft.Spark/Sql/Functions.cs
elvaliuliuliu: Changed.

suhsteve:
Looks like a few tests failed, can you fix those ?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -33,7 +34,29 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
 
                 foreach (object unpickled in unpickledObjects)
                 {
-                    yield return (unpickled as RowConstructor).GetRow();
+                    // yield return (unpickled as RowConstructor).GetRow();
+                    switch (unpickled)
```
imback82: Can you add comments what's happening?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -33,7 +34,29 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
...
+                    {
+                        case RowConstructor rc:
+                            yield return rc.GetRow();
+                            break;
+
+                        case object[] objs:
+                            if ((objs.Length != 1) || !(objs[0] is Row row))
```
imback82: Can you add comment why this is a special case? If you have two columns each with StructType, which case will handle it? (you may want to add a test for that as well).

elvaliuliuliu: Do you mean the schema should look like below?
```cs
{
    new StructField("col1", new StructType(new[]
    {
        new StructField("subCol1", new IntegerType())
    })),
    new StructField("col2", new StructType(new[]
    {
        new StructField("subCol2", new StringType())
    }))
});
```

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
elvaliuliuliu: This actually returns object array of row. I was wondering if this is caused by casting generic row to object?

imback82: Can we understand what's really happening before moving forward? (and add comments to the code)

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -165,8 +165,33 @@ private void WriteOutput(Stream stream, IEnumerable<object> rows, int sizeHint)
...
+            // Cast GenericRow to object if needed.
+            var newRows = new List<object>();
+            if (rows.FirstOrDefault() is GenericRow)
+            {
+                for (int i = 0; i < rows.Count(); ++i)
+                {
+                    object[] cols = (rows.ElementAt(i) as GenericRow).Values;
+                    if (cols.FirstOrDefault() is GenericRow)
+                    {
+                        object[] newCols = new object[cols.Length];
```
imback82: Don't you need to do this recursively? What happens you have a column with struct type in multiple levels?

suhsteve: Instead of making modifications here, how about registering a custom pickler for GenericRow.

imback82: @suhsteve I am a bit confused. Since this is used for `dumps`, wouldn't this require changes on Spark side? Can you provide snippet of what you are proposing?

suhsteve: I haven't tested it yet, but I'm thinking something like:

```csharp
public class GenericRowPickler : IObjectPickler {
    public void pickle(object o, Stream outs, Pickler currentPickler) {
        GenericRow row = (GenericRow)o;
        currentPickler.save(row.Values);
    }
}
```

And then register the customer pickler somewhere like:
```csharp
Pickler.registerCustomPickler(typeof(GenericRow), new GenericRowPickler());
```

imback82: Cool! @elvaliuliuliu, can you try this? This should work.

elvaliuliuliu: Thanks for the suggestion! I have made the changes accordingly. Let me know if any feedback. Thanks a lot!

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -158,25 +158,58 @@ public void TestUdfWithRowType()
...
+
+            Row[] rows1 = _df.Select(udf1(_df["name"])).Collect().ToArray();
+            Assert.Equal(3, rows1.Length);
+
+            foreach (Row row in rows1)
+            {
+                Assert.Equal(2, row.Size());
+                Assert.Equal(1, row.GetAs<int>("col1"));
+                Assert.Equal("abc", row.GetAs<string>("col2"));
+            }
```
suhsteve: Let's put these in {}'s and rename vars like `schema1`, `rows1`, etc to `schema`, `rows`, etc.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -165,7 +165,7 @@ private void WriteOutput(Stream stream, IEnumerable<object> rows, int sizeHint)
             if (s_outputBuffer == null)
                 s_outputBuffer = new byte[sizeHint];
 
-            Pickler pickler = s_pickler ?? (s_pickler = new Pickler(false));
+            Pickler pickler = s_pickler ?? (s_pickler = PythonSerDe.CreatePickler());
```
suhsteve: Is creating a new method to return a new pickler necessary ?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -67,5 +67,31 @@ internal static object[] GetUnpickledObjects(Stream stream, int messageLength)
...
+        }
+
+        /// <summary>
+        /// Register custom pickler for GenericRow object and create pickler.
+        /// </summary>
+        internal static Pickler CreatePickler()
+        {
+            new GenericRowPickler().Register();
+            return new Pickler();
+        }
```
suhsteve: Why are these new methods necessary ?  Can't the `GenericRowPickler` be registered in the same location where the `RowConstructor` is registered ?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -67,5 +67,31 @@ internal static object[] GetUnpickledObjects(Stream stream, int messageLength)
                 ArrayPool<byte>.Shared.Return(buffer);
             }
         }
+
+        /// <summary>
+        /// Custom pickler for GenericRow objects.
+        /// </summary>
+        internal class GenericRowPickler : IObjectPickler
```
suhsteve: Is this the best place to put this class definition ?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Sure, added!

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: I thought `PythonSerDe.cs` would be doing pickling and unpickling. Where would you suggest to put this class?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
elvaliuliuliu: Removed.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: Updated. Thanks!

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
elvaliuliuliu: I have added one line comments. Let me know if this makes to you. Thanks!

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
elvaliuliuliu: Nested case test has been added.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -158,25 +158,61 @@ public void TestUdfWithRowType()
...
-            Assert.Throws<ArgumentException>(() => Udf<string, Row>(
-                (str) =>
+            // Single GenericRow
+            {
+                var schema = new StructType(new[]
+                {
+                    new StructField("col1", new IntegerType()),
+                    new StructField("col2", new StringType())
+                });
+                Func<Column, Column> udf1 = Udf<string>(
```
suhsteve:
```suggestion
                Func<Column, Column> udf = Udf<string>(

```

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -158,25 +158,61 @@ public void TestUdfWithRowType()
...
+                var subSchema2 = new StructType(new[]
+                {
+                    new StructField("subCol2", new StringType())
+                });
+                var schema = new StructType(new[]
+                {
+                    new StructField("col1", subSchema1),
+                    new StructField("col2", subSchema2)
+                });
+                Func<Column, Column> udf2 = Udf<string>(
```
suhsteve:
```suggestion
                Func<Column, Column> udf = Udf<string>(

```

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -158,25 +158,61 @@ public void TestUdfWithRowType()
...
-                    };
-                    var schema = new StructType(structFields);
-                    var row = new Row(new object[] { str }, schema);
-                    return row;
-                }));
+                    Assert.Equal(2, row.Size());
+                    Assert.IsType<Row>(row.Get("col1"));
+                    Assert.IsType<Row>(row.Get("col2"));
+                    Assert.Equal(new Row(new object[] { 1 }, subSchema1), row.GetAs<Row>("col1"));
+                    Assert.Equal(new Row(new object[] { "abc" }, subSchema2), row.GetAs<Row>("col2"));
```
suhsteve: This line is too long

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
suhsteve: Why not put it in a file called `GenericRowPickler.cs`

elvaliuliuliu: Sounds good. Added it to `GenericRowPickler.cs`.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -158,25 +158,63 @@ public void TestUdfWithRowType()
...
-                    {
-                        new StructField("name", new StringType()),
-                    };
-                    var schema = new StructType(structFields);
-                    var row = new Row(new object[] { str }, schema);
-                    return row;
-                }));
+                    Assert.Equal(2, row.Size());
+                    Assert.IsType<Row>(row.Get("col1"));
+                    Assert.IsType<Row>(row.Get("col2"));
```
suhsteve: These aren't necessary.

elvaliuliuliu: Removed.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -158,25 +158,61 @@ public void TestUdfWithRowType()
...
+                {
+                    new StructField("subCol1", new IntegerType())
+                });
+                var subSchema2 = new StructType(new[]
+                {
+                    new StructField("subCol2", new StringType())
+                });
+                var schema = new StructType(new[]
+                {
+                    new StructField("col1", subSchema1),
```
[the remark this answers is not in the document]
elvaliuliuliu: There is single GenericRow example above with non-StructType field. Do you mean like [this](https://github.com/dotnet/spark/blob/e7d9d1d7f4ee791b011684a3d5b5df9c256adec7/src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs#L163)?

src/csharp/Microsoft.Spark/Sql/GenericRowPickler.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// Custom pickler for GenericRow objects.
+    /// </summary>
+    internal class GenericRowPickler : IObjectPickler
+    {
+        public void pickle(object o, Stream outs, Pickler currentPickler)
+        {
+            GenericRow row = (GenericRow)o;
+            currentPickler.save(row.Values);
```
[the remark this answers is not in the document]
elvaliuliuliu: Sure, merged to one line.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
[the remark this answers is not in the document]
elvaliuliuliu: Resolved as per discussion. Thanks!

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -33,7 +34,29 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
...
-                    yield return (unpickled as RowConstructor).GetRow();
+                    switch (unpickled)
+                    {
+                        case RowConstructor rc:
+                            yield return rc.GetRow();
+                            break;
+
+                        // Unpickled object contains single Row
+                        case object[] objs:
+                            if ((objs.Length != 1) || !(objs[0] is Row row))
```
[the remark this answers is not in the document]
elvaliuliuliu: Resolved as per discussion. Thanks!

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
elvaliuliuliu: Resolved as per discussion. Thanks!

Issue #208, opened 2019-08-14
[FEATURE REQUEST]: Foreach and Foreachbatch output sinks


jatcar95:
I am unable to find a way to use Foreach and Foreachbatch sinks with .NET for Spark. Are these supported features, and if not, are there plans to support them in the future?

Here is the Spark documentation I'm referring to: http://spark.apache.org/docs/latest/structured-streaming-programming-guide.html#using-foreach-and-foreachbatch

imback82:
@jatcar95 thanks for the suggestion. This is currently not supported, but we are open to support this feature depending on the community interests. (We are also very open to accept PRs to support this).

inxfidxel:
Would love to see this feature :)

dannydwarren:
One of the struggles we are running into is joining new data in a Kafka stream with data from the beginning of the stream. We think we have a pattern in the Scala API proven out to write to Postgres via JDBC inside of the `.foreachBatch` then joining the batch with the Postgres table immediately after and reducing all the records for the key in Postgres down to a single record and publishing that reduced record to a different stream in Kafka.

We've spent the last day trying to do this in C#, but without the `.foreachBatch` option the C# API is lacking key functionality we need.

AFFogarty:
Please assign to me.

imback82:
@AFFogarty Can you share the initial proposal?

imback82:
@suhsteve can you share the initial proposal when it's ready?

imback82:
These APIs are now available as of v0.12.

PR #387, opened 2020-01-08
Expose DataStreamWriter.Foreach API

suhsteve:
This PR exposes the `DataStreamWriter.Foreach` API.

#208 

Users can use this API by creating a `[Serializable]` class that implements the following interface:
```csharp
    public interface IForeachWriter
    {
        bool Open(long partitionId, long epochId);

        void Process(Row row);

        void Close(Exception errorOrNull);
    }
```

This user-defined class will be wrapped in a Wrapper that will call its respective methods according to the spark ForeachWriter lifecycle specifications.  The lifecycle of the methods are as follows:

```
For each partition with partitionId:
    For each batch/epoch of streaming data(if its streaming query) with epochId:
        Method Open(partitionId, epochId) is called.
        If Open returns true:
            For each row in the partition and batch/epoch, method Process(row) is called.
        Method Close(errorOrNull) is called with error(if any) seen while processing rows.
```

An example of a IForeachWriter can be something like the following:
```csharp
        [Serializable]
        private class TestForeachWriter : IForeachWriter
        {
            [NonSerialized]
            private StreamWriter _streamWriter;

            private readonly string _writePath;

            public TestForeachWriter(string writePath)
            {
                _writePath = writePath;
            }

            public void Close(Exception errorOrNull)
            {
                _streamWriter?.Dispose();
            }

            public bool Open(long partitionId, long epochId)
            {
                try
                {
                    _streamWriter= new StreamWriter(
                        Path.Combine(
                            _writePath,
                            $"sink-foreachWriter-{Guid.NewGuid()}.csv"));
                    return true;
                }
                catch
                {
                    return false;
                }
            }

            public void Process(Row value)
            {
                _streamWriter.WriteLine(string.Join(",", value.Values));
            }
```


src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -59,5 +63,93 @@ public void TestSignaturesV2_3_X()
...
+            Assert.Equal(
+                expected.Select(i => new object[] { i }),
+                foreachWriterOutputDF.Collect().Select(r => r.Values));
+        }
+
+        [Serializable]
+        private class TestForeachWriter : IForeachWriter
+        {
+            [ThreadStatic]
+            private static StreamWriter s_streamWriter;
```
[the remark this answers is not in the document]
suhsteve: I think this  may not be as important for Windows or the unit tests, but if this was running on Linux and if the worker was running multiple `TestForeachWriter` tasks in parallel, then they may end up using the same `StreamWriter` and write to the same file instead of different files.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -181,7 +181,8 @@ private sealed class UdfWrapperData
             List<UdfSerDe.UdfData> udfs)
         {
             UdfSerDe.UdfData udfData = UdfSerDe.Serialize(func);
-            if (udfData.MethodName != UdfWrapperMethodName)
+            if (udfData.MethodName != UdfWrapperMethodName ||
+                !Attribute.IsDefined(func.Target.GetType(), typeof(UdfWrapperAttribute)))
```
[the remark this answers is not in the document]
suhsteve: I believe using an empty `interface` to mark classes is a type of antipattern.  Marking classes with attributes is the suggested route.  https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1040?view=vs-2019#rule-description

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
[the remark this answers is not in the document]
imback82: can you capture your intention in the comment?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -59,5 +63,93 @@ public void TestSignaturesV2_3_X()
...
+            {
+                try
+                {
+                    s_streamWriter = new StreamWriter(
+                        Path.Combine(_writePath,$"sink-foreachWriter-{Guid.NewGuid()}.csv"));
+                    return true;
+                }
+                catch
+                {
+                    return false;
```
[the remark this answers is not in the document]
suhsteve: Added tests for:
- `Open` returns `false`
- `Process` throws `Exception`.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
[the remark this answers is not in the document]
suhsteve: Added to comments.

src/csharp/Microsoft.Spark/TaskContext.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// TaskContext stores information related to a task.
+    /// </summary>
+    internal class TaskContext
+    {
+        [ThreadStatic]
+        internal static TaskContext s_taskContext;
+
+        internal TaskContext()
+        {
+            s_taskContext = this;
```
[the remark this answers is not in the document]
suhsteve: Added `TaskContextHolder`

src/csharp/Microsoft.Spark/TaskContext.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    // TaskContextHolder contains the TaskContext for the current Thread.
+    internal static class TaskContextHolder
+    {
+        [ThreadStatic]
+        internal static TaskContext s_taskContext;
+
+        internal static TaskContext Get() => s_taskContext;
+
+        internal static void Set(TaskContext tc) => s_taskContext = tc;
```
suhsteve: @imback82 would you prefer these _Get/Set_ methods or do you prefer this to be a _Property_ named `TaskContext`?  If it's a property then I will have to make this a `ThreadLocal<TaskContext>` to get the same effect as `[ThreadStatic]`

src/csharp/Microsoft.Spark/TaskContext.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    /// <summary>
+    /// TaskContext stores information related to a task.
+    /// </summary>
+    internal class TaskContext
+    {
+        internal TaskContext()
+        {
+            TaskContextHolder.Set(this);
+        }
```
suhsteve: @imback82 do you prefer this to be set here or at the site where the TaskContext object gets created.

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
```diff
@@ -0,0 +1,184 @@
...
+    /// Interface for writing custom logic to process data generated by a query. This is
+    /// often used to write the output of a streaming query to arbitrary storage systems.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Any implementation of this interface will be used by Spark in the following way:
+    /// <list type="bullet">
+    /// <item>
+    /// <description>
+    /// A single instance of this class is responsible of all the data generated by a single task
```
Niharikadutta:
```suggestion
    /// A single instance of this class is responsible for all the data generated by a single task
```

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
```diff
@@ -0,0 +1,184 @@
...
+    /// in a query.In other words, one instance is responsible for processing one partition of the
```
Niharikadutta:
```suggestion
    /// in a query. In other words, one instance is responsible for processing one partition of the
```

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
suhsteve: At the moment this matches the spark description.  I'll leave this as is.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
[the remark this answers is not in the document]
imback82: sounds good. thanks!

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -59,5 +63,205 @@ public void TestSignaturesV2_3_X()
...
+            // When IForeachWriter.Process throws an Exception, then the exception
+            // is rethrown by ForeachWriterWrapper. We will limit the partitions
+            // to 1 to make validating this scenario simpler. Verify that:
+            // - Triggering stream throws an exception.
+            // - 1 CSV file will be created in the temporary directory.
+            // - 1 Exception will be created in the temporary directory. The
+            //   thrown exception from Process() will be sent to Close().
+            // - The CSV file will not contain valid data to read.
+            TestAndValidateForeach(
+                    streamInputPath,
```
imback82: indentation looks off.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -59,5 +63,205 @@ public void TestSignaturesV2_3_X()
...
+            DataStreamWriter dsw = _spark
+                .ReadStream()
+                .Schema("id INT")
+                .Json(streamInputPath)
+                .Repartition(partitions)
+                .WriteStream()
+                .Foreach(foreachWriter);
+
+            // Trigger the stream batch once.
+            if (foreachThrows)
```
imback82: looks like you can do `if (expectedExceptionFiles > 0)`?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -59,5 +63,205 @@ public void TestSignaturesV2_3_X()
...
+            // on successful calls to the ForeachWriter.Process method.
+            DataFrame foreachWriterOutputDF = _spark
+                .Read()
+                .Schema("id INT")
+                .Csv(dstTempDirectory.Path)
+                .Sort("id");
+
+            // Validated expected *.csv data.
+            Assert.Equal(
+                    expectedOutput.Select(i => new object[] { i }),
```
imback82: indentation looks off

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -59,5 +63,205 @@ public void TestSignaturesV2_3_X()
...
+                    1,
+                    1,
+                    Enumerable.Empty<int>());
+        }
+
+        private void TestAndValidateForeach(
+            string streamInputPath,
+            TestForeachWriter foreachWriter,
+            bool foreachThrows,
+            int partitions,
```
imback82: can't you just use `expectedCSVFiles` instead? (or use `partitions` in place of `expectedCSVFiles`)

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -59,5 +63,205 @@ public void TestSignaturesV2_3_X()
...
+                    foreachWriterOutputDF.Collect().Select(r => r.Values));
+        }
+
+        [Serializable]
+        private class TestForeachWriter : IForeachWriter
+        {
+            // Mark the StreamWriter as ThreadStatic. If there are multiple Tasks
+            // running this ForeachWriter, on the same Worker, and in parallel, then
+            // it may be possible that they may use the same StreamWriter object. This
+            // may cause an unintended side effect of a Task writing the output to a
```
imback82: Can we just remove this side effect? You can use locks such that there is always only one `StreamWriter` per file.

imback82: I also see this comment ` A single instance of this class is responsible of all the data generated by a single task`. Do we need to handle thread issue here?

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
```diff
@@ -0,0 +1,184 @@
...
+        internal ForeachWriterWrapper(IForeachWriter foreachWriter) =>
+            _foreachWriter = foreachWriter;
+
+        internal IEnumerable<object> Execute(int partitionId, IEnumerable<Row> rows)
+        {
+            if (!TaskContextHolder.Get().LocalProperties.TryGetValue(
+                "streaming.sql.batchId",
+                out string epochIdStr) || !long.TryParse(epochIdStr, out long epochId))
+            {
+                throw new Exception("Could not get batch id from TaskContext");
```
imback82: it could also have failed because of `TryParse`.

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
```diff
@@ -0,0 +1,184 @@
...
+            }
+
+            Exception error = null;
+            bool opened = _foreachWriter.Open(partitionId, epochId);
```
imback82: you should be getting a new (copied) instance of _foreachWriter right?

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        internal ForeachWriterWrapperUdfWrapper(
+            Func<int, IEnumerable<Row>, IEnumerable<object>> func)
+        {
+            _func = func;
+        }
+
+        internal IEnumerable<object> Execute(int pid, IEnumerable<object> input)
+        {
+            return _func(
```
imback82: can you add a comment what's happening here?

src/csharp/Microsoft.Spark/TaskContext.cs
imback82: I prefer the latter to be explicit.

src/csharp/Microsoft.Spark/TaskContext.cs
```diff
@@ -0,0 +1,96 @@
...
+                return Key.GetHashCode();
+            }
+        }
+    }
+
+    // TaskContextHolder contains the TaskContext for the current Thread.
+    internal static class TaskContextHolder
+    {
+        [ThreadStatic]
+        internal static TaskContext s_taskContext;
```
imback82: Let's add a short comment why this needs to be thread static.

src/csharp/Microsoft.Spark/TaskContext.cs
imback82: Either one is fine to me.

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
imback82: related to `RowDeserializer`?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
suhsteve: Yeah, sounds good.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
suhsteve: Yup!

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
suhsteve: Thanks

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
suhsteve: Originally I this a static field because the `StreamWriter` was not marked as `[Serializable]`.  But since we are creating the `StreamWriter` object in the `open` call, marking the field as `[NonSerialized]` addresses the issue.

suhsteve: > I also see this comment ` A single instance of this class is responsible of all the data generated by a single task`. Do we need to handle thread issue here?

If all the fields of the ForeachWriter are non static then I don't think there will be issues, but if there are static fields then I think it'll be up to the user to determine how to handle them.

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
suhsteve: good catch.

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
suhsteve: A `ForeachWriterWrapper` object will be deserialized for every task.  Since the state of the `_foreachWriter` will be set during the initial `serialization`, then each task will have its own copy of `ForeachWriterWrapper` and by extension its own copy of `_foreachWriter`.

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
suhsteve: Yes this is related to `RowDeserializer`.  The command execution goes through the `Spark.Utils.UdfUtils.PythonEvalType.NON_UDF` path, which expects a RDD.WorkerFunction.ExecuteDelegate Delegate.

src/csharp/Microsoft.Spark/TaskContext.cs
suhsteve: left as is.

src/csharp/Microsoft.Spark/TaskContext.cs
suhsteve: Registered the `TaskContext` with `TaskContextHolder` in `PayloadProcessor`

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
imback82: I meant to add it to the code :)

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
```diff
@@ -0,0 +1,190 @@
...
+                pid,
+                input.Cast<object[]>().SelectMany(
```
imback82: Would cast here will materialize the `IEnumerable` to array or no?

src/csharp/Microsoft.Spark/TaskContext.cs
imback82: ok.

src/csharp/Microsoft.Spark/RDD/Collector.cs
```diff
@@ -83,5 +86,16 @@ public object Deserialize(Stream stream, int length)
...
         }
+
+        /// <summary>
+        /// Deserializer for Pickled Rows.
+        /// </summary>
+        private sealed class RowDeserializer : IDeserializer
+        {
+            public object Deserialize(Stream stream, int length)
+            {
+                return PythonSerDe.GetUnpickledObjects(stream, length);
```
imback82: Can you add a comment where `RowCollector.GetRow()` is called?

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
suhsteve: I updated the `<summary>` portion of the class, instead of the `Execute` method.

src/csharp/Microsoft.Spark/RDD/WorkerFunction.cs
```diff
@@ -37,20 +37,21 @@ public WorkerFunction(ExecuteDelegate func)
             WorkerFunction outerFunction)
         {
             return new WorkerFunction(
-                new WrokerFuncChainHelper(
```
imback82: This is a breaking change right?

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
suhsteve: Cast will call `CastIterator<RowConstructor>`, which returns an `IEnumerable<RowConstructor>`.  The casting won't occur unless the IEnumerable is materialized to an array or the elements are individually process through some foreach call.

For more details:
- https://github.com/microsoft/referencesource/blob/17b97365645da62cf8a49444d979f94a59bbb155/System.Core/System/Linq/Enumerable.cs#L1031-L1040
- https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.cast?view=netframework-4.8#remarks

src/csharp/Microsoft.Spark/RDD/WorkerFunction.cs
suhsteve: Oh is it ?  I only renamed it because there was a typo in the name of this method.

imback82: I don't we can use the previous worker b/c of this, but it may be ok b/c this was only for RDD. Can you update the compatibility scenario in the description and you may need to update the internal worker version as well?

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
imback82: ok. also, you expect the following right? the size of `object[]` is 1? If so, can we be explicitly about it?

suhsteve: The linq query doesn't make any assumptions regarding the size of `object[]`. It should work if `object[]` is 1 or more, as long as the contents of the `object`s within the `object[]` are `RowConstructor`'s

imback82: it's not about linq. you expect one row in your udf, no?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -59,5 +63,190 @@ public void TestSignaturesV2_3_X()
...
+                expectedExceptionFiles,
+                Directory.GetFiles(dstTempDirectory.Path, "*.exception").Length);
+
+            // Read in the *.csv file(s) generated by the TestForeachWriter.
+            // If there are multiple input files, sorting by "id" will make
+            // validation simpler. Contents of the *.csv will only be populated
+            // on successful calls to the ForeachWriter.Process method.
+            DataFrame foreachWriterOutputDF = _spark
+                .Read()
+                .Schema("id INT")
```
imback82: can we have a test with more complex type? (multi columns, at least one col is struct type, etc.)

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
imback82: I don't think generalizing the solution is a good idea here.

suhsteve: The `ForeachWriterWrapper.Execute()` (_func in this case) expects an `IEnumerable<Row>`

imback82: right, so object[] should have one element, so I would assert here to be explicit.

suhsteve: RowConstructor.GetRow() is now called from the `RowDeserializer` with a comment regarding `AutoBatchedPickler` which determines number of Rows being pickled.

suhsteve: Also added comment to `ForeachWriterWrapperUdfWrapper.Execute` which also mentions `AutoBatchedPickler` as well as its use in the `PythonForeachWriter.scala` class.

src/csharp/Microsoft.Spark/RDD/Collector.cs
suhsteve: `RowConstructor.GetRow()` is now called from the `RowDeserializer` with a comment regarding `AutoBatchedPickler` which determines number of Rows being pickled.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
suhsteve: Added more columns and also a `StructType` as a column.

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                _foreachWriter.Close(error);
+
+                if (error != null)
+                {
+                    throw error;
+                }
+            }
+
+            return Enumerable.Empty<object>();
```
imback82: Can we add a short comment why we are returning empty? (e.g., empty IEnumerable is returned because ForEach is a sink operation, but something needs to be returned to work within the UDF framework).

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
imback82: question, do we need this check? looks like all the UDF functions are `Execute', no? Or are you adding this for robustness?

suhsteve: If a user defines a udf with the method named `Execute` then this will cause errors.  ie, if we look at `ForeachWriterWrapper.Execute` this is the actual Udf we want to use, and the `ForeachWriterWrapperUdfWrapper.Execute` is wrapping it.

src/csharp/Microsoft.Spark/Sql/ForeachWriter.cs
imback82: Yea, now it looks much better, thanks!

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
imback82: Then do we still need this check `udfData.MethodName != UdfWrapperMethodName`?

suhsteve: It should be safe to remove that as well.

src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs
```diff
@@ -181,7 +181,7 @@ private sealed class UdfWrapperData
             List<UdfSerDe.UdfData> udfs)
         {
             UdfSerDe.UdfData udfData = UdfSerDe.Serialize(func);
-            if (udfData.MethodName != UdfWrapperMethodName)
+            if (!Attribute.IsDefined(func.Target.GetType(), typeof(UdfWrapperAttribute)))
```
imback82: remove `UdfWrapperMethodName` constant as well?

suhsteve: oh good catch.

suhsteve: Nm.  It's being used in the `CreateUdfWrapperDelegate` method.

imback82: OK. in this case, we need the `(udfData.MethodName != UdfWrapperMethodName)` check to ensure the name matches up for the wrapper in `CreateUdfWrapperDelegate`.  Can we also rename `ForeachWriterWrapper.Execute` to `ForeachWriterWrapper.Process` or something like that to differentiate? We can leave the attribute check for robustness.

suhsteve: Sure, made the changes.

imback82:
@elvaliuliuliu we need to update the breaking change section in release notes with this change.

PR #404, opened 2020-01-26
Fix UDF returning StructType when it is the only column.

imback82:
This is a follow up to #376 to address the following issue:
```C#
var schema = new StructType(new[] {
    new StructField("col1", new IntegerType()),
    new StructField("col2", new StringType()) });
    
var udf = Udf<string>(str => new GenericRow(new object[] { 1, "abc" }), schema);

Row[] rows = _df.Select(udf(_df["name"]).As("col")).Collect().ToArray();
Assert.Equal(3, rows.Length);
rows[0].GetAs<Row>("col"); // This would fail!
```


imback82:
@elvaliuliuliu will create a separate PR for this.

PR #214, opened 2019-08-20
Modify RowConstructor to work with WithColumn

suhsteve:
This allows nested Rows with nested StructTypes to be properly unpickled as well as allows Rows to be passed to Udfs by using the WithColumn method.

```csharp
            _df = _spark
                .Read()
                .Schema("age INT, name STRING")
                .Json($"{TestEnvironment.ResourceDirectory}people.json");

            Func<Column, Column> AgeIsNull = Udf<Row, bool>(
                r =>
                {
                    var age = r.GetAs<int?>("age");
                    return !age.HasValue;
                });

            string[] allCols = _df.Columns().ToArray();
            DataFrame dummyColDF =
                _df.WithColumn("DummyCol", Struct(allCols[0], allCols.Skip(1).ToArray()));
            DataFrame ageIsNullDF =
                dummyColDF.WithColumn("AgeIsNull", AgeIsNull(dummyColDF["DummyCol"]));
```

Produces the following Schema for the `ageIsNullDF` Dataframe

```json
{
    "type": "struct",
    "fields": [
        {
            "name": "age",
            "type": "integer",
            "nullable": true,
            "metadata": {}
        },
        {
            "name": "name",
            "type": "string",
            "nullable": true,
            "metadata": {}
        },
        {
            "name": "DummyCol",
            "type": {
                "type": "struct",
                "fields": [
                    {
                        "name": "age",
                        "type": "integer",
                        "nullable": true,
                        "metadata": {}
                    },
                    {
                        "name": "name",
                        "type": "string",
                        "nullable": true,
                        "metadata": {}
                    }
                ]
            },
            "nullable": false,
            "metadata": {}
        },
        {
            "name": "AgeIsNull",
            "type": "boolean",
            "nullable": true,
            "metadata": {}
        }
    ]
}
```

__Benchmark__:
[HDInsight](https://azure.microsoft.com/en-us/pricing/details/hdinsight/)
D14v2 16 CPU, 112GB Ram
Spark 2.3.2
tpch query 1
dataset size 2.8GB Parquet (generated with db-gen with a 9 scale factor)
`master` -> `72fdf4834d5d9e7df06efec29233734aaec9445b`
`PR` -> `882cdbe9ac8f657163a6777a72577db45f1b0154`

- local[14]
  
    | Iteration | Time(ms) `master`| Time(ms) `PR`|
    |-----------|------------------|--------------|
    |1|90607|83569|
    |2|93520|88445|
    |3|95719|91437|
    |4|89949|78545|
[...]

PR #406, opened 2020-01-28
Follow up on UDF that takes in and returns Row

elvaliuliuliu:
This is a follow up of #376 and #214 to remove the materialization inside `construct()`. And change it on the worker side. This would make the behavior consistent for UDF that takes in Row and that returns Row.

### DIFF
--- a/src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
@@ -141,8 +141,7 @@ namespace Microsoft.Spark.E2ETest.UdfTests
                 Func<Column, Column> udf = Udf<Row, string>(
                     (row) =>
                     {
-                        string city = row.GetAs<string>("city");
-                        return $"{city}";
+                        return row.GetAs<string>("city");
                     });
 
                 Row[] rows = _df.Select(udf(_df["info1"])).Collect().ToArray();
@@ -153,7 +152,7 @@ namespace Microsoft.Spark.E2ETest.UdfTests
                 Assert.Equal(expected, actual);
             }
 
-            // Row is a part of top-level column.
+            // Multiple Rows
             {
                 Func<Column, Column, Column, Column> udf = Udf<Row, Row, string, string>(
                     (row1, row2, str) =>
@@ -163,7 +162,10 @@ namespace Microsoft.Spark.E2ETest.UdfTests
                         return $"{str}:{city},{state}";
                     });
 
-                Row[] rows = _df.Select(udf(_df["info1"], _df["info2"], _df["name"])).Collect().ToArray();
+                Row[] rows = _df
+                    .Select(udf(_df["info1"], _df["info2"], _df["name"]))
+                    .Collect()
+                    .ToArray();
                 Assert.Equal(3, rows.Length);
 
                 var expected = new[] {
@@ -180,8 +182,7 @@ namespace Microsoft.Spark.E2ETest.UdfTests
                     (row) =>
                     {
                         Row outerCol = row.GetAs<Row>("company");
-                        string job = outerCol.GetAs<string>("job");
-                        return $"{job}";
+                        return outerCol.GetAs<string>("job");
                     });
 
                 Row[] rows = _df.Select(udf(_df["info3"])).Collect().ToArray();
--- a/src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
@@ -139,10 +139,7 @@ namespace Microsoft.Spark.E2ETest.UdfTests
             // Single Row
             {
                 Func<Column, Column> udf = Udf<Row, string>(
-                    (row) =>
-                    {
-                        return row.GetAs<string>("city");
-                    });
+                    (row) => row.GetAs<string>("city"));
 
                 Row[] rows = _df.Select(udf(_df["info1"])).Collect().ToArray();
                 Assert.Equal(3, rows.Length);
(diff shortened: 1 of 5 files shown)


src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
```diff
@@ -65,15 +65,6 @@ public object construct(object[] args)
...
 
-            // When a row is ready to be materialized, then construct() is called
-            // on the RowConstructor which represents the row.
-            if ((args.Length == 1) && (args[0] is RowConstructor rowConstructor))
-            {
-                // Construct the Row and return args containing the Row.
-                args[0] = rowConstructor.GetRow();
-                return args;
-            }
-
```
suhsteve: Can you also get rid of the loop in the .GetRow() method.  Move the materialization of the inner rows to the Row.Convert() method.

imback82: Why is that?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -136,20 +136,61 @@ public void TestUdfWithReturnAsMapType()
...
 
-            var expected = new[] { "Burdwan,Paschimbanga", "Los Angeles,California", "Seattle," };
-            string[] actual = rows.Select(x => x[0].ToString()).ToArray();
-            Assert.Equal(expected, actual);
+                var expected = new[] { "Burdwan", "Los Angeles", "Seattle" };
+                string[] actual = rows.Select(x => x[0].ToString()).ToArray();
+                Assert.Equal(expected, actual);
+            }
+
+            // Row is a part of top-level column.
```
imback82: This comment is a bit misleading. We can say UDF that takes multiple Rows, etc.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -136,20 +136,61 @@ public void TestUdfWithReturnAsMapType()
...
-                    string city = row.GetAs<string>("city");
-                    string state = row.GetAs<string>("state");
-                    return $"{city},{state}";
-                });
+            // Single Row
+            {
+                Func<Column, Column> udf = Udf<Row, string>(
+                    (row) =>
+                    {
+                        string city = row.GetAs<string>("city");
```
imback82: just `return row.GetAs<string>("city");`

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -136,20 +136,61 @@ public void TestUdfWithReturnAsMapType()
...
+                Assert.Equal(expected, actual);
+            }
+
+            // Nested Row
+            {
+                Func<Column, Column> udf = Udf<Row, string>(
+                    (row) =>
+                    {
+                        Row outerCol = row.GetAs<Row>("company");
+                        string job = outerCol.GetAs<string>("job");
```
imback82: just return this.

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
suhsteve: This should save us from looping through the columns of the Row twice.  Since Row.Convert is already looping through the field types, it should be okay to do the conversion there.

imback82: I would like to contain the `RowConstructor` calls as much as possible (we already had few bugs around this). Also, it's much easier to understand and consistent now that `RowConstructor` is used close to  `PythonSerDe.GetUnpickledObjects`, one in `RowCollector` and one in `PicklingSqlCommandExecutor`.

If looping through columns twice is a concern (I would think it would be dominated by materialization anyway and we already have a lot of room for improvements on that code path), we can do that as a separate PR.

suhsteve: Sure we can leave as is then.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -136,20 +136,62 @@ public void TestUdfWithReturnAsMapType()
...
-                    string city = row.GetAs<string>("city");
-                    string state = row.GetAs<string>("state");
-                    return $"{city},{state}";
-                });
+            // Single Row
+            {
+                Func<Column, Column> udf = Udf<Row, string>(
+                    (row) =>
+                    {
+                        return row.GetAs<string>("city");
```
suhsteve: nit:  remove the {}'s and the return.  just make this a single expression.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Sure, I have updated to `Multiple Rows`.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
```diff
@@ -65,15 +65,6 @@ public object construct(object[] args)
                 s_schemaCache = new Dictionary<string, StructType>();
             }
 
-            // When a row is ready to be materialized, then construct() is called
```
imback82: This is actually a breaking change, but @suhsteve is updating the worker version in his PR: https://github.com/dotnet/spark/pull/387/files

Issue #407, opened 2020-01-28
[FEATURE REQUEST]: Change the API of UDF that returns Row


elvaliuliuliu:
Currently, the UDF that returns Row using GenericRow and Schema as below:
```csharp
public static Func<Column, Column> Udf<T>(Func<T, GenericRow> udf, StructType returnType)
{
    return CreateUdf(udf.Method.ToString(), UdfUtils.CreateUdfWrapper(udf), returnType).Apply1;
}
```
This will cause issues for chaining UDFs. We need to modify the API to return Row.

elvaliuliuliu:
I have the following chained UDF in PySpark:
```python
# create dataframe
df = spark.range(0,3);
schema = StructType([StructField("col1", IntegerType(), True), StructField("col2", StringType(), True)])
# UDF that returns Row
def udfFun1(s):
    return [1, "abc"]
udf1 = F.udf(udfFun1, StructType(schema))
udfDf1 = df.select(udf1(df["id"]))

# printSchema() prints
# root
# |-- udfFun1(id): struct (nullable = true)
# |    |-- col1: integer (nullable = true)
# |    |-- col2: string (nullable = true)
udfDf1.printSchema()

# show() prints
# +-----------+
# |udfFun1(id)|
# +-----------+
# |   [1, abc]|
# |   [1, abc]|
# |   [1, abc]|
# +-----------+
udfDf1.show()

# UDF that takes in Row
def udfFun2(s):
    return "hello"
udf2 = F.udf(udfFun2, StringType())
# Chained UDF
udfDf2 = df.select(udf2(udf1(df["id"])))

# printSchema() prints
# root
# |-- udfFun2(udfFun1(id)): string (nullable = true)
udfDf2.printSchema()

# show() prints
# +--------------------+
# |udfFun2(udfFun1(id))|
# +--------------------+
# |               hello|
# |               hello|
# |               hello|
# +--------------------+
udfDf2.show()
```
Can we tell if the Row generated by udf1 has a schema by this example?

imback82:
Can't you update `udfFun2` to look into `s`, which should be Row object?

elvaliuliuliu:
> Can't you update `udfFun2` to look into `s`, which should be Row object?

As per offline discussion, `Schema` is not passed in chained UDF in PySpark.

elvaliuliuliu:
This has been resolved via #411.

PR #411, opened 2020-02-01
Fix the chained UDF with Row type

elvaliuliuliu:
This is a follow PR of #376 which will fix the following chained UDF issue with Row type (#407):
```csharp
    # Chained UDFs with GenericRow type.
    {
	{
	    new StructField("col1", new IntegerType()),
	    new StructField("col2", new StringType())
	});
	Func<Column, Column> udf1 = Udf<string>(
	    str => new GenericRow(new object[] { 1, "abc" }), schema);

	Func<Column, Column> udf2 = Udf<Row, string>(
	    row => row.GetAs<string>(1));

	Row[] rows = _df.Select(udf2(udf1(_df["name"]))).Collect().ToArray();
	Assert.Equal(3, rows.Length);

	var expected = new[] { "abc", "abc", "abc" };
	string[] actual = rows.Select(x => x[0].ToString()).ToArray();
	Assert.Equal(expected, actual);
    }
```
We compared with PySpark UDFs and found that `Schema` is not passed in chained UDF cases. So schema-less `Row` is created to handle chained UDF.


src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -197,15 +197,15 @@ public void TestUdfWithRowType()
...
+            // Single Row.
             {
                 var schema = new StructType(new[]
                 {
                     new StructField("col1", new IntegerType()),
                     new StructField("col2", new StringType())
                 });
                 Func<Column, Column> udf = Udf<string>(
-                    str => new GenericRow(new object[] { 1, "abc" }), schema);
+                    str => new Row(new object[] { 1, "abc" }), schema);
```
imback82: Can we allow GenericRow to be returned as well?

So I am thinking the user can either do
```C#
    str => new GenericRow(new object[] { 1, "abc" }), schema);
```
or
```C#
    str => new Row(new object[] { 1, "abc" }, schema), schema);
```
Can you check if the latter example works? (to see if schema is serialized correctly)

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -219,14 +219,14 @@ public void TestUdfWithReturnAsRowType()
...
-            // Generic row is a part of top-level column.
+            // Row is a part of top-level column.
             {
                 var schema = new StructType(new[]
                 {
                     new StructField("col1", new IntegerType())
                 });
                 Func<Column, Column> udf = Udf<string>(
-                    str => new GenericRow(new object[] { 111 }), schema);
+                    str => new Row(new object[] { 111 }), schema);
```
imback82: This example is not good since this `Row` constructor is internal.

src/csharp/Microsoft.Spark.UnitTest/Sql/RowTests.cs
```diff
@@ -136,7 +136,7 @@ public void RowCollectorTest()
         private Pickler CreatePickler()
         {
             new StructTypePickler().Register();
-            new RowPickler().Register();
+            new Microsoft.Spark.UnitTest.TestUtils.RowPickler().Register();
```
imback82: just `TestUtils.RowPickler`?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
imback82: You probably need this: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/user-defined-conversion-operators

src/csharp/Microsoft.Spark.UnitTest/Sql/RowTests.cs
elvaliuliuliu: Removed. Thanks!

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: The `Row` constructor for the following example will also be internal, right? Same point as your below comment.
```csharp
    str => new Row(new object[] { 1, "abc" }, schema), schema);
```

imback82: Actually, let's be conservative and just allow only the following:
```C#
    str => new GenericRow(new object[] { 1, "abc" }), schema);
```
, and make the user defined conversion operator as internal as well.

The argument is that we don't want to allow users to create a schemaless `Row`. (We should also comment that schemaless Rows can happen within chained UDFs (same behavior as PySpark).)

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Changed to `GenericRow`.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Looks like I cannot make user-defined operator as internal.

After adding the user-defined conversion operator, the following works:
```csharp
GenericRow genericRow = new GenericRow(new object[] { 111 });
Row row = (Row)genericRow;
```

src/csharp/Microsoft.Spark/Sql/RowPickler.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System.IO;
+using Razorvine.Pickle;
+
+namespace Microsoft.Spark.Sql
+{
+    /// <summary>
+    /// Custom pickler for Row objects.
+    /// </summary>
+    internal class RowPickler : IObjectPickler
```
imback82: Can we move GenericRowPickler here and change the file name to `CustomPicklers.cs`?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -197,7 +197,7 @@ public void TestUdfWithRowType()
         [Fact]
         public void TestUdfWithReturnAsRowType()
         {
-            // Single GenericRow
+            // Single GenericRow.
```
imback82:
```C#
// Test UDF that returns a Row object. 
```

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -219,7 +219,7 @@ public void TestUdfWithReturnAsRowType()
                 }
             }
 
-            // Generic row is a part of top-level column.
+            // GenericRow is a part of top-level column.
```
imback82: What's the difference b/w this test and the one right above?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -244,7 +244,7 @@ public void TestUdfWithReturnAsRowType()
                 }
             }
 
-            // Nested GenericRow
+            // Nested GenericRows.
```
imback82: // Test UDF that returns a Row object.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -36,6 +36,9 @@ static PythonSerDe()
             Unpickler.registerConstructor(
                 "pyspark.sql.types", "_create_row_inbound_converter", s_rowConstructor);
 
+            // Register custom pickler for Row objects.
```
imback82: Let's combine this to
```C#
// Register custom picklers
Pickler.registerCustomPickler(typeof(GenericRow), new GenericRowPickler());
Pickler.registerCustomPickler(typeof(Row), new RowPickler());
```

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Select udf returned column and another column to test top-level column. I guess we can remove this. What do you think?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: I have changed to `// Test UDF that returns a nested Row object.`.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: Changed!

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs

src/csharp/Microsoft.Spark/Sql/RowPickler.cs
elvaliuliuliu: Changed. Thanks!

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
imback82: I think this was related to a bug we had before when we had one column in the `Row` object.
Can we change the comments to
```C#
// Test UDF that returns a Row object with multiple columns
```
and
```C#
// Test UDF that returns a Row object with a single column
```
?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -295,6 +295,27 @@ public void TestUdfWithReturnAsRowType()
...
+                Func<Column, Column> udf1 = Udf<string>(
+                    str => new GenericRow(new object[] { 1, "abc" }), schema);
+
+                Func<Column, Column> udf2 = Udf<Row, string>(
+                    row => row.GetAs<string>(1));
+
+                Row[] rows = _df.Select(udf2(udf1(_df["name"]))).Collect().ToArray();
+                Assert.Equal(3, rows.Length);
+
+                var expected = new[] { "abc", "abc", "abc" };
```
imback82: Can we update the UDF to relay the name?
For example,
```C#
Func<Column, Column> udf1 = Udf<string>(
    str => new GenericRow(new object[] { 1, str }), schema);
```

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -295,6 +295,27 @@ public void TestUdfWithReturnAsRowType()
...
+                string[] actual = rows.Select(x => x[0].ToString()).ToArray();
```
imback82: Can we make sure each row has only column?

src/csharp/Microsoft.Spark/Sql/Row.cs
```diff
@@ -36,6 +34,24 @@ internal Row(object[] values, StructType schema)
...
+        /// </summary>
+        /// <param name="genericRow">GenericRow to convert from</param>      
+        internal Row(GenericRow genericRow)
+        {
+            _genericRow = genericRow;
+        }
+
+        /// <summary>
+        /// Returns schema-less Row which can happen within chained UDFs (same behavior as PySpark).
+        /// </summary>
```
imback82: Can we add
```C#
/// <remarks>
/// The use of this conversion operator is discouraged except for the UDF that returns s Row object.
/// </remarks>
```

src/csharp/Microsoft.Spark/Sql/Row.cs
```diff
@@ -36,6 +34,24 @@ internal Row(object[] values, StructType schema)
             Convert();
         }
 
+        /// <summary>
+        /// Constructor for the schema-less Row class used for chained UDFs.
+        /// </summary>
+        /// <param name="genericRow">GenericRow to convert from</param>
```
imback82: there is no conversion happening here.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Sounds good. Comments changed.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Sure, changed.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Test added.

src/csharp/Microsoft.Spark/Sql/Row.cs
elvaliuliuliu: Added.

src/csharp/Microsoft.Spark/Sql/Row.cs
elvaliuliuliu: This constructor is for converting `GenericRow` to schema-less `Row`, right?

imback82: well, it's just setting the given value to its member (not "convert"ing it)

elvaliuliuliu: Got it, make sense.

Issue #413, opened 2020-02-02
[BUG]: Error using Spark .NET Interactive Notebook Experiences on HDInsight Spark Cluster


fabiogouw:
Hello. I'm trying these steps to enable .NET in Jupyter notebooks in a brand new HDInsight Spark Cluster:
https://github.com/dotnet/spark/tree/master/deployment/HDI-Spark/Notebooks

I'm able to see ".NET Spark (C#)" as an option while creating a new notebook. However, while submitting jobs using Spark .NET (for example, just typing _spark_ and sending this command), I'm receiving this error:

_The code failed because of a fatal error:
	Invalid status code '400' from http://headnodehost:8998/sessions with error payload: "Invalid kind: sparkdotnet (through reference chain: org.apache.livy.server.interactive.CreateInteractiveRequest[\"kind\"])"._

I also had to run these procedures https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/hdinsight/hdinsight-release-notes.md because I was running the "ValueError: Cannot convert notebook to v5 because that version doesn't exist" issue.

Can you help me?

Thanks

elvaliuliuliu:
Thanks for your question! Can you attach the full error and output logs from script action?

fabiogouw:
Sure, here's the information requested:

```
Name
Install Sparkdotnet Interactive Notebook Experience
URI
https://myspark56hdistorage.blob.core.windows.net/scripts/install-interactive-notebook.sh
Parameters
0.6.0
Roles
Head,Worker
Status
Succeeded
Debug information
{
    "href": "http://hn0-myspar.bktuwdxhh4qe1nsazaxnbs2fta.nx.internal.cloudapp.net:8080/api/v1/clusters/myspark56/requests/31",
    "tasks": [
        {
            "href": "http://hn0-myspar.bktuwdxhh4qe1nsazaxnbs2fta.nx.internal.cloudapp.net:8080/api/v1/clusters/myspark56/requests/31/tasks/106",
            "Tasks": {
                "attempt_cnt": 1,
                "command": "ACTIONEXECUTE",
                "command_detail": "run_customscriptaction ACTIONEXECUTE",
                "end_time": 1580696084436,
                "error_log": "/var/lib/ambari-agent/data/errors-106.txt",
                "exit_code": 0,
                "host_name": "hn0-myspar.bktuwdxhh4qe1nsazaxnbs2fta.nx.internal.cloudapp.net",
                "id": "106",
[...]

PR #415, opened 2020-02-05
Fix HDI Notebook SparkDotnet installation script

elvaliuliuliu:
This PR addresses #413. It will fix the issue that new livy jars are not picked up because of version difference.


deployment/HDI-Spark/Notebooks/install-interactive-notebook.sh
```diff
@@ -11,6 +11,10 @@ set -e
 if  [[ $HOSTNAME == hn* ]] ;
 then
     # Update Livy Jars
+    sudo rm -rf /usr/hdp/current/livy2-server/jars/*
```
imback82: can we back these up instead of completely removing them?

elvaliuliuliu: Sure, I have changed the command to back these jars up in folder `old`.

PR #417, opened 2020-02-06
C# coding style regarding the use of var keyword.

imback82:
This PR adds C# coding style regarding the use of `var` keyword because it was not being used consistently. This PR also tries to fix inconsistencies in the current repo as much as possible. The following case is not addressed yet: `Foo foo = new Foo(); // NOT OK`.

### DIFF
--- a/docs/coding-guidelines/csharp-coding-style.md
+++ b/docs/coding-guidelines/csharp-coding-style.md
@@ -8,11 +8,11 @@ We use the same [coding style](https://github.com/dotnet/corefx/blob/master/Docu
     var foo = new Foo(); // OK
     Foo foo = new Foo(); // NOT OK
 
-    var bar = (Bar)foo; // OK
-    Bar bar = (Bar)foo; // NOT OK
-
     var bar = foo as Bar; // OK
     Bar bar = foo as Bar; // NOT OK
+    
+    var bar = (Bar)foo; // OK
+    Bar bar = (Bar)foo; // NOT OK
 
     string str = "hello"; // OK
     var str = "hello"; // NOT OK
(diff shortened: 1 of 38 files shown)


imback82:
cc: @rapoth @suhsteve @Niharikadutta @elvaliuliuliu @GoEddie

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -185,8 +185,7 @@ public void TestUdfWithRowType()
                 Row[] rows = _df.Select(udf(_df["info3"])).Collect().ToArray();
                 Assert.Equal(3, rows.Length);
 
-                var expected = new[] { "Developer", "Developer", "Developer" };
-                string[] actual = rows.Select(x => x[0].ToString()).ToArray();
+                var expected = new string[] { "Developer", "Developer", "Developer" };                string[] actual = rows.Select(x => x[0].ToString()).ToArray();
```
elvaliuliuliu:
```suggestion
                var expected = new string[] { "Developer", "Developer", "Developer" };
                string[] actual = rows.Select(x => x[0].ToString()).ToArray();
```

docs/coding-guidelines/csharp-coding-style.md
```diff
@@ -3,6 +3,30 @@ C# Coding Style
 
 We use the same [coding style](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) and [EditorConfig](https://editorconfig.org "EditorConfig homepage") file (`.editorconfig`) used by [dotnet/corefx](https://github.com/dotnet/corefx) with the following differences:
 
+* **`var` must be used when `new`, `as`, or cast operator is used (and it can be used only with these operators).**
```
elvaliuliuliu: Can we change the order of `as` and cast either here or in the following examples to be consistent?

Issue #410, opened 2020-01-31
[FEATURE REQUEST]: Make the DateTime type available in the create dataframe api


ewatson95:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when trying to pass in an object of type DateTime to the CreateDataFrame API. Most data sets, at least from my experience, have some sort of timestamp associated with the data. However, when I try to pass in a DateTime, DateTimeOffset or TimeSpan data type, I get the error: System.DateTime not supported yet. This seems like an important data type for most applications.

**Describe the solution you'd like**
Implement the System.DateTime data type for the CreateDataFrame API. Or has anyone built an implementation of this that I can use until the feature comes out officially?

imback82:
Thanks @ewatson95 for reporting this. This is in our roadmap, and you should see this in upcoming releases. Also #26 

cc @elvaliuliuliu

ewatson95:
Just to be 100% certain we are talking about the same thing, I'm interested in the SparkSession.CreateDataFrame api. I have seen the DateType data type - will this be sufficient for a c# DateTime object to be passed into the CreateDataFrame method?

imback82:
Yes, it is the current plan to use C# DateTime objects.

elvaliuliuliu:
@ewatson95 , thanks for your question! Could you provide a code snippet of how you are using DateType in your application?

imback82:
@elvaliuliuliu Basically, C# [DateTime](https://docs.microsoft.com/en-us/dotnet/api/system.datetime?view=netcore-3.1) objects should be allowed to be used if the type of your `DataFrame`'s fields is `DateType`. This applies to `UDF`, `Collect`, `CreateDataFrame`, etc.

elvaliuliuliu:
> @elvaliuliuliu Basically, C# [DateTime](https://docs.microsoft.com/en-us/dotnet/api/system.datetime?view=netcore-3.1) objects should be allowed to be used if the type of your `DataFrame`'s fields is `DateType`. This applies to `UDF`, `Collect`, `CreateDataFrame`, etc.

Got it! Thanks for clarification! That's also my understanding. It applies not only to `CreateDataFrame` API, but also all related APIs.

ewatson95:
Yes that's it - so i want to do:

`            var schema = new StructType(new List<StructField>()
            {
                new StructField("Point", new StringType()),
                new StructField("Status", new StringType()),
                new StructField("Value", new DoubleType()),
[...]

Issue #26, opened 2018-12-19
Support more data types: ArrayType, MapType, DateType, and StructType


imback82:
The following types are not fully supported:
ArrayType, MapType, DateType, and StructType.

Check Row.cs, etc.

imback82:
cc: @suhsteve / @elvaliuliuliu

imback82:
@elvaliuliuliu which types are remaining now?

elvaliuliuliu:
> @elvaliuliuliu which types are remaining now?

So according to Row.cs, `ArrayType`, `MapType`, `DecimalType`, `DateType` is not supported at the moment. I am working on `MapType`.

Niharikadutta:
@imback82 I can take up DateTime and also expose that type for CreateDataFrame

PR #420, opened 2020-02-07
Support DateType

### LINKED ISSUE #389: [BUG]: Error while trying to enumerate a DataFrame containing a column of type date: The method or operation is not implemented.
**Describe the bug**
An error occurred while trying to enumerate a DataFrame containing a column of type date.

**To Reproduce**

```
public void Test()
{
  var dataFrame1 = spark.Session.Sql("SELECT * FROM VALUES ('2020-1-1'), ('2020-1-2') AS (DateAsString)");
  dataFrame1.Show();
  dataFrame1.PrintSchema();
  var rows1 = dataFrame1.Collect().Count(); // returns 2

  var dataFrame2 = spark.Session.Sql("SELECT * FROM VALUES CAST('2020-1-1' AS date), CAST('2020-1-2' AS date) AS (DateAsDate)");
  dataFrame2.Show();
  dataFrame2.PrintSchema();
  var rows2 = dataFrame2.Collect().Count(); // failed with "System.NotImplementedException" {"The method or operation is not implemented."}
}
```
Command-Line output:
+------------+
|DateAsString|
+------------+
|    2020-1-1|
|    2020-1-2|
+------------+

root
 |-- DateAsString: string (nullable = false)

+----------+
|DateAsDate|
+----------+
|2020-01-01|
|2020-01-02|
+----------+

root
 |-- DateAsDate: date (nullable = true)

**Exception**
"System.NotImplementedException: The method or operation is not implemented.
   at Microsoft.Spark.Sql.Row.Convert()
   at Microsoft.Spark.Sql.Row..ctor(Object[] values, StructType schema)
   at Microsoft.Spark.Sql.RowConstructor.GetRow()
   at Microsoft.Spark.Sql.RowCollector.Collect(ISocketWrapper socket)+MoveNext()
   at Microsoft.Spark.Sql.DataFrame.GetRows(String funcName)+MoveNext()
   at System.Linq.Enumerable.Count[TSource](IEnumerable`1 source)
   at MyTest.Test() in ..."

**Expected behavior**
A DataFrame containing columns of type date can be enumerated.

**Additional context**
Tested with a latest version (0.7.0) of the .net spark library.

elvaliuliuliu:
This PR exposes `DateType` which will support one of the `DataType` in #26. This PR will also fix #389 and #410. It provides the `DateType` support as below:
1. Support `DataFrame.Collect()` for `DateType`.
2. Support `CreateDataFrame` that takes in `Date`. Please see examples as below.
3. Support UDF that takes in `Date` and UDF that returns `Date`. Please see examples as below.
```csharp
// create df using CreateDataFrame.
var data = new List<GenericRow>();
data.Add(new GenericRow(new object[] { "Alice", new Date(2020, 1, 1) }));
data.Add(new GenericRow(new object[] { "Bob", new Date(2020, 1, 2) }));

var schema = new StructType(new List<StructField>()
{
    new StructField("name", new StringType()),
    new StructField("date", new DateType())
});
DataFrame df = spark.CreateDataFrame(data, schema);

// PrintSchema() prints:
// root
//  |-- name: string (nullable = true)
//  |-- date: date (nullable = true)
df.PrintSchema();

// Show() prints:
// +-----+----------+
// | name|      date|
// +-----+----------+
// |Alice|2020-01-01|
// |  Bob|2020-01-02|
// +-----+----------+
df.Show();

// udf that takes in DateTime.
Func<Column, Column> udf1 = Udf<Date, string>(s => s.ToString());
DataFrame udfDf1 = df.Select(udf1(df["date"]).As("udf1"));
Row[] rows1 = udfDf1.Collect().ToArray();

// PrintSchema() prints:
// root
//  |-- udf1: string (nullable = true)
udfDf1.PrintSchema();

// Show() prints:
// +----------+
// |   udf1   |
// +----------+
// |2020-01-01|
// |2020-01-02|
// +---------+
udfDf1.Show();

//udf that returns Date.
Func<Column, Column> udf2 = Udf<string, Date>(s => new Date(2020, 2, 2));
DataFrame udfDf2 = df.Select(udf2(df["name"]).As("udf2"));
Row[] rows2 = udfDf2.Collect().ToArray();

// PrintSchema() prints:
// root
//  |-- udf2: date (nullable = true)
udfDf2.PrintSchema();

// Show() prints:
// +----------+
// |      udf2|
// +----------+
// |2020-02-02|
// |2020-02-02|
// +----------+
udfDf2.Show();
```

### DIFF
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/FunctionsTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/FunctionsTests.cs
@@ -8,6 +8,7 @@ using System.IO;
 using Microsoft.Spark.E2ETest.Utils;
 using Microsoft.Spark.Sql;
 using Microsoft.Spark.Sql.Catalog;
+using Microsoft.Spark.Sql.Types;
 using Xunit;
 using static Microsoft.Spark.Sql.Functions;
 
@@ -650,6 +651,7 @@ namespace Microsoft.Spark.E2ETest.IpcTests
             Udf<int, int>((arg) => arg);
             Udf<long, long>((arg) => arg);
             Udf<short, short>((arg) => arg);
+            Udf<DateTime, Date>((arg) => new Date(arg));
 
             // Test array type.
             Udf<string, string[]>((arg) => new[] { arg });
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/FunctionsTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/FunctionsTests.cs
@@ -651,7 +651,7 @@ namespace Microsoft.Spark.E2ETest.IpcTests
             Udf<int, int>((arg) => arg);
             Udf<long, long>((arg) => arg);
             Udf<short, short>((arg) => arg);
-            Udf<DateTime, Date>((arg) => new Date(arg));
+            Udf<Date, Date>((arg) => arg);
 
             // Test array type.
             Udf<string, string[]>((arg) => new[] { arg });
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
@@ -127,6 +127,19 @@ namespace Microsoft.Spark.E2ETest.IpcTests
                 DataFrame df = _spark.CreateDataFrame(data);
                 ValidateDataFrame(df, data.Select(a => new object[] { a }), schema);
             }
+            
+            // Calling CreateDataFrame(IEnumerable<DateTime> _) without schema
+            {
+                var data = new List<DateTime>(new DateTime[]
+                {
+                    new DateTime(2020, 1, 1),
+                    new DateTime(2020, 1, 2)
+                });
+                StructType schema = SchemaWithSingleColumn(new DateType());
+
+                DataFrame df = _spark.CreateDataFrame(data);
+                ValidateDataFrame(df, data.Select(a => new object[] { a }), schema);
+            }
         }
 
         private void ValidateDataFrame(
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
@@ -79,8 +79,8 @@ namespace Microsoft.Spark.E2ETest.IpcTests
             // Calling CreateDataFrame with schema
             {
                 var data = new List<GenericRow>();
-                data.Add(new GenericRow(new object[] { "Alice", 20, new DateTime(2020, 1, 1) }));
-                data.Add(new GenericRow(new object[] { "Bob", 30, new DateTime(2020, 1, 2) }));
+                data.Add(new GenericRow(new object[] { "Alice", 20, new Date(2020, 1, 1) }));
+                data.Add(new GenericRow(new object[] { "Bob", 30, new Date(2020, 1, 2) }));
 
                 var schema = new StructType(new List<StructField>()
                 {
@@ -128,12 +128,12 @@ namespace Microsoft.Spark.E2ETest.IpcTests
                 ValidateDataFrame(df, data.Select(a => new object[] { a }), schema);
             }
             
-            // Calling CreateDataFrame(IEnumerable<DateTime> _) without schema
+            // Calling CreateDataFrame(IEnumerable<Date> _) without schema
             {
-                var data = new List<DateTime>(new DateTime[]
+                var data = new List<Date>(new Date[]
                 {
-                    new DateTime(2020, 1, 1),
-                    new DateTime(2020, 1, 2)
+                    new Date(2020, 1, 1),
+                    new Date(2020, 1, 2)
                 });
                 StructType schema = SchemaWithSingleColumn(new DateType());
 
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
@@ -130,11 +130,11 @@ namespace Microsoft.Spark.E2ETest.IpcTests
             
             // Calling CreateDataFrame(IEnumerable<Date> _) without schema
             {
-                var data = new List<Date>(new Date[]
+                var data = new Date[]
                 {
                     new Date(2020, 1, 1),
                     new Date(2020, 1, 2)
-                });
+                };
                 StructType schema = SchemaWithSingleColumn(new DateType());
 
                 DataFrame df = _spark.CreateDataFrame(data);
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Spark.E2ETest.Utils;
@@ -78,13 +79,14 @@ namespace Microsoft.Spark.E2ETest.IpcTests
             // Calling CreateDataFrame with schema
             {
                 var data = new List<GenericRow>();
-                data.Add(new GenericRow(new object[] { "Alice", 20 }));
-                data.Add(new GenericRow(new object[] { "Bob", 30 }));
+                data.Add(new GenericRow(new object[] { "Alice", 20, new DateTime(2020, 1, 1) }));
+                data.Add(new GenericRow(new object[] { "Bob", 30, new DateTime(2020, 1, 2) }));
 
                 var schema = new StructType(new List<StructField>()
                 {
                     new StructField("Name", new StringType()),
-                    new StructField("Age", new IntegerType())
+                    new StructField("Age", new IntegerType()),
+                    new StructField("Date", new DateType())
                 });
                 DataFrame df = _spark.CreateDataFrame(data, schema);
                 ValidateDataFrame(df, data.Select(a => a.Values), schema);
--- a/src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSimpleTypesTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSimpleTypesTests.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Spark.E2ETest.UdfTests
             _spark = fixture.Spark;
             _df = _spark
                 .Read()
-                .Schema("name STRING, date DATE")
+                .Schema("name STRING, age INT, date DATE")
                 .Json(Path.Combine($"{TestEnvironment.ResourceDirectory}people.json"));
         }
 
@@ -38,7 +38,7 @@ namespace Microsoft.Spark.E2ETest.UdfTests
             Row[] rows = _df.Select(udf(_df["date"])).Collect().ToArray();
             Assert.Equal(3, rows.Length);
 
-            var expected = new string[] { "1/1/2020", "1/2/2020", "1/3/2020" };
+            var expected = new string[] { "2020-01-01", "2020-01-02", "2020-01-03" };
             string[] rowsToArray = rows.Select(x => x[0].ToString()).ToArray();
             Assert.Equal(expected, rowsToArray);
         }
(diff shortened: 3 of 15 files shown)


src/csharp/Microsoft.Spark/Interop/Ipc/PayloadHelper.cs
```diff
@@ -102,6 +103,10 @@ internal class PayloadHelper
                         SerDe.Write(destination, (double)arg);
                         break;
 
+                    case TypeCode.DateTime:
+                        SerDe.Write(destination, ((DateTime)arg).ToString("yyyy-MM-dd"));
```
imback82: `FromInternal` is getting an integer as a datetime value. Is it possible to pass the integer as value instead of string or is this required? Do you know which code on the scala is changing this back to sql datetype?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
```diff
@@ -78,13 +79,14 @@ public void TestCreateDataFrame()
...
-                data.Add(new GenericRow(new object[] { "Bob", 30 }));
+                data.Add(new GenericRow(new object[] { "Alice", 20, new DateTime(2020, 1, 1) }));
+                data.Add(new GenericRow(new object[] { "Bob", 30, new DateTime(2020, 1, 2) }));
 
                 var schema = new StructType(new List<StructField>()
                 {
                     new StructField("Name", new StringType()),
-                    new StructField("Age", new IntegerType())
+                    new StructField("Age", new IntegerType()),
+                    new StructField("Date", new DateType())
```
imback82: can we also add a test that just tests `IEnumerable<DateTime>`?

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
```diff
@@ -63,10 +64,18 @@ public sealed class BooleanType : AtomicType
...
+    /// Represents a date type. It represents a valid date in the proleptic Gregorian
+    /// calendar. Valid range is [0001-01-01, 9999-12-31].
     /// </summary>
     public sealed class DateType : AtomicType
     {
+        internal override bool NeedConversion() => true;
+
+        internal override object FromInternal(object obj)
+        {
+            var unixTimeEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
imback82: do we need to worry about timezone here?

src/csharp/Microsoft.Spark/Interop/Ipc/PayloadHelper.cs
elvaliuliuliu: The code [here](https://github.com/dotnet/spark/blob/master/src/scala/microsoft-spark-2.4.x/src/main/scala/org/apache/spark/api/dotnet/SerDe.scala#L84) reads datatype as string.

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
elvaliuliuliu: Nope, I have removed timezone.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
elvaliuliuliu: Sure, I have added the test.

lawrencetvo:
All,

While this is being worked on, are there any work around to create a DataFrame with DateType columns populated with values?

Thanks.

elvaliuliuliu:
> All,
> 
> While this is being worked on, are there any work around to create a DataFrame with DateType columns populated with values?
> 
> Thanks.

Yes, this should fix both #389 and #410.

imback82:
> While this is being worked on, are there any work around to create a DataFrame with DateType columns populated with values?

@lawrencetvo The workaround is to create a string column with date representation, and use `ToDate` function to transform the column to `DateType`. You can check here: https://sparkbyexamples.com/spark/spark-convert-timestamp-to-date/

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/FunctionsTests.cs
```diff
@@ -650,6 +651,7 @@ private void TestUdf()
             Udf<int, int>((arg) => arg);
             Udf<long, long>((arg) => arg);
             Udf<short, short>((arg) => arg);
+            Udf<DateTime, Date>((arg) => new Date(arg));
```
imback82: I thought about this a bit more and chaining will not work if UDF takes `DateTime` instead of `Date`. Can we register a custom unpickler for `datetime.date`? Check here: https://github.com/irmen/Pyrolite/blob/pyro4-legacy/dotnet/Razorvine.Pyrolite/Pyrolite/Pickle/Unpickler.cs#L45

elvaliuliuliu: Resolved as per offline discussion. The new changes is in.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
```diff
@@ -125,6 +127,19 @@ public void TestCreateDataFrame()
                 DataFrame df = _spark.CreateDataFrame(data);
                 ValidateDataFrame(df, data.Select(a => new object[] { a }), schema);
             }
+            
+            // Calling CreateDataFrame(IEnumerable<Date> _) without schema
+            {
+                var data = new List<Date>(new Date[]
```
imback82: I don't think you need `List<Data>`

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -188,6 +188,14 @@ public void Dispose()
         public DataFrame CreateDataFrame(IEnumerable<bool> data) =>
             CreateDataFrame(ToGenericRows(data), SchemaWithSingleColumn(new BooleanType()));
 
+        /// <summary>
+        /// Creates a Dataframe given data as <see cref="IEnumerable"/> of type Date
+        /// </summary>
+        /// <param name="data"></param>
```
imback82: comments missing

src/csharp/Microsoft.Spark/Sql/Types/Date.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        /// <summary>
+        /// Constructor for Date class.
+        /// </summary>
+        /// <param name="year">The year (1 through 9999)</param>
+        /// <param name="month">The month (1 through 12)</param>
+        /// <param name="day">The day (1 through the number of days in month)</param>
+        public Date(int year, int month, int day)
+        {
+            Year = year;
```
imback82: Can you check `DateTime` implementation if they check against the range?

src/csharp/Microsoft.Spark/Sql/Types/Date.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <summary>
+        /// Readable string representation for this type.
+        /// </summary>
+        public override string ToString() => $"{Month}/{Day}/{Year}";
+
+        /// <summary>
+        /// Readable string representation for this type using the specified format.
+        /// </summary>
+        /// <param name="format">A standard or custom date and time format string</param>
+        public string ToString(string format) => new DateTime(Year, Month, Day).ToString(format);
```
imback82: Can we not expose this? What's the rationale for this public API?

src/csharp/Microsoft.Spark/Sql/Types/Date.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        /// <summary>
+        /// Checks if the given object is same as the current object.
+        /// </summary>
+        /// <param name="obj">Other object to compare against</param>
+        /// <returns>True if the other object is equal.</returns>
+        public override bool Equals(object obj) =>
+            ReferenceEquals(this, obj) ||
+            ((obj is Date date) && Year.Equals(date.Year) &&
```
imback82:
```C#
ReferenceEquals(this, obj) ||
((obj is Date date) && Year.Equals(date.Year) && Month.Equals(date.Month) &&
    Day.Equals(date.Day));
```

src/csharp/Microsoft.Spark/Sql/Types/Date.cs
```diff
@@ -0,0 +1,94 @@
...
+            Month.Equals(date.Month) && Day.Equals(date.Day));
+
+        /// <summary>
+        /// Returns the hash code of the current object.
+        /// </summary>
+        /// <returns>The hash code of the current object</returns>
+        public override int GetHashCode() => base.GetHashCode();
+
+        /// <summary>
+        /// Returns DateTime object describing this type.
```
imback82: Can we have <returns> tag since this is public API?

src/csharp/Microsoft.Spark/Sql/Types/Date.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        public DateTime ToDateTime() => new DateTime(Year, Month, Day);
+
+        /// <summary>
+        /// Returns an integer object that represents a count of days from 1970-01-01.
+        /// </summary>
+        internal int GetInterval()
+        {
+            var unixTimeEpoch = new DateTime(1970, 1, 1);
```
imback82: can we create a constant for (new DateTime(1970, 1, 1)).Days so that we don't create this object every single time?

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
```diff
@@ -63,10 +64,18 @@ public sealed class BooleanType : AtomicType
...
+    /// Represents a date type. It represents a valid date in the proleptic Gregorian
+    /// calendar. Valid range is [0001-01-01, 9999-12-31].
     /// </summary>
     public sealed class DateType : AtomicType
     {
+        internal override bool NeedConversion() => true;
+
+        internal override object FromInternal(object obj)
+        {
+            var unixTimeEpoch = new DateTime(1970, 1, 1);
```
imback82: ditto

src/csharp/Microsoft.Spark/Sql/Types/Date.cs
imback82: https://referencesource.microsoft.com/#mscorlib/system/datetime.cs

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSimpleTypesTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            Assert.Equal(expected, rowsToArray);
+        }
+
+        /// <summary>
+        /// UDF that returns Date type.
+        /// </summary>
+        [Fact]
+        public void TestUdfWithReturnAsDateType()
+        {
+            Func<Column, Column> udf1 = Udf<string, Date>(str => new Date(2020, 1, 4));
```
imback82: can we vary the values returned?

src/csharp/Microsoft.Spark/Sql/Types/Date.cs
elvaliuliuliu: Yes, it will check the range.

src/csharp/Microsoft.Spark/Sql/Types/Date.cs
elvaliuliuliu: This has been removed. It's not needed.

src/csharp/Microsoft.Spark/Sql/Types/Date.cs
elvaliuliuliu: Changed.

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
elvaliuliuliu: Changed.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
elvaliuliuliu: Removed

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
elvaliuliuliu: Added the comments.

src/csharp/Microsoft.Spark/Sql/Types/Date.cs
elvaliuliuliu: Tag is added. Thanks.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSimpleTypesTests.cs
elvaliuliuliu: Sure, I have modified the test.

src/csharp/Microsoft.Spark/Sql/Types/Date.cs
```diff
@@ -0,0 +1,85 @@
...
+    public class Date
+    {
+        private readonly DateTime _dateTime;
+        private static readonly DateTime s_unixTimeEpoch = new DateTime(1970, 1, 1);
+
+        /// <summary>
+        /// Constructor for Date class.
+        /// </summary>
+        /// <param name="dt">DateTime object</param>
+        public Date(DateTime dt)
```
imback82: can we rename this to `dateTime` since we already have `_dateTime`?

elvaliuliuliu: Updated.

PR #424, opened 2020-02-11
Fixes #421: Orphaning of DotnetBackend and .NET process in Databricks

### LINKED ISSUE #421: [BUG]: Orphaned DotnetBackend thread and .NET process when job cancelled in Databricks
**Describe the bug**

If a Spark.NET job is run in Databricks and that run is cancelled through Databricks, the .NET process will continue to run after Databricks has considered the job cancelled. The happens with the DotnetBackend thread.

This means that the job will continue to run normally, but Databricks will lose track of its existence.

**To Reproduce**

1. Create a simple .NET application with an infinite loop
```csharp
public static void Main(string[] args)
{
    while (true)
    {
        Console.WriteLine($"[{DateTime.Now.ToString("o")}] Spark.NET job still running.");
        Task.Delay(TimeSpan.FromSeconds(5)).GetAwaiter().GetResult();
    }
}
```
2. Submit job to an existing Databricks cluster configured for Spark.NET
3. Check the Databricks cluster's stdout log to verify it's running
4. Navigate to the Databricks job that you submitted and cancel the run
5. Wait for cancellation to finish
6. Check the cluster's stdout log again. Refreshing will show new messages forever.

This doesn't illustrate DotnetBackend continuing to run, but a normal Spark job would continue to succeed. I also have confirmation of the thread being orphaned below.

**Expected behavior**

Cancelling the Databricks job also kills the .NET process and the DotnetBackend thread.

**Screenshots**
image
image

 - OS: Windows 10
 - Browser: Edge Chromium
 - Version: 79.0.309.71 (Official build) (64-bit)

**Additional context**

The orphaning of both the .NET process and the DotnetBackend thread can be confirmed with the help of a Databricks Notebook attached to the respective cluster.

**To confirm the process is running after cancellation (if not checking console output), replacing *placeholder* with a partial or full name of the running .NET assembly:**

```sh
%sh ps aux | grep 'placeholder'
```

**To confirm the DotnetBackend thread is running, replacing *pid* with the process ID of the parent java process:**

The appropriate java pid can be found with **%sh pstree -p** and finding the nearest parent java process of the *pstree* command.

```sh
%sh jstack -l pid
```

The output from *jstack* will be the running threads in the process. If you search the output text for **dotnet**, you'll see **DotnetBackend** is still running.
You can contrast it if you run *jstack* while the job is running, where you'll find additionally find the running **DotnetRunner** waiting on the call to **UNIXProcess.waitFor()**.

Cancelling the job will cancel the execution of **DotnetRunner** by throwing an exception. **logError** will be called in the catch, and the **SparkUserAppException** will be thrown just after, orphaning the thread and process.

Here are some attached outputs from jstack. I started on a fresh cluster for comparison.
[1.before-run.txt](https://github.com/dotnet/spark/files/4173445/1.before-run.txt)
[2.while-running.txt](https://github.com/dotnet/spark/files/4173446/2.while-running.txt)
[3.after-cancel.txt](https://github.com/dotnet/spark/files/4173447/3.after-cancel.txt)

tylorhl:
Addresses #421 and additionally fixes `logError` to actually print the stack trace from the `Throwable` (previously it printed the object string).

### Changes
- Added `closeDotnetProcess` for cleaning up the started .NET process
- Added `finally` for cleaning up `DotnetBackend` as well as the .NET process
- Added `logThrowable` for code reuse

### Tested scenarios
- `process.destroy()` vs `process.destroyForcibly()`
  - `SIGTERM` sent vs `SIGKILL` sent respectively
- Confirming request to terminate is received inside of .NET when `process.destroy()` is called
  - Can be caught via `AppDomain.CurrentDomain.ProcessExit`
- Determined what happens when .NET tries to continue running indefinitely after termination signal
  - Another thread interrupt is made by Databricks after ~6 seconds
  - I did not test what happens after the second interrupt
- Concluded that Windows doesn't have a reasonable graceful termination story
  - Windows doesn't have termination signals like Linux
  - OpenJDK8 calls [TerminateProcess](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminateprocess) to destroy a process on Windows
- Validated that `process.destroyForcibly()` does in fact kill the process ungracefully
- Validated that the `DotnetBackend` thread is getting closed inside of the the Java process

#### Cleanup logic
Databricks sends an interrupt signal to the DotnetRunner's thread, which causes `process.waitFor()` to throw an `InterruptedException`. When cleanup occurs after the initial interrupt, Databricks may send a second interrupt if the DotnetRunner thread hasn't stopped within about 6 seconds (based on log timestamps). The idea here is that if a second interrupt occurs, then we should forcibly kill the .NET process via `process.destroyForcibly()`.

On Windows, if this same logic occurred, the .NET process would be forcibly killed once `process.destroy()` was called. Through minor testing, it doesn't seem like we can manually send an equivalent signal via the stdin of the process.

#### stdin, stdout, stderr notes
It should be noted that once `process.destroy()` is called, the standard streams of the process are closed. Printing to the console from .NET will stop working once called.

#### Throwable logging fix
I noticed that the stack trace from `catch` was printing the stack trace array's object string instead of the actual stack trace, so I added `.mkString("\n")` to the call.

### DIFF
--- a/src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
+++ b/src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
@@ -132,9 +132,9 @@ object DotnetRunner extends Logging {
           process.waitFor()
         } catch {
           case t: Throwable =>
-            logError(s"${t.getMessage} \n ${t.getStackTrace}")
+            logThrowable(t)
         } finally {
-          returnCode = destroyDotnetProcess(process)
+          returnCode = closeDotnetProcess(process)
           closeBackend(dotnetBackend)
         }
 
@@ -235,12 +235,28 @@ object DotnetRunner extends Logging {
     dotnetBackend.close()
   }
 
-  private def destroyDotnetProcess(dotnetProcess: Process): Int = dotnetProcess match {
-    case process: Process if process.isAlive =>
-      logInfo("Destroying .NET Process")
-      process.destroyForcibly().waitFor()
-    case process: Process => process.exitValue()
-    case _ => 0
+  private def closeDotnetProcess(dotnetProcess: Process): Int = {
+    if (dotnetProcess == null) {
+      return -1
+    } else if (!dotnetProcess.isAlive) {
+      return dotnetProcess.exitValue()
+    }
+
+    // Try to (gracefully on Linux) kill the process and resort to force if interrupted
+    var returnCode = -1
+    logInfo("Closing .NET process")
+    try {
+      dotnetProcess.destroy()
+      returnCode = dotnetProcess.waitFor()
+    } catch {
+      case _: InterruptedException =>
+        logInfo("Thread interrupted while waiting for graceful close. Forcefully closing .NET process")
+        returnCode = dotnetProcess.destroyForcibly().waitFor()
+      case t: Throwable =>
+        logThrowable(t)
+    }
+
+    returnCode
   }
 
   private def initializeSettings(args: Array[String]): (Boolean, Int) = {
@@ -257,4 +273,7 @@ object DotnetRunner extends Logging {
 
     (runInDebugMode, portNumber)
   }
+
+  private def logThrowable(throwable: Throwable): Unit =
+    logError(s"${throwable.getMessage} \n ${throwable.getStackTrace.mkString("\n")}")
 }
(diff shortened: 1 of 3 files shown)


src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
```diff
@@ -232,6 +235,30 @@ object DotnetRunner extends Logging {
...
+      return -1
+    } else if (!dotnetProcess.isAlive) {
+      return dotnetProcess.exitValue()
+    }
+
+    // Try to (gracefully on Linux) kill the process and resort to force if interrupted
+    var returnCode = -1
+    logInfo("Closing .NET process")
+    try {
+      dotnetProcess.destroy()
```
imback82: Do we want to call `destroy` before `waitFor` even for the normal exit case? What will happen without this?

imback82: ah nevermind, this is the second attempt.

PR #428, opened 2020-02-22
Support TimestampType

elvaliuliuliu:
This PR exposes `Timestamp` class and supports `TimestampType` as below:
1. Support `DataFrame.Collect()` for `TimestampType`.
2. Support `CreateDataFrame` that takes in `Timestamp`.
3. Support UDF that takes in `Timestamp` and UDF that returns `Timestamp`.


imback82:
@elvaliuliuliu Is this ready for review?

elvaliuliuliu:
> @elvaliuliuliu Is this ready for review?

Yes, this is ready for review! Looks like request reviewers function is removed from github, so I will ping you next time when wip pr is ready for review. Thanks!

src/csharp/Microsoft.Spark/Sql/CustomPicklers.cs
```diff
@@ -40,4 +40,15 @@ public void pickle(object o, Stream outs, Pickler currentPickler)
             currentPickler.save(((Date)o).GetInterval());
         }
     }
+
+    /// <summary>
+    /// Custom pickler for Timestamp objects.
+    /// </summary>
+    internal class TimestampPickler : IObjectPickler
+    {
+        public void pickle(object o, Stream outs, Pickler currentPickler)
```
Niharikadutta: Do we want to rename `outs` to either `outStream` or just `out`? `outs` reads a little funny to me.

elvaliuliuliu: Looking through some other examples, `outs` has been used to stand for `output stream`.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSimpleTypesTests.cs
```diff
@@ -21,10 +22,28 @@ public class UdfSimpleTypesTests
...
+                    null,
+                    new Date(2020, 1, 1),
+                    new Timestamp(2020, 1, 1, 0, 0, 0, 0)
+                }));
+            data.Add(new GenericRow(
+                new object[]
+                {
+                    30,
+                    new Date(2020, 1, 2),
+                    new Timestamp(2020, 1, 2, 15, 30, 30, 0)
```
imback82: Can we use non zero value for microseconds?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSimpleTypesTests.cs
```diff
@@ -74,9 +92,74 @@ public void TestUdfWithReturnAsDateType()
...
+        {
+            Func<Column, Column> udf = Udf<Timestamp, string>(time => time.ToString());
+
+            Row[] rows = _df.Select(udf(_df["time"])).Collect().ToArray();
+            Assert.Equal(2, rows.Length);
+
+            var expected = new string[]
+            {
+                "2020-01-01 00:00:00.000000",
+                "2020-01-02 15:30:30.000000"
```
imback82: I assume this is the same format that Scala/Python uses for toString?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSimpleTypesTests.cs
```diff
@@ -74,9 +92,74 @@ public void TestUdfWithReturnAsDateType()
...
+        }
+
+        /// <summary>
+        /// UDF that returns Timestamp type.
+        /// </summary>
+        [Fact]
+        public void TestUdfWithReturnAsTimestampType()
+        {
+            Func<Column, Column> udf1 = Udf<int?, Timestamp>(
+                s => new Timestamp(2020 + s.GetValueOrDefault(), 1, 4, 0, 0, 0, 0));
```
imback82: nit: Looks like this is carried from the existing code, but let's either use `x` or `i` (for integer) instead of `s`. (or even `age` is fine).

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -196,6 +196,15 @@ public void Dispose()
         public DataFrame CreateDataFrame(IEnumerable<Date> data) =>
             CreateDataFrame(ToGenericRows(data), SchemaWithSingleColumn(new DateType()));
 
+        /// <summary>
+        /// Creates a Dataframe given data as <see cref="IEnumerable"/> of type
+        /// <see cref="Timestamp"/>
+        /// </summary>
+        /// <param name="data"></param>
```
imback82: let's fill in the comment.

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
```diff
@@ -80,10 +80,22 @@ internal override object FromInternal(object obj)
...
     {
+        private static readonly DateTime s_unixTimeEpoch =
+            new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+        internal override bool NeedConversion() => true;
+
+        internal override object FromInternal(object obj)
+        {
+            return new Timestamp(new DateTime((long)obj * 10 + s_unixTimeEpoch.Ticks,
+                DateTimeKind.Utc));
```
imback82: Can you add a short description of what value `obj`  holds as `long` and what this conversion does?

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,143 @@
...
+{
+    /// <summary>
+    /// Represents Timestamp containing year, month, day, hour, minute, second, microsecond in
+    /// Coordinated Universal Time (UTC).
+    /// </summary>
+    public class Timestamp
+    {
+        private readonly DateTime _dateTime;
+        private static readonly DateTime s_unixTimeEpoch =
+            new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
imback82: Maybe, we should put this in a common place?

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        /// <summary>
+        /// Constructor for Timestamp class.
+        /// </summary>
+        /// <param name="dateTime">DateTime object</param>
+        public Timestamp(DateTime dateTime)
+        {
+            _dateTime = dateTime.ToUniversalTime();
+
+            if (dateTime.Kind != DateTimeKind.Utc)
```
imback82: You can check this before the above line. Can you also add a test for this?

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                throw new InvalidTimeZoneException(
+                    $"Invalid TimeZone for Timestamp, please use " +
```
imback82: Can you also print out the timezone set in `dateTime`?

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// Coordinated Universal Time (UTC).
+        /// </summary>
+        /// <param name="year">The year (1 through 9999)</param>
+        /// <param name="month">The month (1 through 12)</param>
+        /// <param name="day">The day (1 through the number of days in month)</param>
+        /// <param name="hour">The hour (0 through 23)</param>
+        /// <param name="minute">The minute (0 through 59)</param>
+        /// <param name="second">The second (0 through 59)</param>
+        /// <param name="microsecond">The microsecond (0 through 999999)</param>
+        public Timestamp(int year, int month, int day, int hour, int minute, int second, int microsecond)
```
imback82: over 100 limit?

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            if (microsecond <= 999999 && microsecond >= 0)
+            {
+                _dateTime = new DateTime(year, month, day, hour, minute, second)
+                    .AddTicks(microsecond * 10).ToUniversalTime();
```
imback82: Shouldn't the given year, month, etc. already be in UTC? Why do we need to call `ToUniversalTime`? I think you need to pass `DateTimeKind` when you constructor the `DateTime` object.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,143 @@
...
+            }
+            else
+            {
+                throw new ArgumentOutOfRangeException($"Invalid microsecond value {microsecond}");
```
imback82: Shall we put the valid range in the exception message as well?

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        /// <summary>
+        /// Returns the second component of the timestamp.
+        /// </summary>
+        public int Second => _dateTime.Second;
+
+        /// <summary>
+        /// Returns the microsecond component of the timestamp.
+        /// </summary>
+        public int Microsecond => (int)(_dateTime.Ticks -
```
imback82: Can we test this as well? Let's add a test case for `Timestamp` class.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,143 @@
...
+            new DateTime(Year, Month, Day, Hour, Minute, Second).Ticks) / 10;
+
+        /// <summary>
+        /// Readable string representation for this type.
+        /// </summary>
+        public override string ToString() => _dateTime.ToUniversalTime()
```
imback82: We should assume `_dateTime` should be already in UTC so that we don't need to call `ToUniversalTime`.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// Checks if the given object is same as the current object.
+        /// </summary>
+        /// <param name="obj">Other object to compare against</param>
+        /// <returns>True if the other object is equal</returns>
+        public override bool Equals(object obj) =>
+            ReferenceEquals(this, obj) ||
+            ((obj is Timestamp timestamp) && Year.Equals(timestamp.Year) &&
+                Month.Equals(timestamp.Month) && Day.Equals(timestamp.Day) &&
+                Hour.Equals(timestamp.Hour) && Minute.Equals(timestamp.Minute) &&
+                Second.Equals(timestamp.Second) && Microsecond.Equals(timestamp.Microsecond));
```
imback82: Just compare `_dateTime`?

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        /// <summary>
+        /// Returns the hash code of the current object.
+        /// </summary>
+        /// <returns>The hash code of the current object</returns>
+        public override int GetHashCode() => base.GetHashCode();
```
imback82: Let's return a hash code of `_dateTime`.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        /// <summary>
+        /// Returns DateTime object describing this type.
+        /// </summary>
+        /// <returns>DateTime object of the current object</returns>
+        public DateTime ToDateTime() => _dateTime;
+
+        /// <summary>
+        /// Returns a double object that represents the number of seconds from the epoch of
```
imback82: Can you convey the intention that the unit is seconds but we preserve the sub-seconds as fraction?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSimpleTypesTests.cs
elvaliuliuliu: Yes, same format.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSimpleTypesTests.cs
elvaliuliuliu: Got it, good catch! So actually non zero microseconds doesn't work for udf, it always converts to zero microseconds when reading input data [here](https://github.com/dotnet/spark/blob/9275678a6e46e06a266e5516b3fea50896ede761/src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs#L135).

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfSimpleTypesTests.cs
elvaliuliuliu: Sure, changed for both here and previous code from `Date`!

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
elvaliuliuliu: Sure, added!

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
elvaliuliuliu: Sure, added.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: This is also used in [DateTimeExtension.cs](https://github.com/dotnet/spark/blob/9275678a6e46e06a266e5516b3fea50896ede761/src/csharp/Microsoft.Spark.Worker/Utils/DateTimeExtension.cs#L13), do you mean this?

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: Added.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: Done.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: So pass `DateTimeKind` does not give the right time, still have 8 hours lack in this case so I have to convert it using `ToUniversalTime`.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: Sounds good! Added.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: This is the same as what happened above regarding passing kind is not working.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: Changed.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: Sure, I have changed the summary and let me know if this makes sense.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: Should I add a unit test for `Timestamp` class?

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: Added this in Unit test.

src/csharp/Microsoft.Spark.UnitTest/Sql/TimestampTests.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace Microsoft.Spark.UnitTest
+{
+    public class TimestampTests
+    {
+        [Fact]
+        public void TimestampTest()
+        {
+            {
+                var timestamp = new Timestamp(
+                    new DateTime(2020, 1, 2, 15, 30, 30, DateTimeKind.Utc));
```
imback82: Can you add a test where `DataTime` has non zero microsecond (using `Ticks`)?

src/csharp/Microsoft.Spark.UnitTest/Sql/TimestampTests.cs
```diff
@@ -0,0 +1,45 @@
...
+                Assert.Equal(15, timestamp.Hour);
+                Assert.Equal(30, timestamp.Minute);
+                Assert.Equal(30, timestamp.Second);
+                Assert.Equal(0, timestamp.Microsecond);
+
+                // Validate ToString().
+                Assert.Equal("2020-01-02 15:30:30.000000", timestamp.ToString());
+
+                // Validate ToDateTime().
+                Assert.IsType<DateTime>(timestamp.ToDateTime());
```
imback82: we need to verify the returned value is same is the one you started with.

imback82: I think we also need to verify that even if you pass in non-UTC `DateTime`, you get the UTC one.

src/csharp/Microsoft.Spark.UnitTest/Sql/TimestampTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            }
+
+            {
+                // Validate Microsecond and TimeZone.
+                var timestamp = new Timestamp(2020, 1, 2, 15, 30, 30, 123456);
+
+                Assert.Equal(123456, timestamp.Microsecond);
```
imback82: Can we just validate all fields? How about the invalid microsecond case?

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
```diff
@@ -69,21 +69,52 @@ public sealed class BooleanType : AtomicType
     /// </summary>
     public sealed class DateType : AtomicType
     {
-        private static readonly DateTime s_unixTimeEpoch = new DateTime(1970, 1, 1);
+        internal static readonly DateTime s_unixTimeEpoch = new DateTime(1970, 1, 1);
```
imback82: I think we can just set this to `new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);` and use from `TimeStamp` as well.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,140 @@
...
+            if (microsecond <= 999999 && microsecond >= 0)
+            {
+                // Create DateTime and AddTicks based on the microsecond value.
+                _dateTime = new DateTime(year, month, day, hour, minute, second, DateTimeKind.Utc)
+                    .AddTicks(microsecond * 10);
+            }
+            else
+            {
+                throw new ArgumentOutOfRangeException($"Invalid microsecond value {microsecond}. " +
+                    $"The microsecond is from 0 through 999999.");
```
imback82: nit: `The microsecond should be in the range of [0, 999999]`?

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,140 @@
...
+        public Timestamp(
+            int year,
+            int month,
+            int day,
+            int hour,
+            int minute,
+            int second,
+            int microsecond)
+        {
+            if (microsecond <= 999999 && microsecond >= 0)
```
imback82: nit: `if ((0 <= microsecond) && (microsecond <= 999999))`

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        /// <summary>
+        /// Returns the second component of the timestamp.
+        /// </summary>
+        public int Second => _dateTime.Second;
+
+        /// <summary>
+        /// Returns the microsecond component of the timestamp.
+        /// </summary>
+        public int Microsecond => (int)(_dateTime.Ticks -
```
imback82: Can't you just do `(int)(_dateTime.Ticks % 10000000 / 10)`?

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
elvaliuliuliu: Sure, I have changed all to use `DateType.s_unixTimeEpoch`.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: Changed.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: Changed!

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
elvaliuliuliu: If we use `(int)(_dateTime.Ticks % 10000000 / 10)`, it will be the whole value represented in microseconds instead of just microsecond value, no?

imback82: Since we are modding, it should just get the microsecond value.

elvaliuliuliu: Yep, right! Thanks.

src/csharp/Microsoft.Spark.UnitTest/Sql/TimestampTests.cs
elvaliuliuliu: Added!

src/csharp/Microsoft.Spark.UnitTest/Sql/TimestampTests.cs
elvaliuliuliu: Added the tests, thanks!

src/csharp/Microsoft.Spark.UnitTest/Sql/TimestampTests.cs
elvaliuliuliu: Yes, added for both!

src/csharp/Microsoft.Spark.UnitTest/Sql/TimestampTests.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace Microsoft.Spark.UnitTest
+{
+    public class TimestampTests
+    {
+        [Fact]
+        public void TimestampTest()
+        {
+            {
+                var timestamp = new Timestamp(
+                    new DateTime(2020, 1, 1, 8, 30, 30, DateTimeKind.Utc).AddTicks(1230));
```
imback82: Shall we create `var testDate = new DateTime(2020, 1, 1, 8, 30, 30, DateTimeKind.Utc).AddTicks(1230);` and reuse it below?

elvaliuliuliu: Sounds good. Changed!

PR #432, opened 2020-02-26
Adding SPARK_VERSION_OVERRIDE environment variable

elvaliuliuliu:
This PR adds the `SPARK_VERSION_OVERRIDE` environment variable.

If `SPARK_VERSION_OVERRIDE` is set, it will override the `DOTNET_WORKER_SPARK_VERSION`. For example, if the driver uses Spark version 2.4.1. And `SPARK_VERSION_OVERRIDE` is set to 2.3.1. When running UDF, the job will hang since the version is not matching.

One the other hand, if we want to run Microsoft.Spark.Worker in specific Spark version. The `SPARK_VERSION_OVERRIDE` should be set to match that version. So it can help avoid (patch)version mismatch problems.

### DIFF
--- a/src/csharp/Microsoft.Spark/Interop/SparkEnvironment.cs
+++ b/src/csharp/Microsoft.Spark/Interop/SparkEnvironment.cs
@@ -14,10 +14,30 @@ namespace Microsoft.Spark.Interop
     /// </summary>
     internal static class SparkEnvironment
     {
-        private static readonly Lazy<Version> s_sparkVersion = new Lazy<Version>(
-            () => new Version((string)JvmBridge.CallStaticJavaMethod(
+        private static readonly ILoggerService s_logger =
+            LoggerServiceFactory.GetLogger(typeof(SparkEnvironment));
+
+        private static Version GetSparkVersion()
+        {
+            var sparkVersion = new Version((string)JvmBridge.CallStaticJavaMethod(
                 "org.apache.spark.deploy.dotnet.DotnetRunner",
-                "SPARK_VERSION")));
+                "SPARK_VERSION"));
+
+            string sparkVersionOverride =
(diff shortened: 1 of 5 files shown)


elvaliuliuliu:
> LGTM if tests pass

Tests has all passed.

src/csharp/Microsoft.Spark/Interop/SparkEnvironment.cs
```diff
@@ -14,10 +14,30 @@ namespace Microsoft.Spark.Interop
...
+            if (!string.IsNullOrEmpty(sparkVersionOverride))
+            {
+                s_logger.LogInfo(
+                    $"Overriding the Spark version from '{sparkVersion}' " +
+                    $"to '{sparkVersionOverride}'.");
+                sparkVersion = new Version(sparkVersionOverride);
+            }
+
+            return sparkVersion;
+        }
```
suhsteve: If the `SPARK_VERSION_OVERRIDE` is defined then return that value, otherwise call and return the value of `org.apache.spark.deploy.dotnet.DotnetRunner.SPARK_VERSION`

imback82: I prefer to print the original version if the version is overridden for debugging.

imback82: I prefer to print the original version if the version is overridden for debugging.

suhsteve: Sure that's fair.

elvaliuliuliu: Sounds good.

PR #391, opened 2020-01-13
Quick doc on how to add a new type to SerDe

GoEddie:
Doc on how to add a new type to SerDe, hope it helps someone else

PR #446, opened 2020-03-05
doc to describe how to add additional types to the SerDe process

GoEddie:
Document to describe how to add a new data type to the SerDe process (was previously https://github.com/dotnet/spark/pull/391)


imback82:
Thanks @GoEddie. There was a bug introduced yesterday in GitHub, which caused this issue. I will merge this once the fix is deployed. Thanks!

PR #451, opened 2020-03-20
Clean up vector UDF examples.

eerhardt:
Clean the Vector Udf examples so they are more "apples to apples" comparison.

/cc @bamurtaugh


imback82:
Btw, when is 0.3 coming out?

pgovind:
> Btw, when is 0.3 coming out?

Today :). I'm bumping up the version now and making a Spark PR locally

eerhardt:
>> Btw, when is 0.3 coming out?
> Today :). I'm bumping up the version now and making a Spark PR locally

We will ensure 0.3 doesn't break Spark before we officially ship 0.3. Because if we need to make a change, we would have to bump the version number.

imback82:
Cool thanks!

Issue #455, opened 2020-03-22
[FEATURE REQUEST]: Allow UDF that returns Row to be registered using UDFRegistration


Velociraptor115:
The implementation of UDF that can return custom Row object as requested in #117 has been done.
However, the APIs for UDF registration haven't been updated to support custom return type

We need to add similar set of methods in UDFRegistration.cs that takes a custom return type as a parameter

imback82:
Fixed by #456.

PR #456, opened 2020-03-22
add returnType to UDFRegistration methods

Velociraptor115:
This PR adds new methods to support schema for Row return type in UDF registration. This implements the feature request of #455 

The current UDF registration API does not accept a schema parameter. This prevents us from registering an UDF like
```
{
    new StructField("col1", new IntegerType())
});
    str => new GenericRow(new object[] { 111 }), schema);

// Though I can get `udf` as a Func<Column, Column>, I can't register it,
// since the spark.Udf().Register* functions do not accept the `returnType` param
```
even though the underlying API supports it.

This PR adds new methods for UDF registration with the returnType param for functions that return Row.
This allows us to register such UDFs like
```
{
    new StructField("col1", new IntegerType())
});
_spark.Udf().Register<string>(
    "my_udf",
    str => new GenericRow(new object[] { 111 }),
    schema);
```


elvaliuliuliu:
@Velociraptor115 : Thanks for your PR and contribution! Can you please fix the failure and also add a test for this feature?

Velociraptor115:
> 
> 
> @Velociraptor115 : Thanks for your PR and contribution! Can you please fix the failure and also add a test for this feature?

Thanks for your quick response.
Would be glad to add tests.

Can you point me to the best place to add them.
I did a quick check for the existing UDF registration methods and couldn't find any places where they were referred, except a couple of places in the examples

imback82:
Thanks @Velociraptor115 for your first PR! Can you update the description with an example that is not working that this PR proposes to fix. And you can convert that example to tests.

Velociraptor115:
I have added tests in `UdfComplexTypesTests.cs`

I did not have a specific example in mind to use for the tests.
My current use case involves registering an UDF in .NET (as I need to use a .NET library)
and invoking that from Scala (since the primary codebase is in Scala)

I have however reused the same tests from `TestUdfWithReturnAsRowType`, rewriting the UDFs to be registered and called from SQL instead.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -320,5 +320,165 @@ public void TestUdfWithReturnAsRowType()
...
+                    "udf5",
+                    row => row.GetAs<string>(1));
+
+                Row[] rows =
+                    _spark.Sql("SELECT udf5(udf4(name)) FROM people")
+                    .Collect()
+                    .ToArray();
+                Assert.Equal(3, rows.Length);
+
+                var expected = new[] { "Michael", "Andy", "Justin" };
```
imback82: nit: Probably existing code, but it should be `var expected = new string[] { ... }`;

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -320,5 +320,165 @@ public void TestUdfWithReturnAsRowType()
...
+
+                _spark.Udf().Register<string>(
+                    "udf1",
+                    str => new GenericRow(new object[] { 1, "abc" }),
+                    schema);
+
+                Row[] rows =
+                    _spark.Sql("SELECT udf1(name) AS col FROM people")
+                    .Collect()
+                    .ToArray();
```
imback82: Either
```C#
Row[] rows =
    _spark.Sql("SELECT udf1(name) AS col FROM people")
        .Collect()
        .ToArray();
```
or
```C#
Row[] rows =_spark.Sql("SELECT udf1(name) AS col FROM people")
    .Collect()
    .ToArray();
```

src/csharp/Microsoft.Spark/Sql/UDFRegistration.cs
```diff
@@ -241,17 +444,28 @@ public void RegisterJavaUDAF(string name, string className)
...
         }
 
         /// <summary>
         /// Helper function to register wrapped udf.
         /// </summary>
-        /// <typeparam name="TResult">Return type of the udf</typeparam>
         /// <param name="name">Name of the udf</param>
         /// <param name="func">Wrapped UDF function</param>
         /// <param name="evalType">The EvalType of the function.</param>
-        internal void Register<TResult>(string name, Delegate func, UdfUtils.PythonEvalType evalType)
```
imback82: Can we leave this API and just call the new one, so that we don't have to change existing calls?
```C#
internal void Register<TResult>(string name, Delegate func, UdfUtils.PythonEvalType evalType)
{
    return Register(name, func, evalType, UdfUtils.GetReturnType(typeof(TResult));
}
```

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
Velociraptor115: I have made these changes in the latest commits

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
Velociraptor115: I have made these changes in the latest commits (used 1st style)

src/csharp/Microsoft.Spark/Sql/UDFRegistration.cs
Velociraptor115: I left the previous internal method as it is, but changed the body to delegate to a newer method which accepts the returnType param.
I have also reverted the changes in the two extension files which called the earlier API
Can you please check and verify if this is ok?

src/csharp/Microsoft.Spark/Sql/UDFRegistration.cs
```diff
@@ -241,7 +444,18 @@ public void RegisterJavaUDAF(string name, string className)
         /// <param name="func">Wrapped UDF function</param>
         private void Register<TResult>(string name, Delegate func)
         {
-            Register<TResult>(name, func, UdfUtils.PythonEvalType.SQL_BATCHED_UDF);
+            Register(name, func, UdfUtils.PythonEvalType.SQL_BATCHED_UDF, UdfUtils.GetReturnType(typeof(TResult)));
```
imback82: Can you revert this change?

src/csharp/Microsoft.Spark/Sql/UDFRegistration.cs
```diff
@@ -252,6 +466,18 @@ private void Register<TResult>(string name, Delegate func)
...
+            Register(name, func, evalType, UdfUtils.GetReturnType(typeof(TResult)));
+        }
+
+        /// <summary>
+        /// Helper function to register wrapped udf.
+        /// </summary>
+        /// <param name="name">Name of the udf</param>
+        /// <param name="func">Wrapped UDF function</param>
+        /// <param name="evalType">The EvalType of the function.</param>
+        /// <param name="returnType">The return type of the function in JSON format.</param>
```
imback82: Can we remove `.` for `<param>`?

src/csharp/Microsoft.Spark/Sql/UDFRegistration.cs
Velociraptor115: Sure

src/csharp/Microsoft.Spark/Sql/UDFRegistration.cs
Velociraptor115: Sure

src/csharp/Microsoft.Spark/Sql/UDFRegistration.cs
```diff
@@ -244,14 +447,37 @@ private void Register<TResult>(string name, Delegate func)
...
+        }
+
+        /// <summary>
+        /// Helper function to register wrapped udf.
+        /// </summary>
+        /// <param name="name">Name of the udf</param>
+        /// <param name="func">Wrapped UDF function</param>
+        /// <param name="evalType">The EvalType of the function</param>
+        /// <param name="returnType">The return type of the function in JSON format</param>
+        internal void Register(string name, Delegate func, UdfUtils.PythonEvalType evalType, string returnType)
```
imback82: Can this be `private` for now?

Velociraptor115: Done

Velociraptor115:
Thanks for guiding me through my first PR @imback82 😁

imback82:
Thanks for your first contribution!

PR #457, opened 2020-03-22
Upgrade dotnet/arcade sdk to release/3.x branch

imback82:
There were build issues when VS is upgraded to 16.5. Upgrading dotnet/arcade to the latest version fixed them.


eerhardt:
I wonder if you should be using a `dotnet/arcade` version from the `release/3.x` branch, since we are still using the `3.1` SDK. See [Arcade Servicing](https://github.com/dotnet/arcade/blob/master/Documentation/Policy/ArcadeServicing.md) for information on the servicing.

At some point, I'd imagine the `5.0` version of arcade will take a dependency on the `5.0` SDK. Which means you'd have to upgrade to the 5.0 SDK.

imback82:
> I wonder if you should be using a `dotnet/arcade` version from the `release/3.x` branch, since we are still using the `3.1` SDK. See [Arcade Servicing](https://github.com/dotnet/arcade/blob/master/Documentation/Policy/ArcadeServicing.md) for information on the servicing.

Thanks @eerhardt for the suggestion. I moved to the arcade version in `release/3.x` branch.

PR #506, opened 2020-05-04
Forward Compatibility test pipeline

### LINKED ISSUE #462: [FEATURE REQUEST]: Forward compatibility E2E tests
Similar to #453, I think it's equally important to have forward compatibility tests. There are scenarios where the version of worker binaries is newer than the client side Nuget. This testing framework should help easily identify the forward compatibility.

For this, we need to set up a new project that depends on old Microsoft.Spark Nuget, and run it against the Worker in the master branch. One idea is to copy E2E tests from the known working release version, and modify it to depend on that version of Nuget.

elvaliuliuliu:
This PR adds forward compatibility by checking out the oldest compatible release version and run E2E tests against latest Microsoft.Spark.Worker from current branch.

This should fix #462


azure-pipelines.yml
```diff
@@ -10,7 +10,9 @@ variables:
   buildConfiguration: 'Release'
   _SignType: real
   _TeamName: DotNetSpark
+  MSBUILDSINGLELOADCONTEXT: 1
   oldestCompatibleWorker: '0.9.0'
+  oldestCompatibleRelease: 'v0.9.0'
```
imback82, line 15: oldestCompatibleWorker => backwardCompatibleRelease
oldestCompatibleRelease => forwardCompatibleRelease

Add comment that backwardCompatibleRelease/forwardCompatibleRelease is the "oldest" releases that work with the current release.

imback82: remove `v` and add it when you need it.

azure-pipelines.yml
```diff
@@ -27,6 +29,14 @@ variables:
   DOTNET_CLI_TELEMETRY_OPTOUT: 1
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
 
+resources:
+  repositories:
+  - repository: oldestCompatibleRelease
```
imback82, line 34: change this name as well

azure-pipelines.yml
```diff
@@ -38,29 +48,34 @@ jobs:
...
   - task: BatchScript@1
     displayName: Download Spark Distros & Winutils.exe
     inputs:
-      filename: script\download-spark-distros.cmd
+      filename: master\script\download-spark-distros.cmd
       arguments: $(Build.BinariesDirectory)
 
   - task: BatchScript@1
-    displayName: Download oldest compatible worker version
+    displayName: Download oldest compatible worker v$(oldestCompatibleWorker)
```
imback82, line 73: Download backward compatible worker

azure-pipelines.yml
elvaliuliuliu: Done

azure-pipelines.yml
```diff
@@ -169,112 +185,225 @@ jobs:
...
       command: test
-      projects: '**/Microsoft.Spark*.E2ETest/*.csproj'
+      projects: 'master/**/Microsoft.Spark*.E2ETest/*.csproj'
       arguments: '--configuration $(buildConfiguration) --filter $(TestsToFilterOut)'
     env:
       SPARK_HOME: $(Build.BinariesDirectory)\spark-2.4.5-bin-hadoop2.7
       DOTNET_WORKER_DIR: $(OldestCompatibleDotnetWorkerDir)
 
+  - task: Maven@3
+    displayName: 'Maven build src for oldest compatible release v$(forwardCompatibleRelease)'
```
imback82, line 295: forward?

azure-pipelines.yml
```diff
@@ -169,112 +185,225 @@ jobs:
...
+    inputs:
+      mavenPomFile: $(forwardCompatibleRelease)/src/scala/pom.xml
+      
+  - script: $(forwardCompatibleRelease)\build.cmd -pack
+              -c $(buildConfiguration)
+              -ci
+              $(_OfficialBuildIdArgs)
+              /p:PublishSparkWorker=true
+              /p:SparkWorkerPublishDir=$(Build.ArtifactStagingDirectory)\$(forwardCompatibleRelease)\Microsoft.Spark.Worker
+    displayName: '.NET build for oldest compatible release v$(oldestCompatibleRelease)'
```
imback82, line 305: ditto

azure-pipelines.yml
elvaliuliuliu: Do we also need to change the display name for each e2e test with forward and backward?

imback82: yes, let's do that and stick with backward compatible and forward compatible.

PR #511, opened 2020-05-07
Fix release build pipeline and avoid unnecessary artifacts copies

imback82:
Fix release build pipeline and avoid unnecessary artifacts copies

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -300,8 +300,7 @@ jobs:
               -c $(buildConfiguration)
               -ci
               $(_OfficialBuildIdArgs)
-              /p:PublishSparkWorker=true
-              /p:SparkWorkerPublishDir=$(Build.ArtifactStagingDirectory)\$(forwardCompatibleRelease)\Microsoft.Spark.Worker
+              /p:PublishSparkWorker=false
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -300,8 +300,7 @@ jobs:
               -c $(buildConfiguration)
               -ci
               $(_OfficialBuildIdArgs)
-              /p:PublishSparkWorker=true
-              /p:SparkWorkerPublishDir=$(Build.ArtifactStagingDirectory)\$(forwardCompatibleRelease)\Microsoft.Spark.Worker
+              /p:PublishSparkWorker=false
```
imback82, line 303: No need to publish worker for 0.9, which will introduce unnecessary artifacts.

script/package-worker.ps1
```diff
@@ -1,6 +1,3 @@
-Install-PackageProvider -Name NuGet -Force -Scope CurrentUser
-Install-Module -Name Microsoft.PowerShell.Archive -Scope CurrentUser -Force -AllowClobber -Verbose -MinimumVersion 1.2.3.0
```
imback82, line 2: This is no long needed and this causes release pipeline to fail.

PR #515, opened 2020-05-12
Resolve nuget dependencies for UDFs defined in dotnet-interactive

suhsteve:
In a [dotnet-interactive](https://github.com/dotnet/interactive) REPL session (driver), nuget dependencies will be systematically added to `SparkContext.AddFiles(..)`.

- `<packagename>.<version>.nupkg`
nuget packages

- `dependencyProviderMetadata`
Contains native probing paths required by `Microsoft.DotNet.DependencyManager.DependencyProvider`
Contains the assembly probing paths required by `Microsoft.DotNet.DependencyManager.DependencyProvider`
Contains nuget metadata such as filename, package name, and version to properly unpackage nuget packages.  The metadata will be delimited by the char `/`.

---------------

On the `Microsoft.Spark.Worker`, in order to resolve the nuget dependencies referenced by the `dotnet-interactive` session, we use the files added to the `SparkContext.AddFiles(...)`.  These files can be accessed from the `Worker` by checking the `SparkFiles.GetRootDirectory()` path.

There are two parts
deserialize `dependencyProviderMetadata`
1) Using the `NuGetMetadata`, we locate and unpack `.nupkg` files to `/path/to/working/directory/.nuget/packages/<package name>/<package version>`
2) We use the `AssemblyProbingPaths` and `NativeProbingPaths` to instantiate a `Microsoft.DotNet.DependencyManager.DependencyProvider` class. This class will register itself to the `AssemblyLoadContext Resolver`.

>**note**: Because `SparkContext.AddFiles(..)` prevents **_overwriting/deleting_** files once they have been added to the `SparkContext`, **numbered** identifiers are added to relevant files:
- dependencyProviderMetadata => dependencyProviderMetadata_`<zero padded ulong>`


src/csharp/Microsoft.Spark.Worker/Microsoft.Spark.Worker.csproj
```diff
@@ -6,6 +6,7 @@
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp3.1</TargetFrameworks>
     <RootNamespace>Microsoft.Spark.Worker</RootNamespace>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <RestoreAdditionalProjectSources>https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet3.1/nuget/v3/index.json</RestoreAdditionalProjectSources>
```
imback82, line 9: Indentation looks off?

imback82: Can you also add a comment which package is being downloaded from this source?

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,163 @@
...
+                }
+
+                UnpackPackages(sparkFilesPath, unpackPath, metadata.NuGets);
+
+                (s_dependencyProvider as IDisposable)?.Dispose();
+                s_dependencyProvider = new DependencyProvider(AssemblyProbingPaths, NativeProbingRoots);
+            }
+        }
+
+        private static string FindHighestFile(string src, string pattern)
```
imback82, line 131: Instead of putting these here, can't you create `DependencyProvider` class that encapsulate some of the logics introduced in this file and have corresponding unit tests? You can hide `Microsoft.DotNet.DependencyManager.DependencyProvider` in that class.

src/csharp/Microsoft.Spark/Interop/SparkEnvironment.cs
```diff
@@ -46,14 +45,22 @@ internal static Version SparkVersion
             }
         }
 
+        private static IJvmBridgeFactory s_jvmBridgeFactory = new JvmBridgeFactory();
+        internal static void SetJvmBridgeFactory(IJvmBridgeFactory jvmBridgeFactory)
+        {
+            s_jvmBridgeFactory = jvmBridgeFactory;
+        }
+
+
```
imback82, line 54: nit: extra new line?

src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
```diff
@@ -183,22 +182,23 @@ internal static JvmObjectReference CreatePythonFunction(IJvmBridge jvm, byte[] c
...
-            environmentVars.Put("DOTNET_WORKER_SPARK_VERSION", SparkEnvironment.SparkVersion.ToString());
+            // DOTNET_WORKER_SPARK_VERSION is used to handle different versions
+            // of Spark on the worker.
+            environmentVars.Put(
+                "DOTNET_WORKER_SPARK_VERSION",
+                SparkEnvironment.SparkVersion.ToString());
+
+            string replMode =
+                EnvironmentUtils.GetEnvironmentVariableAsBool("DOTNET_SPARK_REPL_MODE").ToString();
+            environmentVars.Put("DOTNET_SPARK_REPL_MODE", replMode);
```
imback82, line 201: nit: do we always to set this or only when it's running a repl?  Also, I suggested this name, but mode may not be a good name. Maybe `DOTNET_SPARK_RUNNING_REPL` or a name that conveys a boolean may be better?

src/csharp/Microsoft.Spark.Worker/Processor/PayloadProcessor.cs
```diff
@@ -87,6 +74,15 @@ internal Payload Process(Stream stream)
             }
 
             payload.IncludeItems = ReadIncludeItems(stream);
+
+#if NETCOREAPP
```
imback82, line 78: so basically, .NET framework will not work right? If so, let's remember to put this in the release note.

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,163 @@
...
+        ///
+        /// Note: Because <see cref="SparkContext.AddFile(string, bool)"/> prevents
+        /// overwriting/deleting files once they have been added to the
+        /// <see cref="SparkContext"/>, numbered identifiers are added to relevant files:
+        /// - <see cref="DependencyProviderUtils.FilePattern"/>
+        /// </summary>
+        /// <param name="stageId">The current Stage ID</param>
+        internal static void RegisterAssemblyHandler(int stageId)
+        {
+            if (!EnvironmentUtils.GetEnvironmentVariableAsBool("DOTNET_SPARK_REPL_MODE") ||
```
imback82, line 71: maybe we should cache this.

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,163 @@
...
+
+                if (string.IsNullOrEmpty(metadataFile) || metadataFile.Equals(s_lastFileRead))
+                {
+                    return;
+                }
+                s_lastFileRead = metadataFile;
+
+                using FileStream fileStream = File.OpenRead(metadataFile);
+                BinaryFormatter formatter = new BinaryFormatter();
+                DependencyProviderUtils.Metadata metadata =
```
imback82, line 101: `var`

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,163 @@
...
+                    FindHighestFile(sparkFilesPath, DependencyProviderUtils.FilePattern);
+
+                if (string.IsNullOrEmpty(metadataFile) || metadataFile.Equals(s_lastFileRead))
+                {
+                    return;
+                }
+                s_lastFileRead = metadataFile;
+
+                using FileStream fileStream = File.OpenRead(metadataFile);
+                BinaryFormatter formatter = new BinaryFormatter();
```
imback82, line 100: `var`

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,163 @@
...
+                    foreach (string dependency in metadata.NativeProbingPaths)
+                    {
+                        yield return Path.Combine(unpackPath, dependency);
+                    }
+                }
+
+                UnpackPackages(sparkFilesPath, unpackPath, metadata.NuGets);
+
+                (s_dependencyProvider as IDisposable)?.Dispose();
+                s_dependencyProvider = new DependencyProvider(AssemblyProbingPaths, NativeProbingRoots);
```
imback82, line 127: I guess this wouldn't affect the running tasks? Note that stages can run in parallel.

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,163 @@
...
+            }
+        }
+
+        private static string FindHighestFile(string src, string pattern)
+        {
+            string[] files = Directory.GetFiles(src, pattern);
+            if (files.Length > 0)
+            {
+                Array.Sort(files);
+                return files.Last();
```
imback82, line 137: The highest file should have accumulated info right? In that case should we try to remove the old files proactively or no?

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -153,7 +166,15 @@ internal static Assembly ResolveAssembly(string assemblyName)
                     }
                 }
 
-                throw new FileNotFoundException($"Assembly '{assemblyName}' file not found '{simpleAsmName}[{string.Join(",", s_extensions)}]' in '{string.Join(",", s_searchPaths)}'");
+                s_logger.LogWarn(
```
imback82, line 169: Why the behavior change?

src/csharp/Microsoft.Spark.Worker/Processor/PayloadProcessor.cs
suhsteve: I suppose we don't really need this `#if`.  It should compile fine with either framework.  The only issue I foresee is the `dotnet-interactive` session is only pulling in dotnet and netstandard nugets. The `AssemblyProbingPaths` would only have paths to these two frameworks, and if a Worker required a `net461` specific nugets, then it wouldn't be available.

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
suhsteve: I was under the impression that a Worker process would only be running parallel tasks for a specific stage.  Can a Worker process run multiple stages in parallel ?  If so, will we be locking too often ?

Changed the dispose order a bit.  I now register a new `DependencyProvider` for the new set of Assembly/Native probing paths before disposing the previously registered one.

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
suhsteve: Correct.  The highest file should contain the super set of all nuget, assembly, and native probing paths.

We can remove them proactively, but I don't think we'll have that many files.  There would only be a new file for every cell that introduces a new nuget package.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
suhsteve: Because we threw an exception in `AssemblyLoader`, the `AssemblyLoadContext.Resolving` event would stop checking any other event handlers that were registered.

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext.resolving?view=netcore-3.1
```If more than one event handler is registered for this event, the event handlers are called in order until an event handler returns a value that isn't null. Subsequent event handlers are ignored.```

If we wanted to keep the the same behavior, then we would constantly need to remove and readd our `AssemblyLoader` event handler with the `AssemblyLoadContext` every time we remove/readd the DependencyProvider.

File: src/csharp/Microsoft.Spark.Worker/Microsoft.Spark.Worker.csproj (hunk 1 above)
Comment (suhsteve) [reply src/csharp/Microsoft.Spark.Worker/Microsoft.Spark.Worker.csproj:9]: I think indentation looks off because it's a long line.  Added a comment, however in the future we should be able to remove this additional repo once this package gets published to nuget.org.

File: src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs (hunk 2 above)
Comment (suhsteve) [reply src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs:131]: I moved some of the helper methods around and added tests.  Let me know if it's sufficient.

File: src/csharp/Microsoft.Spark/Interop/SparkEnvironment.cs (hunk 3 above)
Comment (suhsteve) [reply src/csharp/Microsoft.Spark/Interop/SparkEnvironment.cs:54]: removed.

File: src/csharp/Microsoft.Spark/Utils/UdfUtils.cs (hunk 4 above)
Comment (suhsteve) [reply src/csharp/Microsoft.Spark/Utils/UdfUtils.cs:201]: Only need to set it if it's running the REPL.  Modified code a bit and renamed env variable.

File: src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs (hunk 6 above)
Comment (suhsteve) [reply src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs:71]: Cached.

File: src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs (hunk 7 above)
Comment (suhsteve) [reply src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs:101]: changed

File: src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs (hunk 8 above)
Comment (suhsteve) [reply src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs:100]: changed

File: src/csharp/Microsoft.Spark/Utils/UdfUtils.cs (hunk 4 above)
Comment (suhsteve) [reply src/csharp/Microsoft.Spark/Utils/UdfUtils.cs]: Hmm, where do you think we should set this env variable?  Should we set it from the extension or whatever is going to start the dotnet-interactive session (livy) ?

File: src/csharp/Microsoft.Spark/Utils/UdfUtils.cs [hunk 12]
```diff
@@ -183,22 +182,24 @@ internal static JvmObjectReference CreatePythonFunction(IJvmBridge jvm, byte[] c
...
-            string assemblySearchPath = string.Join(",",
-                new[]
-                {
-                    Environment.GetEnvironmentVariable(
-                        AssemblySearchPathResolver.AssemblySearchPathsEnvVarName),
-                    SparkFiles.GetRootDirectory()
-                }.Where(s => !string.IsNullOrWhiteSpace(s)));
-
+            string assemblySearchPath = Environment.GetEnvironmentVariable(
+                AssemblySearchPathResolver.AssemblySearchPathsEnvVarName);
```
imback82, line 186: Do we lose anything by removing `SparkFiles.GetRootDirectory()` or this just moved to the worker side (with the same value)?

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+#if NETCOREAPP
+using System.Runtime.Loader;
+#endif
+
+namespace Microsoft.Spark.Worker.Utils
+{
+    internal static class AssemblyLoaderHelper
+    {
+        private static readonly Lazy<bool> s_runningREPL = new Lazy<bool>(
+            () => !EnvironmentUtils.GetEnvironmentVariableAsBool("DOTNET_SPARK_RUNNING_REPL"));
```
imback82, line 18: why `!`?

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                DependencyProviderUtils.Metadata metadata =
+                    DependencyProviderUtils.Deserialize(metadataFile);
+
+                string unpackPath = Path.Combine(
+                    Directory.GetCurrentDirectory(),
+                    Path.Combine(".nuget", "packages"));
+                Directory.CreateDirectory(unpackPath);
+
+                DependencyProvider.UnpackPackages(sparkFilesPath, unpackPath, metadata.NuGets);
```
imback82, line 100: Does `AssemblyLoaderHelper` need to know all these logics to use `DependencyProvider` and `DependencyProviderUtils`? To me, most of the logic here should be behind `DependencyProvider` such that we just do:
```C#
var dependencyProvider = new DependencyProvider(sparkFilesPath);
s_dependencyProvider?.Dispose();
s_dependencyProvider = dependencyProvider;
```
and you can hide caching (`s_lastFileRead`) behind `DepedencyProvider`, no need to expose `UnpackPackages` as `static`, etc.

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Linq;
+using System.Runtime.Serialization.Formatters.Binary;
+
+namespace Microsoft.Spark.Utils
+{
+    internal class DependencyProviderUtils
+    {
+        internal static readonly string s_filePattern = "dependencyProviderMetadata_*";
+
+        internal static void Serialize(Metadata metadata, string path)
```
imback82, line 16: I would expose this as a member function of `Metadata`.

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            using FileStream fileStream = File.OpenWrite(path);
+            var formatter = new BinaryFormatter();
+            formatter.Serialize(fileStream, metadata);
+        }
+
+        internal static Metadata Deserialize(string path)
```
imback82, line 23: ditto as a static member function.

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
```diff
@@ -0,0 +1,105 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Runtime.Serialization.Formatters.Binary;
+
+namespace Microsoft.Spark.Utils
+{
+    internal class DependencyProviderUtils
```
imback82, line 12: any reason this is under `Microsoft.Spark` and `DependencyProvider` is under `Microsoft.Spark.Worker`? I guess it's coming in extension change? If so, should we just put them into one place?

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+        ///
+        /// Note: Because <see cref="SparkContext.AddFile(string, bool)"/> prevents
+        /// overwriting/deleting files once they have been added to the
+        /// <see cref="SparkContext"/>, numbered identifiers are added to relevant files:
+        /// - <see cref="DependencyProviderUtils.CreateFileName(ulong)"/>
+        /// </summary>
+        /// <param name="stageId">The current Stage ID</param>
+        internal static void RegisterAssemblyHandler(int stageId)
+        {
+            if (s_runningREPL.Value || (stageId == s_stageId))
```
imback82, line 66: I think this should be `!s_runningREPL.Value` and not `!` above right?

src/csharp/Microsoft.Spark.Worker/Processor/PayloadProcessor.cs
```diff
@@ -79,14 +65,22 @@ internal Payload Process(Stream stream)
...
-            if (Utils.SettingUtils.IsDatabricks)
+            if (SettingUtils.IsDatabricks)
             {
                 SerDe.ReadString(stream);
                 SerDe.ReadString(stream);
             }
 
             payload.IncludeItems = ReadIncludeItems(stream);
+
+            // Register additional assembly handlers after SparkFilesDir has been set
```
imback82, line 78: Maybe it's better to move few lines up to the place where SparkFilesDir is read.

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
suhsteve: good catch.

src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
suhsteve: Functionally will be the same.

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
suhsteve: Moved.

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
suhsteve: Moved

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
suhsteve: Yeah, good catch.

src/csharp/Microsoft.Spark.Worker/Processor/PayloadProcessor.cs
suhsteve: Moved.

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
suhsteve: Since the Metadata is used by the extension I put `DependencyProviderUtils` in `Microsoft.Spark`.  I only put `DependencyProvider` in `Microsoft.Spark.Worker` because the worker is the only consumer of the class.  But if it makes it easier I can move everything into `DependencyProvider` and put it in `Microsoft.Spark`

src/csharp/Microsoft.Spark.UnitTest/DependencyProviderUtilsTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.Spark.UnitTest.TestUtils;
+using Microsoft.Spark.Utils;
+using Xunit;
+
+namespace Microsoft.Spark.UnitTest
+{
+    public class DependencyProviderUtilsTests
+    {
+        [Fact]
+        public void TestSerDe()
```
imback82, line 12: nit: TestMetadataSerDe

src/csharp/Microsoft.Spark.UnitTest/DependencyProviderUtilsTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            using var tempDir = new TemporaryDirectory();
+            foreach (ulong num in Enumerable.Range(1, 20))
+            {
+                var filePath =
+                    Path.Combine(tempDir.Path, DependencyProviderUtils.CreateFileName(num));
+                File.Create(filePath).Dispose();
+            }
+
+            var expecteFile = "dependencyProviderMetadata_00000000000000000020";
```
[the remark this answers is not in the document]
imback82: and `string` I think.

imback82:
Generally LGTM, few minor comments/questions.

src/csharp/Microsoft.Spark/Utils/UdfUtils.cs
imback82: good question. I was thinking whoever is starting the dotnet process, but I guess extension can still set it as well to be sure?

imback82: I think whoever is starting up the `DotnetRunner` (or `DotnetBackend`).

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        internal static readonly string s_filePattern = "dependencyProviderMetadata_*";
+
+        internal static string FindHighestFile(string path)
```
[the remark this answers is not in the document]
suhsteve: I think that would depend on the usage.  In the extension high = latest.  I'll update method name.

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
```diff
@@ -0,0 +1,105 @@
...
+            if (files.Length > 0)
+            {
+                Array.Sort(files);
+                return files.Last();
+            }
+
+            return null;
+        }
+
+        internal static string CreateFileName(ulong number) =>
```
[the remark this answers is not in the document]
suhsteve: added sample.

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
```diff
@@ -0,0 +1,105 @@
...
+using System;
+using System.IO;
+using System.Linq;
+using System.Runtime.Serialization.Formatters.Binary;
+
+namespace Microsoft.Spark.Utils
+{
+    internal class DependencyProviderUtils
+    {
+        internal static readonly string s_filePattern = "dependencyProviderMetadata_*";
```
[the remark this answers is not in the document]
suhsteve: oops sorry.  changed.

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            public bool Equals(Metadata other)
+            {
+                return (other != null) &&
+                    ArrayEquals(AssemblyProbingPaths, other.AssemblyProbingPaths) &&
+                    ArrayEquals(NativeProbingPaths, other.NativeProbingPaths) &&
+                    ArrayEquals(NuGets, other.NuGets);
+            }
+
+            private static bool ArrayEquals<T>(T[] array1, T[] array2)
```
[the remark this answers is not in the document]
suhsteve: Moved to `CollectionUtils`

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -65,13 +74,14 @@ internal static class AssemblyLoader
...
-        internal static Func<string, Assembly> LoadFromName { get; set; } = Assembly.Load;
+        private static readonly ILoggerService s_logger =
+            LoggerServiceFactory.GetLogger(typeof(AssemblyLoader));
 
         private static readonly Dictionary<string, Assembly> s_assemblyCache =
             new Dictionary<string, Assembly>();
 
-        private static readonly string[] s_searchPaths =
-            AssemblySearchPathResolver.GetAssemblySearchPaths();
+        private static readonly Lazy<string[]> s_searchPaths =
```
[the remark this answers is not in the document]
suhsteve: Added comment.

src/csharp/Microsoft.Spark/Interop/SparkEnvironment.cs
```diff
@@ -46,14 +45,21 @@ internal static Version SparkVersion
             }
         }
 
+        private static IJvmBridgeFactory s_jvmBridgeFactory = new JvmBridgeFactory();
+        internal static void SetJvmBridgeFactory(IJvmBridgeFactory jvmBridgeFactory)
```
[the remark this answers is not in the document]
suhsteve: Updated pattern.

src/csharp/Microsoft.Spark.Worker/Utils/DependencyProvider.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// </summary>
+    internal class DependencyProvider : IDisposable
+    {
+        private static string s_lastFileRead;
+
+        private DepManager.DependencyProvider _dependencyProvider;
+        private readonly string _src;
+        private readonly string _dst;
+
+        public DependencyProvider(string src, string dst)
```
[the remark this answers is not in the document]
suhsteve: updated access modifier.

src/csharp/Microsoft.Spark.Worker/Utils/DependencyProvider.cs
```diff
@@ -0,0 +1,115 @@
...
+            foreach (DependencyProviderUtils.NuGetMetadata metadata in nugetMetadata)
+            {
+                var packageDirectory = new DirectoryInfo(
+                    Path.Combine(dst, metadata.PackageName.ToLower(), metadata.PackageVersion));
+                if (!packageDirectory.Exists)
+                {
+                    ZipFile.ExtractToDirectory(
+                        Path.Combine(src, metadata.FileName),
+                        packageDirectory.FullName);
+                }
```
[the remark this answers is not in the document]
suhsteve: Hmm.  Has the potential to be too frequent.

if we had 5 nugets in stage 1
if we introduced 1 new nuget in stage 2, then we would have 5 warning messages.
if we introduced 1 new nuget in stage 3, then we would have 6 warning messages.
etc.

I can add log and we can remove it later if there are complaints that it is too verbose ?

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+
+#if NETCOREAPP
+using System.Runtime.Loader;
+#endif
+
+namespace Microsoft.Spark.Worker.Utils
+{
+    internal static class AssemblyLoaderHelper
+    {
+        private static readonly Lazy<bool> s_runningREPL = new Lazy<bool>(
```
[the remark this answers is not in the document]
suhsteve: Nope.  Updated.

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+                s_stageId = stageId;
+
+                var dependencyProvider = new DependencyProvider(
+                    SparkFiles.GetRootDirectory(),
+                    Directory.GetCurrentDirectory());
+                if (dependencyProvider.TryLoad())
+                {
+                    s_dependencyProvider?.Dispose();
+                    s_dependencyProvider = dependencyProvider;
+                }
```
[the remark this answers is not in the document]
suhsteve: Fails as in returns false or throws an exception ?

src/csharp/Microsoft.Spark.Worker.UnitTest/DependencyProviderTests.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Microsoft.Spark.Worker.UnitTest
+{
+    [Collection("Spark Unit Tests")]
+    public class DependencyProviderTests
+    {
+        [Fact]
+        public void TestsUnpackPackages()
+        {
+            var packageFileName = "package.name.1.0.0.nupkg";
```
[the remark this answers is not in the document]
suhsteve: changed.

src/csharp/Microsoft.Spark.Worker.UnitTest/DependencyProviderTests.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                var dependencyProvider = new DependencyProvider(nupkgDir.Path, unpackDir.Path);
+                Assert.True(dependencyProvider.TryLoad());
+                string expectedPackagePath =
+                    Path.Combine(unpackDir.Path, ".nuget", "packages", packageName, packageVersion);
+                string expectedFilePath = Path.Combine(expectedPackagePath, emptyFileName);
+                Assert.True(File.Exists(expectedFilePath));
+            }
+
+            {
```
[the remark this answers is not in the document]
suhsteve: Added comments.

src/csharp/Microsoft.Spark.UnitTest/DependencyProviderUtilsTests.cs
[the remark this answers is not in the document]
suhsteve: updated.

src/csharp/Microsoft.Spark.UnitTest/DependencyProviderUtilsTests.cs
suhsteve: updated

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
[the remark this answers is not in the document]
imback82: If it returns false since we are not catching any exceptions here anyway.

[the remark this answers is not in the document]
suhsteve: It will only return false if there are no metadata files (ie, no nuget dependencies or if spark did not distribute the files to the worker) and if the cached `s_lastFileRead` matches the latest metadata file in the SparkFiles path.  In either case I don't think we need logs.

src/csharp/Microsoft.Spark.Worker/Utils/DependencyProvider.cs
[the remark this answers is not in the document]
imback82: How can you introduce new Nugets across different stages? You mean across different jobs? if those are expected frequently and innocuous, we don't need to handle them, I guess.

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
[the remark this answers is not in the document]
imback82: > ie, no nuget dependencies or if spark did not distribute the files to the worker

This sounds like a critical error case, no?

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
suhsteve: Hmm.  In the batch scenario, if the user sees this warning, it should be proceeded by an exception that will be thrown by the assembly loading class that is using this event handler.

In the interactive scenario the following may happen.  User sees this warning.  The next event handler registered by the DependencyProvider is processed.
- Finds assembly in the probing paths defined in the DependencyProvider.
- Unable to find assembly and an exception is thrown similar to the batch scenario.

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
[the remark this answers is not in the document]
suhsteve: How do we determine if it's the former or latter ?

src/csharp/Microsoft.Spark.Worker/Utils/DependencyProvider.cs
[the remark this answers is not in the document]
suhsteve: Sorry yes.  I meant single stage jobs.

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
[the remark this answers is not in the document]
imback82: It depends on how you implement `TryLoad`. If you think it's a critical error, returning false in that case may not work.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
imback82: So, is this a noise (since an exception will be always thrown) or will be this be helpful for further debugging?

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
```diff
@@ -0,0 +1,109 @@
...
+                return base.GetHashCode();
+            }
+
+            public override bool Equals(object obj)
+            {
+                return (obj is Metadata metadata) &&
+                    Equals(metadata);
+            }
+
+            public bool Equals(Metadata other)
```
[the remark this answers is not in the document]
suhsteve: changed to private.

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+                    return;
+                }
+                s_stageId = stageId;
+
+                var dependencyProvider = new DependencyProvider(
+                    SparkFiles.GetRootDirectory(),
+                    Directory.GetCurrentDirectory());
+                if (dependencyProvider.TryLoad())
+                {
+                    s_dependencyProvider?.Dispose();
```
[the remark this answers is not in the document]
suhsteve: Removed previous `Dispose` pattern.  Instead we will update the `Metadata` files to only include incremental updates and load a new `DependencyProvider` for each metadata file.

src/csharp/Microsoft.Spark.UnitTest/DependencyProviderUtilsTests.cs
```diff
@@ -0,0 +1,56 @@
...
+                }
+            };
+
+            string serializedFilePath = Path.Combine(tempDir.Path, "serializedMetadata");
+            metadata.Serialize(serializedFilePath);
+
+            DependencyProviderUtils.Metadata deserializedMetadata =
+                DependencyProviderUtils.Metadata.Deserialize(serializedFilePath);
+
+            Assert.Equal(metadata, deserializedMetadata);
```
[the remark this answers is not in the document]
suhsteve: Sorry i'm not sure what you're referring to.

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+                return base.GetHashCode();
+            }
+
+            public override bool Equals(object obj)
+            {
+                return (obj is NuGetMetadata nugetMetadata) &&
+                    Equals(nugetMetadata);
+            }
+
+            public bool Equals(NuGetMetadata other)
```
imback82, line 47: `private` here as well?

src/csharp/Microsoft.Spark.UnitTest/DependencyProviderUtilsTests.cs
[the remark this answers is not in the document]
imback82: This will invoke `Metadata.Equals`?

src/csharp/Microsoft.Spark.Worker/Utils/DependencyProvider.cs
```diff
@@ -0,0 +1,87 @@
...
+    ///   nugets.
+    /// - Uses <see cref="DependencyProviderUtils.Metadata.AssemblyProbingPaths"/> and
+    ///   <see cref="DependencyProviderUtils.Metadata.NativeProbingPaths"/> to construct
+    ///   a <see cref="DepManager.DependencyProvider"/>.
+    /// </summary>
+    internal class DependencyProvider : IDisposable
+    {
+        private readonly DepManager.DependencyProvider _dependencyProvider;
+
+        internal DependencyProvider(string metadataFile, string src, string dst)
```
imback82, line 26: nit: `metadataFilePath`, `srcPath`, `dstPath`?

src/csharp/Microsoft.Spark.UnitTest/DependencyProviderUtilsTests.cs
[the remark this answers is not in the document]
suhsteve: Yes it will call `Metadata.Equals`

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
```diff
@@ -0,0 +1,93 @@
...
+
+namespace Microsoft.Spark.Worker.Utils
+{
+    internal static class AssemblyLoaderHelper
+    {
+        private static readonly ILoggerService s_logger =
+            LoggerServiceFactory.GetLogger(typeof(AssemblyLoaderHelper));
+
+        // A mapping between a metadata file's path to its respective DependencyProvider.
+        private static readonly ConcurrentDictionary<string, Lazy<DependencyProvider>> s_depProvs =
```
imback82, line 23: can we spell it out to `s_dependencyProviders` to be consistent with other names?

suhsteve: It will exceed the 100 char limit if you are okay with this.

src/csharp/Microsoft.Spark.UnitTest/DependencyProviderUtilsTests.cs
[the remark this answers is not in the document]
imback82: Can you depend on the function to verify Ser/De since it can have a bug (not checking all fields, for example)?

src/csharp/Microsoft.Spark.Worker/Utils/DependencyProvider.cs
suhsteve: renamed

src/csharp/Microsoft.Spark.Worker/Utils/AssemblyLoaderHelper.cs
suhsteve: renamed

src/csharp/Microsoft.Spark/Utils/DependencyProviderUtils.cs
suhsteve: updated.

PR #530, opened 2020-05-29
Fix NRE for TimestampType and DateType and support nullable value types

### LINKED ISSUE #436: [BUG]: Null values cannot be read from DataFrame
**Describe the bug**
While trying to read a null value from a DataFrame a NullReferenceException occurred:
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Spark.Sql.Types.DateType.FromInternal(Object obj)
   at Microsoft.Spark.Sql.Row.Convert()
   at Microsoft.Spark.Sql.Row..ctor(Object[] values, StructType schema)
   at Microsoft.Spark.Sql.RowConstructor.GetRow()
   at Microsoft.Spark.Sql.RowCollector.Collect(ISocketWrapper socket)+MoveNext()
   at Microsoft.Spark.Sql.DataFrame.GetRows(String funcName)+MoveNext()
   ...
```

**To Reproduce**

1. Create a DataFrame with a null value. Try to enumerate the rows:
```
var dates = new[] { DateTime.Today, DateTime.Today.AddDays(-1) };
var columnName = "Date";
var df = spark.Session.CreateDataFrame(dates.Select(date => new Date(date)).Append(null)).WithColumnRenamed("_1", columnName);
foreach (var row in df.Collect())
{
    var value = row.Get("Date"); // Case1: NRE
    var value2 = row.Get<Date?>("Date"); // // Case2: NRE, null expected
}
```
2. NullReferenceException is thrown (see above)

**Additional info**
```
df.PrintSchema();

root
 |-- Date: date (nullable = true)

df.Show();

+----------+
|      Date|
+----------+
|2020-03-02|
|2020-03-01|
|      null|
+----------+
```

**Expected behavior**
- row.Get("Date") returns null instead of throwing the NRE
- row.Get<Date?>("Date") returns null instead of throwing the NRE

**Is any workaround exists to read the null value by now?**

### LINKED ISSUE #527: [BUG]: TimestampType.FromInternal does not handle nulls
**Describe the bug**
When iterating over a DataFrame with nullable timestamp columns, a NullReferenceException pops up. The issue is with TimestampType.FromInternal which does not contain a null check.

**To Reproduce**

1. Try foreach-ing the output of ToLocalIterator() on a DataFrame with timestamp columns containing nulls.
2. Notice the exception.

**Expected behavior**
It is expected that .NET for Spark should support nullable timestamp columsn just like Spark itself already does.

 - OS: Ubuntu 20.04 x64 (kernel 5.6.13)

elvaliuliuliu:
This PR will fix NullReferenceException for `TimestampType` and `DateType` by adding null checks,  thus DataFrame could be iterated over with `Timestamp` and `Date` columns containing null values.

This should fix #436 and fix #527. Thanks @borgdylan for providing fixes!

This PR also adds a new set of APIs for nullable value types when creating DataFrame with `CreateDataFrame`.


imback82:
Did you check the value types as well (https://github.com/dotnet/spark/issues/436#issuecomment-593482480)?

elvaliuliuliu:
> Did you check the value types as well ([#436 (comment)](https://github.com/dotnet/spark/issues/436#issuecomment-593482480))?

By adding `#nullable enable` like below should fix this. Please let me know if I misunderstand this.

```csharp
#nullable enable
var value2 = row.GetAs<Date?>(0);
```

imback82:
What I meant was using int, double, etc. (whether null values are supported correctly for all the datatypes we support)

elvaliuliuliu:
Got it, I will check!

imback82:
Any update on this?

elvaliuliuliu:
> Any update on this?

Added `null` check for value types. Please let me know if that makes sense.

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -187,15 +187,15 @@ public void Dispose()
...
+        public DataFrame CreateDataFrame(IEnumerable<double?> data) =>
             CreateDataFrame(ToGenericRows(data), SchemaWithSingleColumn(new DoubleType()));
 
         /// <summary>
         /// Creates a Dataframe given data as <see cref="IEnumerable"/> of type <see cref="bool"/>
         /// </summary>
         /// <param name="data"><see cref="IEnumerable"/> of type <see cref="bool"/></param>
         /// <returns>Dataframe object</returns>
-        public DataFrame CreateDataFrame(IEnumerable<bool> data) =>
+        public DataFrame CreateDataFrame(IEnumerable<bool?> data) =>
```
imback82, line 198: I don't think you can just replace this. What do you see if you run printSchema on the returned data frame? The `nullable` value should be different. If they are different (bool? => nullable = true, and bool => nullable = false), you should introduce a new set of APIs.

elvaliuliuliu: Both cases return `boolean (nullable = true)` when running `df.PrintSchema()`. I have added a new set of APIs for nullable value types.

imback82: Can you check if we can make `nullable = false` in case we pass `bool`? That's the behavior I see in Scala.

imback82: If not, we can have just one API set. I mentioned "If they are **different** (bool? => nullable = true, and bool => nullable = false), you should introduce a new set of APIs."

elvaliuliuliu: But without a new set of APIs, `var data = new List<bool>(new bool[] { true, false });` this won't work.

imback82: Ah, it should be pretty simple to handle this case. For non-nullable types, set `StructField`'s `isNullable` to `false`.

elvaliuliuliu: oh I see, let me check if we can make `nullable = false` in case we pass `bool`.

imback82: You can just update `SchemaWithSingleColumn` to take in `isNullable = true`.

imback82:
gentle ping. We plan to do a new release tomorrow and are you interested in putting this in the next release?

elvaliuliuliu:
Yes, I will work on it tonight

imback82:
great, thanks!

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
elvaliuliuliu: I have added this and make `nullable = false` when passing `bool`.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
```diff
@@ -104,6 +104,15 @@ public void TestCreateDataFrame()
             // Calling CreateDataFrame(IEnumerable<int> _) without schema
             {
                 var data = new List<int>(new int[] { 1, 2 });
```
suhsteve, line 106:
```suggestion
                var data = new int[] { 1, 2 };
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
```diff
@@ -104,6 +104,15 @@ public void TestCreateDataFrame()
...
+                StructType schema = SchemaWithSingleColumn(new IntegerType(), false);
+
+                DataFrame df = _spark.CreateDataFrame(data);
+                ValidateDataFrame(df, data.Select(a => new object[] { a }), schema);
+            }
+
+            // Calling CreateDataFrame(IEnumerable<int?> _) without schema
+            {
+                var data = new List<int?>(new int?[] { 1, 2, null });
```
suhsteve, line 115:
```suggestion
                var data = new int?[] { 1, 2, null };
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
```diff
@@ -113,6 +122,15 @@ public void TestCreateDataFrame()
             // Calling CreateDataFrame(IEnumerable<double> _) without schema
             {
                 var data = new List<double>(new double[] { 1.2, 2.3 });
```
suhsteve, line 124:
```suggestion
                var data = new double[] { 1.2, 2.3 };
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
```diff
@@ -113,6 +122,15 @@ public void TestCreateDataFrame()
...
+                StructType schema = SchemaWithSingleColumn(new DoubleType(), false);
+
+                DataFrame df = _spark.CreateDataFrame(data);
+                ValidateDataFrame(df, data.Select(a => new object[] { a }), schema);
+            }
+
+            // Calling CreateDataFrame(IEnumerable<double?> _) without schema
+            {
+                var data = new List<double?>(new double?[] { 1.2, 2.3, null });
```
suhsteve, line 133:
```suggestion
                var data = new double?[] { 1.2, 2.3, null };
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
```diff
@@ -122,18 +140,28 @@ public void TestCreateDataFrame()
             // Calling CreateDataFrame(IEnumerable<bool> _) without schema
             {
                 var data = new List<bool>(new bool[] { true, false });
```
suhsteve, line 142:
```suggestion
                var data = new bool[] { true, false };
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/SparkSessionTests.cs
```diff
@@ -122,18 +140,28 @@ public void TestCreateDataFrame()
...
+                StructType schema = SchemaWithSingleColumn(new BooleanType(), false);
+
+                DataFrame df = _spark.CreateDataFrame(data);
+                ValidateDataFrame(df, data.Select(a => new object[] { a }), schema);
+            }
+
+            // Calling CreateDataFrame(IEnumerable<bool?> _) without schema
+            {
+                var data = new List<bool?>(new bool?[] { true, false, null });
```
suhsteve, line 151:
```suggestion
                var data = new bool?[] { true, false, null };
```

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -172,6 +172,14 @@ public void Dispose()
         /// <param name="data"><see cref="IEnumerable"/> of type <see cref="int"/></param>
         /// <returns>Dataframe object</returns>
         public DataFrame CreateDataFrame(IEnumerable<int> data) =>
+            CreateDataFrame(ToGenericRows(data), SchemaWithSingleColumn(new IntegerType(), false));
+
+        /// <summary>
+        /// Creates a Dataframe given data as <see cref="IEnumerable"/> of type <see cref="Nullable{Int32}"/>
```
suhsteve, line 178: `>` 100 char

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -172,6 +172,14 @@ public void Dispose()
...
+        /// </summary>
+        /// <param name="data"><see cref="IEnumerable"/> of type <see cref="Nullable{Int32}"/></param>
```
suhsteve, line 180: `>` 100 char

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -188,6 +196,14 @@ public void Dispose()
         /// <param name="data"><see cref="IEnumerable"/> of type <see cref="double"/></param>
         /// <returns>Dataframe object</returns>
         public DataFrame CreateDataFrame(IEnumerable<double> data) =>
+            CreateDataFrame(ToGenericRows(data), SchemaWithSingleColumn(new DoubleType(), false));
+
+        /// <summary>
+        /// Creates a Dataframe given data as <see cref="IEnumerable"/> of type <see cref="Nullable{Double}"/>
```
suhsteve, line 202: `>` 100 char

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -188,6 +196,14 @@ public void Dispose()
...
+        /// </summary>
+        /// <param name="data"><see cref="IEnumerable"/> of type <see cref="Nullable{Double}"/></param>
```
suhsteve, line 204: `>` 100 char

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -196,6 +212,14 @@ public void Dispose()
         /// <param name="data"><see cref="IEnumerable"/> of type <see cref="bool"/></param>
         /// <returns>Dataframe object</returns>
         public DataFrame CreateDataFrame(IEnumerable<bool> data) =>
+            CreateDataFrame(ToGenericRows(data), SchemaWithSingleColumn(new BooleanType(), false));
+
+        /// <summary>
+        /// Creates a Dataframe given data as <see cref="IEnumerable"/> of type <see cref="Nullable{Boolean}"/>
```
suhsteve, line 218: `>` 100 char

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -196,6 +212,14 @@ public void Dispose()
...
+        /// </summary>
+        /// <param name="data"><see cref="IEnumerable"/> of type <see cref="Nullable{Boolean}"/></param>
```
suhsteve, line 220: `>` 100 char

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
```diff
@@ -101,6 +106,11 @@ public sealed class TimestampType : AtomicType
...
         internal override object FromInternal(object obj)
         {
+            if (obj == null)
+            {
+                return null;
+            }
+
             // Known issue that if the original type is "long" and its value can be fit into the
             // "int", Pickler will serialize the value as int.
             if (obj is long val)
```
suhsteve, line 116:
```suggestion
            if (obj is long)
```

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
elvaliuliuliu: Resolved all above comments. Thanks!

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
elvaliuliuliu: `val` is needed.

suhsteve: The way this if statement is structured is a bit weird, since you already have a casted `val` from the `(obj is long val)`, and in the if body you are casting obj again:
```
            if (obj is long val)
            {
                val = (long)obj;
            }
```

How about something like `long val = (obj is long v) ? v : (int)obj;` instead.

suhsteve: Or maybe something like this:

```csharp
            long val = obj switch
            {
                int i => i,
                long l => l,
                _ => throw new Exception("..."),
            };
```

elvaliuliuliu: I have updaded it accordingly . Thanks for your suggestions!

Issue #333, opened 2019-11-12
[BUG]: APIs utilizing SparkSession.getActiveSession() are broken


imback82:
As discussed in #332, the active session is stored in a thread-local variable, and since there is no one-to-one mapping from CLR thread to JVM thread, any calls to APIs (`DeltaTable.IsDeltaTable`, etc.) that utilize `SparkSession.getActiveSession()` can result in an error where the active session is not found.

One easy fix is to change the netty server on the JVM works on a single worker thread. One downside is that you cannot have multiple threads on .NET side where each thread has its own active session (although this scenario is not common). Also, we need to investigate what happens when there is a long running job and .NET GC runs to clean up objects on the JVM side since there are only one thread on the JVM side.

Another solution is to have a managed thread pool on the JVM side such that we keep track of one-to-one mapping from .NET threads to JVM threads.

PR #536, opened 2020-06-05
Support ArrayType

elvaliuliuliu:
This PR will add support to `ArrayType` which will fix part of #26 as follows:

1. Enable udf take in and return `ArrayType` by casting unpickled objects to the appropriate type. Support scenarios like simple array, array of arrays and array of `Row`s.
2. Convert internal SQL objects to c# objects when it's `ArrayType`.


laneser:
This PR is good!
Because we have lots of parquet data which contains array type, and cannot read out by dotnet spark.
And this PR did the right job.

And, I saw the same check failed error that I cannot solve:

```
Test run for D:\a\1\s\master\artifacts\bin\Microsoft.Spark.Extensions.Delta.E2ETest\Release\netcoreapp3.1\Microsoft.Spark.Extensions.Delta.E2ETest.dll(.NETCoreApp,Version=v3.1)
Microsoft (R) Test Execution Command Line Tool Version 16.6.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

A total of 1 test files matched the specified pattern.
[2020-06-05T09:17:10.1955680Z] [fv-az352] [Info] [ConfigurationService] 'DOTNETBACKEND_PORT' environment variable is not set.
[2020-06-05T09:17:10.1970219Z] [fv-az352] [Info] [ConfigurationService] Using port 5567 for connection.
[2020-06-05T09:17:10.1973451Z] [fv-az352] [Info] [JvmBridge] JvMBridge port is 5567
[2020-06-05T09:18:00.8539213Z] [fv-az352] [Error] [JvmBridge] JVM method execution failed: Static method 'forPath' failed for class 'io.delta.tables.DeltaTable' when called with 1 arguments ([Index=1, Type=String, Value=C:\Users\VssAdministrator\AppData\Local\Temp\6cd7297f-60dd-463a-a915-b0ed79c2244f\sink-delta-table], )
[2020-06-05T09:18:00.8540601Z] [fv-az352] [Error] [JvmBridge] java.lang.IllegalArgumentException: Could not find active SparkSession
```

Maybe the E2E test sometimes just cannot find the active sparksession?

In my case, just re-run the test and it will passed...

suhsteve:
> This PR is good!
> Because we have lots of parquet data which contains array type, and cannot read out by dotnet spark.
> And this PR did the right job.
> 
> And, I saw the same check failed error that I cannot solve:
> 
> ```
> Test run for D:\a\1\s\master\artifacts\bin\Microsoft.Spark.Extensions.Delta.E2ETest\Release\netcoreapp3.1\Microsoft.Spark.Extensions.Delta.E2ETest.dll(.NETCoreApp,Version=v3.1)
> Microsoft (R) Test Execution Command Line Tool Version 16.6.0
> Copyright (c) Microsoft Corporation.  All rights reserved.
> 
> Starting test execution, please wait...
> 
> A total of 1 test files matched the specified pattern.
> [2020-06-05T09:17:10.1955680Z] [fv-az352] [Info] [ConfigurationService] 'DOTNETBACKEND_PORT' environment variable is not set.
> [2020-06-05T09:17:10.1970219Z] [fv-az352] [Info] [ConfigurationService] Using port 5567 for connection.
> [2020-06-05T09:17:10.1973451Z] [fv-az352] [Info] [JvmBridge] JvMBridge port is 5567
> [2020-06-05T09:18:00.8539213Z] [fv-az352] [Error] [JvmBridge] JVM method execution failed: Static method 'forPath' failed for class 'io.delta.tables.DeltaTable' when called with 1 arguments ([Index=1, Type=String, Value=C:\Users\VssAdministrator\AppData\Local\Temp\6cd7297f-60dd-463a-a915-b0ed79c2244f\sink-delta-table], )
> [2020-06-05T09:18:00.8540601Z] [fv-az352] [Error] [JvmBridge] java.lang.IllegalArgumentException: Could not find active SparkSession
> ```
> 
> Maybe the E2E test sometimes just cannot find the active sparksession?
> 
> In my case, just re-run the test and it will passed...

@laneser this error is related to #333 .  Internally Delta will call `getActiveSession()` and if it's lucky then it will use the thread where the active session is defined.

laneser:
@suhsteve 
Do you mean that just change the
```
DeltaTable deltaTable = DeltaTable.ForPath(path);
```
To
```
DeltaTable deltaTable = DeltaTable.ForPath(_spark, path);
```
https://github.com/dotnet/spark/blob/ce2317774c25550ea3784a1af4f02befd2eea70b/src/csharp/Extensions/Microsoft.Spark.Extensions.Delta.E2ETest/DeltaTableTests.cs#L51

will fix the unit-test ?! which do not call ForPath (use GetActiveSession version)?

elvaliuliuliu:
I will try get this PR in soon!

imback82:
Thanks @elvaliuliuliu. What work is remaining to make this non-WIP?

elvaliuliuliu:
> Thanks @elvaliuliuliu. What work is remaining to make this non-WIP?

Currently, udf takes in `ArrayList` like the example [here](https://github.com/dotnet/spark/blob/3106e8e87f811979cd0a5a1957ac6bcf177be2ce/src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs#L35), I am thinking to support this. Should I include it in this PR?

imback82:
Got it. Yes, let's include it in this PR. Thanks.

rapoth:
@elvaliuliuliu Were you planning on updating this PR?

elvaliuliuliu:
Sorry got side-tracked. I will update this PR!

elvaliuliuliu:
This PR should be ready for review. I think FC and BC tests failed as expected, it should work with the current code, please advise. Thank you!

Niharikadutta:
Thanks for working on this @elvaliuliuliu !
@imback82 @suhsteve Do we want to exclude these tests from the FC and BC pipelines?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -67,7 +68,23 @@ internal static object[] GetUnpickledObjects(Stream stream, int messageLength)
                     stackCapacity: 102); // Spark sends batches of 100 rows, and +2 is for markers.
                 s_rowConstructor.Reset();
                 Debug.Assert(unpickledItems != null);
-                return (unpickledItems as object[]);
+                // Cast unpickledItems if necessary.
+                foreach (object objArr in (object[])unpickledItems)
+                {
+                    if (objArr.GetType() == typeof(object[]))
```
imback82, line 74: Can you add a detailed comment what's happening in this block? Also, would this incur perf hit since you are inspecting every row coming in?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -67,7 +68,23 @@ internal static object[] GetUnpickledObjects(Stream stream, int messageLength)
...
+                    {
+                        object obj = ((object[])objArr)[0];
+                        if (obj == null)
+                        {
+                            continue;
+                        }
+                        if (obj.GetType() == typeof(ArrayList))
+                        {
+                            return CastUnpickledItems.UnpickleArray(unpickledItems);
```
imback82, line 83: if you just return here, what will happen to the rest of rows? Would that be ok?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: I updated the comments. I am adding the `foreach` loop to deal with `null` case. Basically, this block is to check if `unpickledItems` contains `ArrayList`, if so then we need to cast it to the appropriate array type using `UnpickleArray`. I added the `break` in the loop for the perf concern. Please advise and let me know if this makes sense. Thanks!

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: I think it should be okay, since `UnpickleArray` takes `unpickledItems` and within the func itself, it will deal with all the rows when it's `ArrayList`.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
imback82: What if there are 1000 objects in `unpickledItems` and 1000th one is `ArrayList`?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
imback82: If you get `ArrayList`, all the objects in `unpickledItems` will be a type of `ArrayList`?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
imback82: Maybe the better question is https://github.com/dotnet/spark/pull/536/files#r475892642

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: Yes, from what I have observed. It will look like below:
image

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: Please see the reply below. Thanks!

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
imback82: Maybe we should change this function to `IEnumerable<object> GetUnpickledObjects` and use `yield return`.

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// Cast UnpickledItems.
+    /// </summary>
+    internal class CastUnpickledItems
+    {
+        /// <summary>
+        /// Cast UnpickledItems from ArrayList to Array.
+        /// </summary>
+        /// <param name="unpickledItems">Unpickled items</param>
+        /// <returns>Cast unpickled items</returns>
+        public static object[] UnpickleArray(object unpickledItems)
```
imback82, line 21: I don't think you are "unpickling" here, but just casting.

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,62 @@
...
+                        case TypeCode.Double:
+                            castObjArr.Add((double[])arrli.ToArray(typeof(double)));
+                            break;
+                        case TypeCode.Byte:
+                            castObjArr.Add((byte[])arrli.ToArray(typeof(byte)));
+                            break;
+                        case TypeCode.String:
+                            castObjArr.Add((string[])arrli.ToArray(typeof(string)));
+                            break;
+                        default:
```
imback82, line 52: Shall we support more types? How about array of `Row`s? or array of Arrays?

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -69,9 +69,22 @@ private DataType FromJson(JObject json)
             return this;
         }
 
-        internal override bool NeedConversion() => true;
+        internal override bool NeedConversion() => ElementType.NeedConversion();
 
-        internal override object FromInternal(object obj) => throw new NotImplementedException();
+        internal override object FromInternal(object obj)
```
imback82, line 74: Do we have a test that captures `Collect` with array types?

elvaliuliuliu: This [ udf test](https://github.com/dotnet/spark/blob/e162ff8731f9cb53668a807ca421daa24f2b9723/src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs#L52) should cover `Collect`.

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: Yea, I am thinking of mapping the supported array type [here](https://github.com/dotnet/spark/blob/6bdc9befbe5a619076e36f1459e9a8669490c494/src/csharp/Microsoft.Spark/Interop/Ipc/PayloadHelper.cs#L311). I will add `double[][]`. What do you think?

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: You are right, changed!

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <summary>
+    /// Cast UnpickledItems.
+    /// </summary>
+    internal class CastUnpickledItems
+    {
+        /// <summary>
+        /// Cast UnpickledItems from ArrayList to Array.
+        /// </summary>
+        /// <param name="unpickledItems">Unpickled items</param>
+        /// <returns>Cast unpickled items</returns>
```
Niharikadutta, line 20: Can we make the return type more explicit/descriptive here?

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,62 @@
...
+        public static object[] CastToArray(object unpickledItems)
+        {
+            var castUnpickledItems = new List<object>();
+            foreach (object[] objArr in (object[])unpickledItems)
+            {
+                var castObjArr = new List<object>();
+                ArrayList arrli = (ArrayList)objArr[0];
```
Niharikadutta, line 27: Let's rename `arrli` to maybe `arrList` or something similar?

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: Changed!

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: Sounds good, changed.

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,62 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.Spark.Sql
+{
+    /// <summary>
+    /// Cast UnpickledItems.
+    /// </summary>
+    internal class CastUnpickledItems
```
Niharikadutta, line 14: Can we make this class static?

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </summary>
+        /// <param name="unpickledItems">Unpickled items</param>
+        /// <returns>Array after casting</returns>
+        public static object[] CastToArray(object unpickledItems)
+        {
+            var castUnpickledItems = new List<object>();
+            foreach (object[] objArr in (object[])unpickledItems)
+            {
+                var castObjArr = new List<object>();
+                ArrayList arrList = (ArrayList)objArr[0];
```
Niharikadutta, line 27:
```suggestion
                var arrList = (ArrayList)objArr[0];
```

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: Yes, may I ask why this should be a static class?

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: Changed, thanks!

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
Niharikadutta: Is there any need to instantiate this class? If not, we should make it static.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: Do you mean something like below?
```csharp
        internal static IEnumerable<object> GetUnpickledObjects(Stream stream, int messageLength)
        {
            byte[] buffer = ArrayPool<byte>.Shared.Rent(messageLength);

            try
            { 
                ...
                // Check if unpickler returns ArrayList.
                // If so, it needs to be cast to the appropriate array type using CastToArray.
                foreach (object objArr in (object[])unpickledItems)
                {
                    if (objArr.GetType() == typeof(object[]))
                    {
                        object obj = ((object[])objArr)[0];
                        if (obj == null)
                        {
                            continue;
                        }
                        if (obj.GetType() == typeof(ArrayList))
                        {
                            yield return CastUnpickledItems.CastToArray(unpickledItems);
                        }
                        else
                        {
                            yield return unpickledItems;
                        }
                    }
                    else
                    {
                        yield return unpickledItems;
                    }
                }
            }
            finally
            {
                ArrayPool<byte>.Shared.Return(buffer);
            }
        }
```

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: Added support for array of `Row`s and array of Arrays. Thanks!

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,134 @@
...
+                                    foreach (ArrayList al in arrList)
+                                    {
+                                        stringArr.Add((string[])al.ToArray(typeof(string)));
+                                    }
+                                    castObjArr.Add(stringArr.ToArray());
+                                    break;
+                                default:
+                                    throw new NotSupportedException(
+                                        string.Format("Array of type {0} not supported yet", t));
+                            }
```
Niharikadutta, line 99: Can you abstract this logic into a separate function and use that instead of this duplication?

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// </summary>
+        /// <param name="unpickledItems">Unpickled items that contains array of rows</param>
+        /// <returns>Array of rows after casting</returns>
+        public static object[] CastToRowArray(object unpickledItems)
+        {
+            var castUnpickledItems = new List<object>();
+            foreach (RowConstructor rowConstructor in (object[])unpickledItems)
+            {
+                var castObjArr = new List<object>();
+                Row row = (rowConstructor as RowConstructor).GetRow();
```
Niharikadutta, line 122: Remove redundant casting

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,134 @@
...
+    /// Cast UnpickledItems.
+    /// </summary>
+    internal static class CastUnpickledItems
+    {
+        /// <summary>
+        /// Cast unpickledItems from arraylist to the appropriate array.
+        /// </summary>
+        /// <param name="unpickledItems">Unpickled items that contains simple array</param>
+        /// <returns>Simple array after casting</returns>
+        public static object[] CastToSimpleArray(object unpickledItems)
```
suhsteve, line 21: Looks like you have a case for 2d jagged arrays.  What about 3,4,..,n d arrays ?

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -69,9 +69,22 @@ private DataType FromJson(JObject json)
...
+        {
+            if (!NeedConversion())
+            {
+                return obj;
+            }
+
+            var convertedObj = new List<object>();
+            foreach(object o in (dynamic)obj)
```
suhsteve, line 82: Why `dyanmic` ?

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -60,21 +80,63 @@ public void TestUdfWithArrayType()
...
+                {
+                new string[] { "Michael", "MichaelMichael" },
+                new string[] { "Andy", "AndyAndy" },
+                new string[] { "Justin", "JustinJustin" }
+                };
+
+                for (int i = 0; i < rows.Length; ++i)
+                {
+                    Assert.Equal(1, rows[i].Size());
+                    Assert.Equal(expected[i], rows[i].Get(0));
```
suhsteve, line 100: What's the type of `rows[i].Get(0)` ?

suhsteve:
We are targeting a release by end of month and we would like this PR in within the next 2-3 days.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -34,24 +34,44 @@ public UdfComplexTypesTests(SparkFixture fixture)
...
+
+                var expected = new string[] { "101", "101.1", "101.2" };
+                string[] rowsToArray = rows.Select(x => x[0].ToString()).ToArray();
+                Assert.Equal(expected, rowsToArray);
+            }
+
+            // Array of Rows.
+            {
+                Func<Column, Column> udf = Udf<Row[], string>(
+                rowArr => rowArr[0].GetAs<string>("city"));
```
suhsteve, line 66: Can you also test adding in an array as a field to the row ?

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
suhsteve: I agree with @imback82 .  You are iterating through all the object[] entries here, and the consumer of `GetUnpickledObjects` will also end up iterating through the object[] again.  It may be better to make this an `IEnumerable`

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
```diff
@@ -67,7 +68,48 @@ internal static object[] GetUnpickledObjects(Stream stream, int messageLength)
...
+                        }
+                        else
+                        {
+                            break;
+                        }
+                    }
+                    else
+                    {
+                        break;
+                    }
```
suhsteve, line 110: Do we need all these if/else/continue/breaks?  This seems fishy and the code/logic can be redone to avoid all this ?

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: Yes, changed. Thanks!

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: Removed.

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
elvaliuliuliu: `obj` could be `ArrayList`, `Date[]` , etc

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Added the type in the test.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: You are right, this part needs to be redone. I will update tonight.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: Sure, will make the change tonight.

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: Will check and add the support for n d arrays

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Sure, will add the test tonight.

suhsteve:
@elvaliuliuliu how's the progress going?  Tests are failing.

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
elvaliuliuliu: Added this test.

elvaliuliuliu:
> @elvaliuliuliu how's the progress going? Tests are failing.

Working on n n arrays. Some of the tests are passing, it failed at different spark version with error "System.IO.DirectoryNotFoundException : Could not find a part of the path 'D:\a\1\b\spark-2.3.4-bin-hadoop2.7\RELEASE'", looking into it.

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: Added support for n d arrays.

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
```diff
@@ -81,9 +81,9 @@ public sealed class DateType : AtomicType
         /// </summary>
         internal override object FromInternal(object obj)
         {
-            if (obj == null)
+            if (obj == null || obj.GetType() == typeof(Date))
```
suhsteve, line 84: This seems fishy.  You shouldn't be passing in an already converted type to this method.

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
```diff
@@ -106,9 +106,9 @@ public sealed class TimestampType : AtomicType
         /// </summary>
         internal override object FromInternal(object obj)
         {
-            if (obj == null)
+            if (obj == null || obj.GetType() == typeof(Timestamp))
```
suhsteve, line 109: ditto

src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs
```diff
@@ -34,24 +34,46 @@ public UdfComplexTypesTests(SparkFixture fixture)
...
+                Assert.Equal(3, rows.Length);
+
+                var expected = new string[] { "1", "3,5", "2,4" };
+                string[] rowsToArray = rows.Select(x => x[0].ToString()).ToArray();
+                Assert.Equal(expected, rowsToArray);
+            }
+
+            // Array of Arrays.
+            {
+                Func<Column, Column> udf = Udf<double[][], double>(
```
suhsteve, line 52: can you pass in multiple arrays ?

Udf<double, double[], double[][], double[][][], double>  ?

elvaliuliuliu: Added tests for multiple arrays. Working on `ArrayList` cases as mentioned.

elvaliuliuliu: Do we also want to support the option of return value as `ArrayList`?
Like `Udf<double, ArrayList, ... , ArrayList>`?

@imback82 ideas?

imback82: I guess it can be useful if you want to chain with Udf that takes in `ArrayList` (just for consistency).

elvaliuliuliu: Got it, thanks!
 
The current logic is to cast `unpickledItems` to typed array if it contains `ArrayList` [here](https://github.com/dotnet/spark/blob/46b0d77d11fa916f92ad23d15cde794ad4ca5a68/src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs#L137). Another way might be casting input(after unpickling) [here](https://github.com/dotnet/spark/blob/46b0d77d11fa916f92ad23d15cde794ad4ca5a68/src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs#L152) when executing udf functions?

I am thinking either to pass [commands](https://github.com/dotnet/spark/blob/46b0d77d11fa916f92ad23d15cde794ad4ca5a68/src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs#L100) in `GetUnpickledObjects` or switch to the latter method, if we want to support `ArrayList` like `Udf<ArrayList, ArrayList, double>`. Please  suggest. Thanks.

suhsteve:
> We are targeting a release by end of month and we would like this PR in within the next 2-3 days.

Can we get everything wrapped up by today ?

elvaliuliuliu:
> > We are targeting a release by end of month and we would like this PR in within the next 2-3 days.
> 
> Can we get everything wrapped up by today ?

I am finishing up the following left parts now:

- [Support `ArrayList` when defining udf](https://github.com/dotnet/spark/pull/536#discussion_r490007140). **Need inputs/ideas**.
- [Change to `IEnumerable<object> GetUnpickledObjects`](https://github.com/dotnet/spark/pull/536#discussion_r488135842).
- Clean up and other changes. ([`CastToRowArray()`](https://github.com/dotnet/spark/blob/13fff5dc28d5821800215f6a515546f92b3c49e7/src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs#L45), [`GetUnpickledObjects`](https://github.com/dotnet/spark/blob/13fff5dc28d5821800215f6a515546f92b3c49e7/src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs#L74), 2 comments for [simpletypes](https://github.com/dotnet/spark/pull/536#discussion_r489593474)).

Please lmk if more comments. Thanks.

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
elvaliuliuliu: Changed.

src/csharp/Microsoft.Spark/Sql/Types/SimpleTypes.cs
elvaliuliuliu: Changed.

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: I have refactored this part. Please take a look, thanks!

src/csharp/Microsoft.Spark/Utils/PythonSerDe.cs
elvaliuliuliu: Redone.

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        /// <param name="unpickledItems">Unpickled objects to be cast as necessary.</param>
+        /// <returns>Unpickled objects after casting.</returns>
+        public static object[] Cast(object unpickledItems)
+        {
+            var castUnpickledItems = new List<object>();
+            foreach (object obj in (object[])unpickledItems)
+            {
+                castUnpickledItems.Add(
+                    (obj.GetType() == typeof(RowConstructor)) ? obj : CastArray(obj));
```
imback82, line 29: I am not sure if falling back to `CastArray` always works.

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        /// <summary>
+        /// Cast simple array and array of arrays.
+        /// </summary>
+        /// <param name="obj">object to be cast as necessary.</param>
+        /// <returns>Typed array after casting.</returns>
+        public static object CastArray(object obj)
+        {
+            if (obj is object[] objArr)
```
imback82, line 42: does this work with `Row[]`?

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                return objArr.Select(x => CastHelper(x)).ToArray();
+            }
+
+            // Array of arrays.
```
imback82, line 47: how about array of array of array? Don't you need to handle this recursively?

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: I have updated the function name accordingly. Sorry for the confusion. If not `RowConstructor`, it will fall into `object []`. `CastHelper` will help decide and cast `object[]` as needed.

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: From what I have observed, if udf takes `Row[]` like the example [here](https://github.com/dotnet/spark/blob/25abbfa90357b658c23db29e08cacb67b82361a1/src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs#L73). It will be `RowConstructor`, and [`FromInternal`](https://github.com/dotnet/spark/blob/25abbfa90357b658c23db29e08cacb67b82361a1/src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs#L299) should handle such cases.

src/csharp/Microsoft.Spark/Sql/CastUnpickledItems.cs
elvaliuliuliu: I have changed the function name and description accordingly to make it more clear. This should be handled recursively as covered in [test](https://github.com/dotnet/spark/blob/25abbfa90357b658c23db29e08cacb67b82361a1/src/csharp/Microsoft.Spark.E2ETest/UdfTests/UdfComplexTypesTests.cs#L52). Thanks!

suhsteve:
Closing as #670 has been merged.

PR #542, opened 2020-06-09
Expose logger for override by consumer

### LINKED ISSUE #473: [FEATURE REQUEST]: Allow injection of custom LoggerService
It's not possible to redirect all logging to the applications own logging pipeline currently as both `LoggerServiceFactory` and `ILoggerService` are internal.

I would like to inject my own logging service (either [seq](https://datalust.co/) or Application Insights depending on environment) and have all logs consolidated.

It seems log4net is suppoerted via config file however I have other correlation information that I need to add which I cannot see how to accomplish.

Is this possible or is there another way to achieve this?

thrixton:
Expose `LoggerServiceFactory` and `ILoggerService` as public for override by consuming applications

Fixes #473

### DIFF
--- a/src/csharp/Microsoft.Spark/Services/ILoggerService.cs
+++ b/src/csharp/Microsoft.Spark/Services/ILoggerService.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Spark.Services
     /// <summary>
     /// Defines a logger what be used in service
     /// </summary>
-    internal interface ILoggerService
+    public interface ILoggerService
     {
         /// <summary>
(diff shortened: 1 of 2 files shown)


sunildixit:
Is there a reason this is still out here? Can this be merged?

sunildixit:
@imback82 Is there a reason this is held back at this time?

imback82:
@sunildixit The reason was to fix this in a right way: https://github.com/dotnet/spark/issues/473#issuecomment-643663712

There is a workaround if you need to access these: https://github.com/dotnet/spark/issues/473#issuecomment-608260897

PR #547, opened 2020-06-14
Support Spark 2.4.6

imback82:
Spark 2.4.6 was released recently and we will support it in the next release (0.12).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -551,7 +551,7 @@ stages:
       inputs:
         command: test
         projects: '**/Microsoft.Spark*.E2ETest/*.csproj'
-        arguments: '--configuration $(buildConfiguration)'
+        arguments: '--configuration $(buildConfiguration) --filter $(TestsToFilterOut)'
       env:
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -526,4 +535,14 @@ stages:
         projects: '**/Microsoft.Spark*.E2ETest/*.csproj'
         arguments: '--configuration $(buildConfiguration) --filter $(TestsToFilterOut)'
       env:
-        SPARK_HOME: $(Build.BinariesDirectory)\spark-2.4.5-bin-hadoop2.7
\ No newline at end of file
+        SPARK_HOME: $(Build.BinariesDirectory)\spark-2.4.5-bin-hadoop2.7
+
+    - task: DotNetCoreCLI@2
+      displayName: 'E2E tests for Spark 2.4.6'
```
imback82, line 541: Note that we cannot add this for forward compatibility tests since the driver JAR (0.9.0) doesn't support this Spark version.

Issue #539, opened 2020-06-09
[FEATURE REQUEST]: run debugger as a process


mikeTWC1984:
**Is your feature request related to a problem? Please describe.**

I'm trying to adopt .net spark to use with powershell. It's actually works great. Powershell need to submit debugger first, which you can do manually within interactive shell. However this is very inconvenient. I'd like powershell to start debugger automatically. The problem is that debugger can only run in interactive shell. If you run command below as a child process it will quit righ after "press enter" message

```
 spark-submit --class org.apache.spark.deploy.dotnet.DotnetRunner microsoft-spark-2.4.x-0.10.0.jar debug
```
There are some hacky ways to emulate interactive shell in the background, but if debugger can run as foreground process it would solve the problem. It would also be helpful for containerizing it.

Other option is to start debugger somehow automatically. Say if you run ./net/powershell without submitting spark job, it will start looking for debugger and throw error. Instead it could start it automatically. So this way we can get similar experience to PySpark, it doesn't require to start debugger explicitly to run scripts

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

mikeTWC1984:
Turns out it was easy fix (just commenting few lines). Will just use a custom version of this jar

JavierLight:
Hi @mikeTWC1984 , could you please provide a more detailed description of your fix? I'm trying to containerize a simple dotnet app and can't get the debugger to work, because I need to run two interactive processes in the docker container entrypoint.

I'm using this to launch my `DotnetApp.dll` once the 5567 port is listening:

`( tail -f -n0 sparkdotnetdebug.log & ) | grep -q "Press enter to exit"`

Is there another way to execute code from this library in a dotnet DLL without the need of the debugger?

mikeTWC1984:
@JavierLight  What I did just removed those 3 lines
https://github.com/dotnet/spark/blob/c8992e0787c6f50f2aa3572839fe67b872a3363a/src/scala/microsoft-spark-2.4.x/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala#L158-L160
Then you can stop debugger with ctrl+C. You can also use netstat (both windows and linux has it) to find out which process is listening on debug port and kill it. 
You would need to rebuild your target jar, but it's fairly simple (there are instructions on this repo, as far as I remember you just need jdk and maven)

PR #562, opened 2020-06-22
Launch dotnet backend jvm bridge process ondemand

laneser:

ref https://github.com/dotnet/spark/issues/539

The SparkSession builder pattern could detect the backend jvm bridge is running or not, then launch the jvm bridge process on-demand.

In old way, spark-submit launch jvm bridge, the jvm bridge launch dotnet process. 
But that is inconvenient to debug.

The SparkSession builder pattern doesn't need jvm bridge launched first.

To launch a default SparkSession
```C#
```
And dotnet run could do the spark job.

To launch a local SparkSession
```C#
var spark = SparkSession.Builder().Master("local").GetOrCreate();
```

No need to setup those config at spark-submit.

But still support the old method, because we could detect if the jvm bridge is started or not.


imback82:
Thanks @laneser for the PR. However, I am not fully convinced with this approach:
- This requires another set of documentation: e.g., it will not work if `SPARK_HOME` is not set.
- There are objects you can create other than `SparkSession` which invokes getting the `JvmBridge`, e.g., `SparkConf`.

Is there anything we can to update the current doc to make the experience better instead of introducing another option?

I am open for further discussion to make the debugging experience seamless. cc: @suhsteve

laneser:
Thanks for your information!
I will try to setup the dotnet spark jvm bridge through the `spark.driver.extraClassPath` and let `buildSparkSubmitCommand` do the same things for dotnet spark.

laneser:
I think the JVMBridgeHelper works like

pyspark launch gateway method

https://github.com/apache/spark/blob/feeca63198466640ac461a2a34922493fa6162a8/python/pyspark/context.py#L319

https://github.com/apache/spark/blob/95aec091e4d8a45e648ce84d32d912f585eeb151/python/pyspark/java_gateway.py#L40

The SparkConf could not work if jvm bridge does not exists.

laneser:
Though the checks have passed, but I have no idea why it was not working before... 😿

PR #541, opened 2020-06-09
Expose JVM exceptions

thrixton:
Fixes #472

Further scope to parse and expose further information from JVM exceptions but as a base it's a start.

Issue #502, opened 2020-04-30
[FEATURE REQUEST]: Expose exceptions


michael-damatov:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when any exception I get is an `Exception`. For example, when reading a delta file fails, I do not know why: it could be a concurrency error, or, simply, the file doesn't exist. Reading/parsing the exception text is not a viable option.

**Describe the solution you'd like**
Please introduce more exception types and extend the documentation on methods to describe what exceptions could be thrown.

michael-damatov:
possible duplicate of #472

imback82:
@michael-damatov Would #472 address your issue, or do you need more info? You can help reviewing the PR as well.

michael-damatov:
Simply throwing `JvmException` instead of `Exception` does not help to distinguish between different cases.

#472 is definitely not a duplicate of this issue.

imback82:
https://github.com/dotnet/spark/issues/472#issuecomment-643057149 would not be enough? Exposing more exception types is not a scalable solution.

michael-damatov:
I do not expect from an exception to be scalable because it has already been thrown. I'm pretty sure that the scala code somehow distinguish between exceptions: either by using different exception types or by providing additional properties (e.g. "error code").

If the `JvmException` added with #472 would expose additional properties, which are helpful to distinguish between different cases, it would be enough.

rapoth:
Since #566 has been merged, I am closing this issue. Please feel free to re-open if it has not answered your question.

PR #566, opened 2020-06-25
Expose JVM exceptions

### LINKED ISSUE #472: [FEATURE REQUEST]: Expose JVM exceptions
Currently an exception is only logged and then translated per the below.
```
[2020-04-03T00:31:33.1953860Z] [0ecc54d4f7e9] [Error] [JvmBridge] JVM method execution failed: Nonstatic method sql failed for class 435 when called with 1 arguments ([Index=1, Type=String, Value=select
 NOTACOLUMN from tmp], )
[2020-04-03T00:31:33.1955104Z] [0ecc54d4f7e9] [Error] [JvmBridge] org.apache.spark.sql.AnalysisException: cannot resolve '`NOTACOLUMN`' given input columns: [tmp.Parameter]; line 1 pos 7;
'Project ['NOTACOLUMN]
+- SubqueryAlias `tmp`
   +- Relation[Parameter#20315] parquet<...SNIP...>
```
Is it possible maybe throw an AggregateException with both exceptions or pass an InnerException containing the full original exception?

thrixton:
Wrap spark exceptions
Fixes #472
Ref also #541, #502


src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmBridgeTests.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
```
suhsteve, line 1: Can you add the license header ?

src/csharp/Microsoft.Spark/JvmException.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
```
suhsteve, line 1: Can you add the license header ?

src/csharp/Microsoft.Spark/JvmException.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Microsoft.Spark
+{
+    /// <summary>
+    /// Contains the message returned from the <see cref="Interop.Ipc.JvmBridge"/> on an error
```
suhsteve, line 6:
```suggestion
    /// Contains the message returned from the <see cref="Interop.Ipc.JvmBridge"/> on an error.
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmBridgeTests.cs
thrixton: Done

src/csharp/Microsoft.Spark/JvmException.cs
thrixton: Done

PR #570, opened 2020-06-27
[Bug] AssemblyLoader should use absolute assembly path when loading assemblies

suhsteve:
`AssemblyLoader.ResolveAssembly` uses `AssemblySearchPathResolver.GetAssemblySearchPaths` to build paths when checking for assemblies.  An assembly was found to exist within one of the built paths, but this was a relative path.  When resolving dotnet core assemblies, `AssemblyLoader.ResolveAssembly` uses `AssemblyLoadContext.Default.LoadFromAssemblyPath` and this requires an `absolute` assembly path.

### DIFF
--- a/src/csharp/Microsoft.Spark.UnitTest/AssemblyLoaderTests.cs
+++ b/src/csharp/Microsoft.Spark.UnitTest/AssemblyLoaderTests.cs
@@ -23,7 +23,6 @@ namespace Microsoft.Spark.UnitTest
             _mockJvm = _fixture.MockJvm;
         }
 
-
         [Fact]
         public void TestAssemblySearchPathResolver()
         {
--- a/src/csharp/Microsoft.Spark.UnitTest/AssemblyLoaderTests.cs
+++ b/src/csharp/Microsoft.Spark.UnitTest/AssemblyLoaderTests.cs
@@ -4,7 +4,11 @@
 
 using System;
 using System.IO;
+using System.Reflection;
+using System.Runtime.Loader;
+using Microsoft.Spark.Interop.Ipc;
 using Microsoft.Spark.Utils;
+using Moq;
 using Xunit;
 
 namespace Microsoft.Spark.UnitTest
@@ -12,6 +16,14 @@ namespace Microsoft.Spark.UnitTest
     [Collection("Spark Unit Tests")]
     public class AssemblyLoaderTests
     {
+        private readonly Mock<IJvmBridge> _mockJvm;
+
+        public AssemblyLoaderTests(SparkFixture _fixture)
+        {
+            _mockJvm = _fixture.MockJvm;
+        }
+
+
         [Fact]
         public void TestAssemblySearchPathResolver()
         {
@@ -44,6 +56,18 @@ namespace Microsoft.Spark.UnitTest
             Environment.SetEnvironmentVariable(
                 AssemblySearchPathResolver.AssemblySearchPathsEnvVarName,
                 null);
+        }
+
+        [Fact]
+        public void TestResolveAssemblyWithRelativePath()
+        {
+            _mockJvm.Setup(m => m.CallStaticJavaMethod(
+                "org.apache.spark.SparkFiles",
+                "getRootDirectory"))
+                .Returns(".");
+
+            AssemblyLoader.LoadFromFile = AssemblyLoadContext.Default.LoadFromAssemblyPath;
+            AssemblyLoader.ResolveAssembly(Assembly.GetExecutingAssembly().FullName);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -189,12 +189,12 @@ private static bool TryLoadAssembly(string assemblyFileName, ref Assembly assemb
...
             {
-                string assemblyPath = Path.Combine(searchPath, assemblyFileName);
-                if (File.Exists(assemblyPath))
+                FileInfo assemblyFile = new FileInfo(Path.Combine(searchPath, assemblyFileName));
+                if (assemblyFile.Exists)
                 {
                     try
                     {
-                        assembly = LoadFromFile(assemblyPath);
+                        assembly = LoadFromFile(assemblyFile.FullName);
```
suhsteve, line 197: Instead of making modifications here, we can instead change line 63.
from:
```csharp
searchPaths.Add(sparkFilesPath);
```
to:
```csharp
searchPaths.Add(Path.GetFullPath(sparkFilesPath));
```

imback82:
Can you update the description please?

src/csharp/Microsoft.Spark.UnitTest/AssemblyLoaderTests.cs
```diff
@@ -4,14 +4,26 @@
...
     [Collection("Spark Unit Tests")]
     public class AssemblyLoaderTests
     {
+        private readonly Mock<IJvmBridge> _mockJvm;
+
+        public AssemblyLoaderTests(SparkFixture _fixture)
+        {
+            _mockJvm = _fixture.MockJvm;
+        }
+
```
imback82, line 25: nit: extra empty line.

src/csharp/Microsoft.Spark.UnitTest/AssemblyLoaderTests.cs
```diff
@@ -45,5 +57,17 @@ public void TestAssemblySearchPathResolver()
...
+        [Fact]
+        public void TestResolveAssemblyWithRelativePath()
+        {
+            _mockJvm.Setup(m => m.CallStaticJavaMethod(
+                "org.apache.spark.SparkFiles",
+                "getRootDirectory"))
+                .Returns(".");
+
+            AssemblyLoader.LoadFromFile = AssemblyLoadContext.Default.LoadFromAssemblyPath;
+            AssemblyLoader.ResolveAssembly(Assembly.GetExecutingAssembly().FullName);
```
imback82, line 70: So, does this throw without the fix? Also, we can do `Assert.NoThrows` to be explicit?

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
imback82: I think this is a better place since we have multiple calls to `searchPaths.Add`.

src/csharp/Microsoft.Spark/Utils/AssemblyLoader.cs
```diff
@@ -189,12 +189,12 @@ private static bool TryLoadAssembly(string assemblyFileName, ref Assembly assemb
         {
             foreach (string searchPath in s_searchPaths.Value)
             {
-                string assemblyPath = Path.Combine(searchPath, assemblyFileName);
-                if (File.Exists(assemblyPath))
+                FileInfo assemblyFile = new FileInfo(Path.Combine(searchPath, assemblyFileName));
```
imback82, line 192: nit: `var`

src/csharp/Microsoft.Spark.UnitTest/AssemblyLoaderTests.cs
suhsteve: Yup.  Without the fix this will throw the exception that we saw.

src/csharp/Microsoft.Spark.UnitTest/AssemblyLoaderTests.cs
suhsteve: Yup.  Without the fix this will throw the exception that we saw.

suhsteve: I didn't find a `Assert.NoThrows` method.  Also `Assert.DoesNotThrow` was removed from Xunit https://github.com/xunit/xunit/issues/188

imback82: then can we at least verify the returned `Assembly`?

PR #573, opened 2020-06-29
Ml/missing methods

GoEddie:
This change adds a base class that the other Spark.ML.Features objects can use so each object doesn't have to re-implement the same methods. 

This also extends the existing features with the params methods: GetParams, Set, Clear, ExplainParams, ExplainParam.


imback82:
Thanks @GoEddie for the PR! Can you fix the tests please? (Is the hang from your changes or not?)

> This also extends the existing features with the params methods: GetParams, Set, Clear, ExplainParams, ExplainParam.

Also, can this be done as a separate PR to minimize the changes for this PR (or is this tightly coupled)?

GoEddie:
@imback82 no idea about the tests, i'll re-submit this as a new pr (will be easier to separate) and if that still causes a hang I will investigate it

PR #581, opened 2020-07-06
[WIP] possible approach to fixing #333 get/set SparkSession on all calls across to the JVM

GoEddie:
Hi - I was wondering if this was a possible approach to fixing #333?

The idea is that when we return the response from the JVM to take a note of the current SparkSession, store it the SparkSession and return the hash code which is stored on the managed thread in c#.

When making a call from c#, include the hash code of the last set SparkSession on the managed thread and calling SparkSession.setActiveSession

If this isn't an approach you like let me know, happy to bin it but it works for me so delta calls stop failing.

If this is a good approach, i'll redo it a bit tider!

### DIFF
--- a/src/csharp/Microsoft.Spark/Interop/Ipc/IJvmBridge.cs
+++ b/src/csharp/Microsoft.Spark/Interop/Ipc/IJvmBridge.cs
@@ -65,5 +65,6 @@ namespace Microsoft.Spark.Interop.Ipc
             JvmObjectReference objectId,
             string methodName,
             params object[] args);
+        
     }
 }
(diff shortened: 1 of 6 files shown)


GoEddie:
The tests are failing as I only put the scale changes into 2.4 to demo it.

imback82:
Thanks @GoEddie for the PR. However, the issue is not about just `SparkSession`, but anywhere thread local variables are used on the Scala side. I think we should follow the approach suggested in #333 where a .NET thread is mapped to a thread in JVM. We are also tracking this as a GA (v1.0) blocker.

PR #588, opened 2020-07-09
Run Delta Lake tests against Delta Lake 0.6.1

AFFogarty:
This PR updates`Microsoft.Spark.Extensions.Delta.E2ETest` to run tests against Delta Lake 0.6.1, which is the latest version of Delta Lake that targets Spark 2.4.x.


AFFogarty:
Not sure why an unrelated test failed.

AFFogarty:
Ok, looks like the tests pass now.

PR #609, opened 2020-07-31
removing .ionide folder

GoEddie:
My last pr included a .ionide folder which was created by the f# vs code extension, this removes the folder and adds an entry to .gitignore so it doesn't come back.


GoEddie:
Can someone rerun the build? When I try to re-run I get a message to say I am not authorised. I don’t believe that the build failure was caused by this commit (might be wrong but would be surprised!)

imback82:
You can push an empty commit to trigger as well.

GoEddie:
Ok thanks

PR #612, opened 2020-08-05
udf-broadcast-variable-concurrency-fix

bolcman:
adding additional  FileShare.Read flag.

### DIFF
--- a/src/csharp/Microsoft.Spark.Worker/Processor/BroadcastVariableProcessor.cs
+++ b/src/csharp/Microsoft.Spark.Worker/Processor/BroadcastVariableProcessor.cs
@@ -54,7 +54,8 @@ namespace Microsoft.Spark.Worker.Processor
                     else
                     {
                         string path = SerDe.ReadString(stream);
-                        using FileStream fStream = File.Open(path, FileMode.Open, FileAccess.Read, FileShare.Read);
+                        using FileStream fStream = 
+                            File.Open(path, FileMode.Open, FileAccess.Read, FileShare.Read);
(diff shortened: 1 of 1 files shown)


src/csharp/Microsoft.Spark.Worker/Processor/BroadcastVariableProcessor.cs
```diff
@@ -54,7 +54,7 @@ internal BroadcastVariables Process(Stream stream)
                     else
                     {
                         string path = SerDe.ReadString(stream);
-                        using FileStream fStream = File.Open(path, FileMode.Open, FileAccess.Read);
+                        using FileStream fStream = File.Open(path, FileMode.Open, FileAccess.Read, FileShare.Read);
```
suhsteve, line 57: Can you break this line apart and get it < 100 char?  There is a style guide [here](https://github.com/dotnet/spark/blob/master/docs/coding-guidelines/csharp-coding-style.md)

bolcman: yes, sorry about that , just pushed the formatted line.

suhsteve:
Hmm, would it be too much to ask for a test to cover this scenario ?

imback82:
Yea it would be non-trivial with the current set up. We can tackle this as a follow up PR if needed.

bolcman:
> ea it would be non-trivial with the current set up. We can tackle this as a follow up PR if needed.

imback82:
@bolcman why is this closed? This PR is not merged yet.

Issue #594, opened 2020-07-20
[FEATURE REQUEST]: ToString should return column expr


samueleresca:
**Is your feature request related to a problem? Please describe.**

At the moment the default `Column.ToString` implementation returns the following:

```
Column("att1").ToString() == "Microsoft.Spark.Sql.Column";
```

The `Column.ToString` method uses the default implementation of the `Object` class. Do you think is sensible to return the column name / column expr every time the `Column.ToString` method is called?  

**Describe the solution you'd like**
The `Column.ToString` should return column expression:
```
Column("att1").ToString() == "att1";
Column("att1 > 1").ToString() == "att1 > 1";
```

**Additional context**

Something similar is present in [apache/spark ](https://github.com/apache/spark/blob/a4ca355af8556e8c5948e492ef70ef0b48416dc4/sql/core/src/main/scala/org/apache/spark/sql/Column.scala#L153)I think

GoEddie:
We started discussing this exact thing here: https://github.com/dotnet/spark/issues/585

GoEddie:
I can't add a label but this is a good first issue to implement, add a method to Column:

`public override string ToString() => (string)_jvmObject.Invoke("toString");`

and a test should do it (i'd probably include a comment to say that it doesn't return the value of the column as that is a pretty common misconception with spark)

samueleresca:
> I can't add a label but this is a good first issue to implement, add a method to Column:
> 
> `public override string ToString() => (string)_jvmObject.Invoke("toString");`
> 
> and a test should do it (i'd probably include a comment to say that it doesn't return the value of the column as that is a pretty common misconception with spark)

The `Invoke` method is not even needed in this case because it would return another `JvmObjectReference` type to be  wrapped into a new `Column` instance. I think we can directly call the `_jvmObject.ToString()` since the `_jvmObject` represents the instance of the column in this case. Will do a couple of tests during the weekend.

samueleresca:
> I can't add a label but this is a good first issue to implement, add a method to Column:
> 
> `public override string ToString() => (string)_jvmObject.Invoke("toString");`
> 
> and a test should do it (i'd probably include a comment to say that it doesn't return the value of the column as that is a pretty common misconception with spark)
 
@GoEddie can you check if the implementation I wrote, does it make sense?

samueleresca:
Closing the issue. Related implementation in #624

Issue #585, opened 2020-07-08
[Implementation Question]: Should we override Equals and ToString?


GoEddie:
Does it make sense to override these in spark objects and pass them back to the JVM (or provide our own implementation)? 

Is there a preference for overriding or not overriding? I did a quick search and could see a few "override ToString" so wasn't sure what to do.

imback82:
Could you elaborate on the use case?

GoEddie:
In spark toString is overridden (not sure if it is always or just sometimes) so doing:

https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/Column.scala#L153

```scala
col("a").toString
```
returns a string "a"

if you do that on a Col in .NET you get:

```csharp
Col("a").ToString()
```

returns a string "Microsoft.Spark.Sql.Column".

It feels like if we forwarded ToString and Equals then we will be more consistent with whatever spark does (or doesn't if it doesn't override them in some cases).

imback82:
Got it. Yea, I think we can forward them to Scala side.

GoEddie:
Thanks, Will do

imback82:
I am curious what the notebook experience will be like since it depends on `ToString`. @suhsteve any concern?

suhsteve:
There may be some unintended side effects by override Equals ?  One example may be if a user is adding/checking these jvm objects against C# dictionary/hashset, any collisions would create jvm calls.

suhsteve:
> I am curious what the notebook experience will be like since it depends on `ToString`. @suhsteve any concern?

I don't see anything too concerning with toString with notebooks atm.

imback82:
I am thinking we should support `Equals` and `ToString` for both C# side and JVM side (C# as the first citizen and a variation of `Equals` and `ToString` for the Scala). What do you guys think?

GoEddie:
I’d say pass toString on, this is another example: https://github.com/dotnet/spark/issues/594

imback82:
Yea, we can start with `ToString`. Should we pass thru for all the wrapper classes or the ones that override `toString` on the Scala side? (I see that we have few classes that override `ToString` to call the Scala side b/c it's overridden on the Scala side).

PR #624, opened 2020-08-12
ToString override for returning the column name

samueleresca:
Related issue #594 . Discussed in #585.

- [X] There's a descriptive title that will make sense to other developers some time from now. 
- [X] There's associated issues. All PR's should have issue(s) associated - unless a trivial self-evident change such as fixing a typo. You can use the format `Fixes #nnnn` in your description to cause GitHub to automatically close the issue(s) when your PR is merged.
- [X] Your change description explains what the change does, why you chose your approach, and anything else that reviewers should know.
- [X] You have included any necessary tests in the same PR.


GoEddie:
Hi @samueleresca yep that is what I did in FeatureBase

src/csharp/Microsoft.Spark/Sql/Column.cs
```diff
@@ -886,6 +886,12 @@ public override bool Equals(object obj)
         /// <returns>Hash code for this object</returns>
         public override int GetHashCode() => _jvmObject.GetHashCode();
 
+        /// <summary>
+        /// Invoke the toString method of the column instance
+        /// </summary>
+        /// <returns>Column name of this column</returns>
+        public override string ToString() => (string) _jvmObject.Invoke("toString");
```
suhsteve, line 893:
```suggestion
        public override string ToString() => (string)_jvmObject.Invoke("toString");
```

Issue #25, opened 2019-01-15
Use C# pattern matching for matching types.


imback82:
There are many places where we are doing the following:
PayloadHelper.GetTypeId():

            if (type == typeof(int))
            {
                return new[] { Convert.ToByte('i') };
            }
            if (type == typeof(long))
            {
                return new[] { Convert.ToByte('g') };
            }
            ..

Use C#'s[ pattern matching](https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching) instead if applicable.

CleverFool77:
Hey, is this issue closed, or are there still references where C# pattern matching is yet to be implemented?

sheddy123:
@wudanzy is this still open? I can work on this

PR #636, opened 2020-08-20
Reduce trials that getting same object type

KimKiHyuk:
https://github.com/dotnet/spark/issues/25

I've found that ```ArrowArrayHelpers.cs``` has type check logics that are redundant

this PR would increase code readability

### DIFF
--- a/src/csharp/Microsoft.Spark/Sql/ArrowArrayHelpers.cs
+++ b/src/csharp/Microsoft.Spark/Sql/ArrowArrayHelpers.cs
@@ -112,55 +112,57 @@ namespace Microsoft.Spark.Sql
         {
             IArrowType arrowType = null;
 
-            if (typeof(T) == typeof(BooleanArray))
+            Type type = typeof(T);
+
+            if (type == typeof(BooleanArray))
             {
                 arrowType = BooleanType.Default;
             }
-            else if (typeof(T) == typeof(Int8Array))
+            else if (type == typeof(Int8Array))
             {
                 arrowType = Int8Type.Default;
             }
-            else if (typeof(T) == typeof(UInt8Array))
+            else if (type == typeof(UInt8Array))
             {
                 arrowType = UInt8Type.Default;
             }
-            else if (typeof(T) == typeof(Int16Array))
+            else if (type == typeof(Int16Array))
             {
                 arrowType = Int16Type.Default;
             }
-            else if (typeof(T) == typeof(UInt16Array))
+            else if (type == typeof(UInt16Array))
             {
                 arrowType = UInt16Type.Default;
             }
-            else if (typeof(T) == typeof(Int32Array))
+            else if (type == typeof(Int32Array))
             {
                 arrowType = Int32Type.Default;
             }
-            else if (typeof(T) == typeof(UInt32Array))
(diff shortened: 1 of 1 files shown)


src/csharp/Microsoft.Spark/Sql/ArrowArrayHelpers.cs
```diff
@@ -112,55 +112,57 @@ private static ArrayData BuildEmptyArrayDataFromArrayType<T>()
         {
             IArrowType arrowType = null;
 
-            if (typeof(T) == typeof(BooleanArray))
+            Type type = typeof(T);
+
```
suhsteve, line 116:
```suggestion
```

KimKiHyuk: resolved

suhsteve:
If we are to follow #25 then we may want to consider the pattern matching.

```csharp
            Type type = typeof(T);
            if (typeof(T) == typeof(BooleanArray))
            {
                arrowType = BooleanType.Default;
            }
            else if (typeof(T) == typeof(Int8Array))
            {
                arrowType = Int8Type.Default;
            }
            else if (typeof(T) == typeof(UInt8Array))
            {
                arrowType = UInt8Type.Default;
            }
```
vs
```csharp
            Type type = typeof(T);
            switch (typeof(T))
            {
                case var _ when type == typeof(BooleanArray):
                    arrowType = BooleanType.Default;
                    break;
                case var _ when type == typeof(Int8Array):
                    arrowType = Int8Type.Default;
                    break;
                case var _ when type == typeof(UInt8Array):
                    arrowType = UInt8Type.Default;
                    break;
            }
```

@imback82 what do you think ?

KimKiHyuk:
> If we are to follow #25 then we may want to consider the pattern matching.
> 
> ```cs
>             IArrowType arrowType = null;
> 
>             Type type = typeof(T);
>             if (type == typeof(BooleanArray))
>             {
>                 arrowType = BooleanType.Default;
>             }
>             else if (type == typeof(Int8Array))
>             {
>                 arrowType = Int8Type.Default;
>             }
>             else if (type == typeof(UInt8Array))
>             {
>                 arrowType = UInt8Type.Default;
>             }
> ```
> 
> vs
> 
> ```cs
>             Type type = typeof(T);
>             IArrowType arrowType = type switch
>             {
>                 var _ when type == typeof(BooleanArray) => BooleanType.Default,
>                 var _ when type == typeof(Int8Array) => Int8Type.Default,
>                 var _ when type == typeof(UInt8Array) => UInt8Type.Default,
>                 _ => null
>             };
> ```
> 
> @imback82 what do you think ?

super clean code! 

i wonder the difference between var _ and _

```cs
            IArrowType arrowType = type switch
            {
                _ when type == typeof(BooleanArray) => BooleanType.Default,
                _ when type == typeof(Int8Array) => Int8Type.Default,
                _ when type == typeof(UInt8Array) => UInt8Type.Default,
                _ => null
            };

```
should i use `var _`  or it's ok just ` _ ` ?

imback82:
The current PR looks fine to me. We have more places to move to using pattern matching, so we can do it as a separate PR if needed.

Issue #633, opened 2020-08-19
[FEATURE REQUEST]: Spark 3.0 Readiness


suhsteve:
# APIs

## SparkSession
* [x] `csharp` public static void SetActiveSession(SparkSession session) (#641)
* [x] `csharp` public static void ClearActiveSession() (#641)
* [x] `csharp` public static SparkSession GetActiveSession() (#641)
* [x] `csharp` public DataFrame ExecuteCommand(string runner, string command, Dictionary<string, string> options) (#647)

## DataFrame
* [x] `csharp` public DataFrame Transform(Func<DataFrame, DataFrame> func) (#688)
* [x] `csharp` public IEnumerable<Row> Tail(int n) (#647)
* [x] `csharp` public void PrintSchema(int level) (#647)
* [x] `csharp` public void Explain(string mode) (#647)
* [x] `csharp` public DataFrame Observe(string name, Column expr, params Column[] exprs) (#647)
* [x] `csharp` WriteTo(string table) (#677)

## DataFrameStatFunctions
* [x] `csharp` public DataFrame SampleBy<T>(Column column, IDictionary<T, double> fractions, long seed) (#647)

## DataFrameWriterV2
* [x] `csharp` public DataFrameWriterV2 Using(string provider) (#677)
* [x] `csharp` public DataFrameWriterV2 Option(string key, string value) (#677)
* [x] `csharp` public DataFrameWriterV2 Option(string key, bool value) (#677)
* [x] `csharp` public DataFrameWriterV2 Option(string key, long value) (#677)
* [x] `csharp` public DataFrameWriterV2 Option(string key, double value) (#677)
* [x] `csharp` public DataFrameWriterV2 Options(Dictionary<string, string> options) (#677)
* [x] `csharp` public DataFrameWriterV2 TableProperty(string property, string value) (#677)
* [x] `csharp` public DataFrameWriterV2 PartitionedBy(Column column, params Column[] columns) (#677)
* [x] `csharp` public void Create() (#677)
* [x] `csharp` public void Replace() (#677)
* [x] `csharp` public void CreateOrReplace() (#677)
* [x] `csharp` public void Append() (#677)
* [x] `csharp` public void Overwrite(Column condition) (#677)
* [x] `csharp` public void OverwritePartitions() (#677)

## RelationalGroupedDataset
* [ ] `scala` ~~def as[K: Encoder, T: Encoder]: KeyValueGroupedDataset[K, T]~~

## Functions
[...]

PR #641, opened 2020-08-25
Full support for multithreaded applications

AFFogarty:
## Design

This PR enables multi-threaded user applications. The PR accomplishes this by creating a 1-to-1 mapping between .NET `Thread` objects and JVM single-thread `ExecutorService` objects. `ExecutorService` is a Java interface that wraps a worker thread with its own task queue.

When a .NET thread calls a Spark function, the `JvmBridge` includes the thread's `ManagedThreadId` in the payload body.  When the request is received on the JVM side, `DotnetBackendHandler` uses the `ManagedThreadId` to map the request onto its corresponding `ExecutorService`.  The `ExecutorService` then executes the function on its worker thread.

This pattern guarantees that Spark functions called by a particular .NET thread will always be executed on the same JVM thread.   In addition, two different .NET threads will always execute Spark functions on two different JVM threads.

An object called `ThreadPool` manages the lifecycle for `ExecutorService` objects on the JVM side. A corresponding class called `JvmThreadPool` manages the threads on the .NET side, and does periodic garbage collection.

## Additional features

This PR also enables `GetActiveSession`, `SetActiveSession`, and `ClearActiveSession` APIs on `SparkSession`.  These APIs were previously unavailable because thread-local sessions were unsupported.

## Testing

This PR adds a new E2E test class `JvmThreadPoolTests` that tests the basic functionality of the thread pool.

Since these changes are on the code path for all Spark API calls, these changes are exercised by all existing E2E tests.

## Limitations

The only limitation is that in Spark, `ActiveSession` gives a guarantee that a child thread will inherit its parent's active session.  My design does not support that.  However, I think that this is not a problem since .NET's thread model doesn't support parent/child relationships between threads.

This PR will address bug #333


src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <summary>
+        /// Construct the JvmThreadPool.
+        /// </summary>
+        /// <param name="jvmBridge">The JvmBridge used to call JVM methods.</param>
+        /// <param name="threadGcInterval">The interval to GC finished threads.</param>
+        public JvmThreadPool(IJvmBridge jvmBridge, TimeSpan threadGcInterval)
+        {
+            _jvmBridge = jvmBridge;
+            _activeThreads = new ConcurrentDictionary<int, Thread>();
+            _activeThreadMonitor = new Thread(delegate ()
```
suhsteve, line 31: does the timer need to be in a new thread?  Will you be able to achieve the same thing if you define the timer outside this thread object ?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,74 @@
...
+                using var timer = new Timer((state) =>
+                {
+                    foreach (Thread thread in _activeThreads.Values)
+                    {
+                        if (!thread.IsAlive)
+                        {
+                            TryRemoveThread(thread.ManagedThreadId);
+                        }
+                    }
+                }, null, 0, (int)threadGcInterval.TotalMilliseconds);
```
suhsteve, line 42:
```suggestion
                }, null, threadGcInterval, threadGcInterval);
```

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,74 @@
...
+            {
+                using var timer = new Timer((state) =>
+                {
+                    foreach (Thread thread in _activeThreads.Values)
+                    {
+                        if (!thread.IsAlive)
+                        {
+                            TryRemoveThread(thread.ManagedThreadId);
+                        }
+                    }
```
suhsteve, line 41:
```suggestion
                    foreach (KeyValuePair<int, Thread> kvp in _activeThreads)
                    {
                        if (!kvp.Value.IsAlive)
                        {
                            TryRemoveThread(kvp.Key);
                        }
                    }
```

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// Try to remove a thread.
+        /// </summary>
+        /// <param name="managedThreadId">The ID of the thread to remove.</param>
+        /// <returns>True if success, false if the thread cannot be found.</returns>
+        public bool TryRemoveThread(int managedThreadId)
+        {
+            if (_activeThreads.TryRemove(managedThreadId, out Thread thread))
+            {
+                _jvmBridge.CallStaticJavaMethod(
+                    "DotnetHandler", "rmThread", thread.ManagedThreadId);
```
suhsteve, line 67:
```suggestion
                    "DotnetHandler", "rmThread", managedThreadId);
```

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -61,10 +61,39 @@ internal SparkSession(JvmObjectReference jvmObject)
...
-        /// thread-local variable, which stores the session variable. Since the Netty server
-        /// that handles the requests is multi-threaded, any thread can invoke these APIs,
-        /// resulting in unexpected behaviors if different threads are used.
+        /// <summary>
+        /// Changes the SparkSession that will be returned in this thread and its children when
+        /// SparkSession.GetOrCreate() is called. This can be used to ensure that a given thread
+        /// receives a SparkSession with an isolated session, instead of the global (first created)
+        /// context.
+        /// </summary>
+        /// <param name="session"></param>
```
suhsteve, line 70: `<param>` section is empty.

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -61,10 +61,39 @@ internal SparkSession(JvmObjectReference jvmObject)
...
+        public static void SetActiveSession(SparkSession session) =>
+            session._jvmObject.Jvm.CallStaticJavaMethod(
+                s_sparkSessionClassName, "setActiveSession", session);
+
+        /// <summary>
+        /// Clears the active SparkSession for current thread. Subsequent calls to getOrCreate will
+        /// return the first created context instead of a thread-local override.
```
suhsteve, line 77:
```suggestion
        /// Clears the active SparkSession for current thread. Subsequent calls to
        /// <see cref="Builder.GetOrCreate"/> will return the first created context
        /// instead of a thread-local override.
```

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -61,10 +61,39 @@ internal SparkSession(JvmObjectReference jvmObject)
...
 
-        /// Note that *ActiveSession() APIs are not exposed because these APIs work with a
-        /// thread-local variable, which stores the session variable. Since the Netty server
-        /// that handles the requests is multi-threaded, any thread can invoke these APIs,
-        /// resulting in unexpected behaviors if different threads are used.
+        /// <summary>
+        /// Changes the SparkSession that will be returned in this thread and its children when
+        /// SparkSession.GetOrCreate() is called. This can be used to ensure that a given thread
+        /// receives a SparkSession with an isolated session, instead of the global (first created)
+        /// context.
```
suhsteve, line 68: I don't think SparkSession.GetOrCreate exists.  Probably referring to the builder. 

```suggestion
        /// Changes the SparkSession that will be returned in this thread and its children when
        /// <see cref="Builder.GetOrCreate"/> is called. This can be used to ensure that a given
        /// thread receives a SparkSession with an isolated session, instead of the global
        /// (first created) context.
```

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
```diff
@@ -0,0 +1,58 @@
...
+   * @param threadId Integer id of the thread.
+   * @param task Function to run on the thread.
+   */
+  def run(threadId: Int, task: () => Unit): Unit = {
+    val executor = getOrCreateExecutor(threadId)
+    val future = executor.submit(new Runnable {
+      override def run(): Unit = task()
+    })
+
+    future.get()
```
suhsteve, line 37: I'm curious how `channelRead0` works.   Is each msg started in a new thread?  If so then I think this is okay.  But if channelRead0 reuses a thread that's currently blocking here then I wonder if we would ever hit a deadlock.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/DotnetBackendHandler.scala
```diff
@@ -57,32 +58,45 @@ class DotnetBackendHandler(server: DotnetBackend)
             val t = readObjectType(dis)
             assert(t == 'c')
             val objToRemove = readString(dis)
-            JVMObjectTracker.remove(objToRemove)
+            ThreadPool.run(threadId, () => JVMObjectTracker.remove(objToRemove))
```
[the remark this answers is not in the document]
AFFogarty: I'm mostly doing this to be consistent. I can try removing the `ThreadPool` invocation here and see what happens in the tests.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
```diff
@@ -0,0 +1,58 @@
...
+   */
+  def deleteThread(threadId: Int): Unit = synchronized {
+    executors.remove(threadId).foreach(_.shutdown)
+  }
+
+  /**
+   * Get the executor if it exists, otherwise create a new one.
+   *
+   * @param id Integer id of the thread.
+   * @return
```
Niharikadutta, line 53: Description missing for return

src/scala/microsoft-spark-2.4.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
```diff
@@ -0,0 +1,58 @@
...
+   */
+  def deleteThread(threadId: Int): Unit = synchronized {
+    executors.remove(threadId).foreach(_.shutdown)
+  }
+
+  /**
+   * Get the executor if it exists, otherwise create a new one.
+   *
+   * @param id Integer id of the thread.
+   * @return
```
Niharikadutta, line 53: Description missing for return

src/scala/microsoft-spark-3.0.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
```diff
@@ -0,0 +1,58 @@
...
+   */
+  def deleteThread(threadId: Int): Unit = synchronized {
+    executors.remove(threadId).foreach(_.shutdown)
+  }
+
+  /**
+   * Get the executor if it exists, otherwise create a new one.
+   *
+   * @param id Integer id of the thread.
+   * @return
```
Niharikadutta, line 53: Description missing for return

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
AFFogarty: Changed.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
AFFogarty: Changed.  It turns out that `Timer` does its own threading.

src/scala/microsoft-spark-2.4.x/src/main/scala/org/apache/spark/api/dotnet/DotnetBackendHandler.scala
```diff
@@ -8,14 +8,14 @@ package org.apache.spark.api.dotnet
 
 import java.io.{ByteArrayInputStream, ByteArrayOutputStream, DataInputStream, DataOutputStream}
 
+import scala.collection.mutable.HashMap
+import scala.language.existentials
```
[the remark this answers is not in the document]
AFFogarty: Fixed.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/DotnetBackendHandler.scala
```diff
@@ -8,14 +8,14 @@ package org.apache.spark.api.dotnet
 
 import java.io.{ByteArrayInputStream, ByteArrayOutputStream, DataInputStream, DataOutputStream}
 
+import scala.collection.mutable.HashMap
+import scala.language.existentials
+
```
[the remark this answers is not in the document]
AFFogarty: Fixed.

src/scala/microsoft-spark-3.0.x/src/main/scala/org/apache/spark/api/dotnet/DotnetBackendHandler.scala
```diff
@@ -8,14 +8,14 @@ package org.apache.spark.api.dotnet
 
 import java.io.{ByteArrayInputStream, ByteArrayOutputStream, DataInputStream, DataOutputStream}
 
+import scala.collection.mutable.HashMap
+import scala.language.existentials
+
```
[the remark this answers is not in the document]
AFFogarty: Fixed.

src/scala/microsoft-spark-2.4.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
AFFogarty: Fixed.

src/scala/microsoft-spark-3.0.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
AFFogarty: Fixed.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/DotnetBackendHandler.scala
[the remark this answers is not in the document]
suhsteve: It may be safer to not run it in the ThreadPool because of the race condition above.  Ie, if the timer kicked in and ran "rmThread" and removed the `ExecutorService` from the ThreadPool map and later the destructor for the dotnet jvmobject was called and ran "rm" after, then it would spin up another `ExecutorService` for the thread that was calling "rm"  (although this threadId here may be the GC thread?)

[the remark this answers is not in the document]
AFFogarty: I took a look at `JVMObjectTracker.remove` and all it does is remove the object from the `HashMap`.  So, I agree we don't need to run this function on the `ThreadPool`.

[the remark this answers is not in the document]
AFFogarty: Reverted this line.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/DotnetBackendHandler.scala
```diff
@@ -65,24 +66,37 @@ class DotnetBackendHandler(server: DotnetBackend)
...
+              logError(s"Removing thread $threadId failed", e)
+              writeInt(dos, -1)
+          }
         case "connectCallback" =>
           assert(readObjectType(dis) == 'c')
           val address = readString(dis)
           assert(readObjectType(dis) == 'i')
           val port = readInt(dis)
-          DotnetBackend.setCallbackClient(address, port);
+          ThreadPool.run(threadId, () => DotnetBackend.setCallbackClient(address, port))
```
suhsteve, line 86: We should be okay not running this part and "closeCallback" in the threadpool.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
AFFogarty: Fixed.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/DotnetBackendHandler.scala
```diff
@@ -57,32 +58,45 @@ class DotnetBackendHandler(server: DotnetBackend)
...
           writeInt(dos, 0)
           writeType(dos, "void")
         case "closeCallback" =>
           logInfo("Requesting to close callback client")
-          DotnetBackend.shutdownCallbackClient()
+          ThreadPool.run(threadId, DotnetBackend.shutdownCallbackClient)
```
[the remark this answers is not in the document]
AFFogarty: Removed `ThreadPool` call.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/DotnetBackendHandler.scala
AFFogarty: Removed `ThreadPool` call.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,86 @@
...
+        private readonly IJvmBridge _jvmBridge;
+        private readonly ConcurrentDictionary<int, Thread> _activeThreads;
+        private readonly Timer _activeThreadMonitor;
+
+        /// <summary>
+        /// Construct the JvmThreadPool.
+        /// </summary>
+        /// <param name="jvmBridge">The JvmBridge used to call JVM methods.</param>
+        /// <param name="threadGcInterval">The interval to GC finished threads.</param>
+        public JvmThreadPool(IJvmBridge jvmBridge, TimeSpan threadGcInterval)
```
Niharikadutta, line 28: nit: Can we name this variable `threadGCInterval` as per [these guidelines](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)?redirectedfrom=MSDN)

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,86 @@
...
+            _activeThreadMonitor.Dispose();
+            GarbageCollectThreads();
+        }
+
+        /// <summary>
+        /// Try to add a thread to the pool.
+        /// </summary>
+        /// <param name="thread">The thread to add.</param>
+        /// <returns>True if success, false if already added.</returns>
+        public bool TryAddThread(Thread thread)
```
Niharikadutta, line 50: Single line function, use `=>`

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        /// <summary>
+        /// Try to remove a thread.
+        /// </summary>
+        /// <param name="managedThreadId">The ID of the thread to remove.</param>
+        /// <returns>True if success, false if the thread cannot be found.</returns>
+        public bool TryRemoveThread(int managedThreadId)
+        {
+            if (_activeThreads.TryRemove(managedThreadId, out Thread thread))
```
Niharikadutta, line 62: nit: Let's use the discard variable `_` here for `thread`?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// the JvmBridge.
+        /// </summary>
+        [Fact]
+        public void TestThreadRm()
+        {
+            // Create a thread and ensure that it is initialized in the JVM ThreadPool.
+            var thread = new Thread(() => _spark.Sql("SELECT TRUE"));
+            thread.Start();
+            thread.Join();
+            _jvmBridge.CallStaticJavaMethod("DotnetHandler", "rmThread", thread.ManagedThreadId);
```
Niharikadutta, line 98: Should we add an `Assert.False()` here to make sure `TryRemoveThread()` fails after calling `rmThread`?

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/DotnetBackendHandler.scala
```diff
@@ -168,13 +182,15 @@ class DotnetBackendHandler(server: DotnetBackend)
...
-        }).mkString(", ")
+        val argsStr = args
+          .map(arg => {
+            if (arg != null) {
+              s"[Type=${arg.getClass.getCanonicalName}, Value: $arg]"
+            } else {
+              "[Value: NULL]"
+            }
+          })
+          .mkString(", ")
```
Niharikadutta, line 193: Was this change made only for aesthetic reasons?

AFFogarty: I'm not sure why, but scalafmt automatically made this change.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolTests.cs
AFFogarty: The JVM method `rmThread` is a no-op if `rmThread` has already been called for that thread on the JVM side.  So calling `rmThread` manually here doesn't actually influence the return value of `TryRemoveThread()`.  For this test, I just wanted to check that `rmThread` returns successfully without throwing exceptions.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolTests.cs
```diff
@@ -0,0 +1,101 @@
...
+                    var activeSession = SparkSession.GetActiveSession();
+                    Assert.NotNull(activeSession);
+                    Assert.Equal(appName, activeSession.Conf().Get("spark.app.name", null));
+                });
+
+                thread.Start();
+                while (thread.IsAlive)
+                {
+                    Thread.Sleep(1000);
+                }
```
suhsteve, line 53: can you just `thread.Join();`?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// Try to remove a thread.
+        /// </summary>
+        /// <param name="managedThreadId">The ID of the thread to remove.</param>
+        /// <returns>True if success, false if the thread cannot be found.</returns>
+        public bool TryRemoveThread(int managedThreadId)
+        {
+            if (_activeThreads.TryRemove(managedThreadId, out _))
+            {
+                _jvmBridge.CallStaticJavaMethod(
+                    "DotnetHandler", "rmThread", managedThreadId);
```
suhsteve, line 63:
```suggestion
                _jvmBridge.CallStaticJavaMethod("DotnetHandler", "rmThread", managedThreadId);
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolTests.cs
AFFogarty: Fixed.

suhsteve: Looks like you removed the thread.Start();

AFFogarty: Oops.  Fixed.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        /// <summary>
+        /// Remove any threads that are no longer active.
+        /// </summary>
+        private void GarbageCollectThreads()
+        {
+            foreach (KeyValuePair<int, Thread> kvp in _activeThreads)
+            {
+                if (!kvp.Value.IsAlive)
```
suhsteve, line 76: According to the property definition `true if this thread has been started and has not terminated normally or aborted; otherwise, false.`      Does this mean that it will be `true` if it started but has terminated abnormally ?   If so, should we also check the `ThreadState` ?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Microsoft.Spark.Interop.Ipc
+{
+    /// <summary>
+    /// This class corresponds to the ThreadPool we maintain on the JVM side. This class keeps
+    /// track of which .NET threads are still alive, and issues an rmThread command if a thread is
+    /// not.
```
imback82, line 15: Can you add a comment that one .NET thread maps to one (always same) JVM thread?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -45,6 +47,8 @@ internal JvmBridge(int portNumber)
 
             _portNumber = portNumber;
             _logger.LogInfo($"JvMBridge port is {portNumber}");
+
+            _jvmThreadPool = new JvmThreadPool(this, TimeSpan.FromMinutes(30));
```
imback82, line 51: Can we make this configurable using env variable? Also, make the default value to be either 5 min or 10 minutes?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <summary>
+        /// Construct the JvmThreadPool.
+        /// </summary>
+        /// <param name="jvmBridge">The JvmBridge used to call JVM methods.</param>
+        /// <param name="threadGCInterval">The interval to GC finished threads.</param>
+        public JvmThreadPool(IJvmBridge jvmBridge, TimeSpan threadGCInterval)
+        {
+            _jvmBridge = jvmBridge;
+            _activeThreads = new ConcurrentDictionary<int, Thread>();
+            _activeThreadMonitor = new Timer(
```
imback82, line 32: Can we create a `Timer` only when `_activeThreads` is >1?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
imback82: Also, comment on the thread-safe methods.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,83 @@
...
+            _activeThreads.TryAdd(thread.ManagedThreadId, thread);
+
+        /// <summary>
+        /// Try to remove a thread.
+        /// </summary>
+        /// <param name="managedThreadId">The ID of the thread to remove.</param>
+        /// <returns>True if success, false if the thread cannot be found.</returns>
+        public bool TryRemoveThread(int managedThreadId)
+        {
+            if (_activeThreads.TryRemove(managedThreadId, out _))
```
imback82, line 60: Can you mention somewhere this `_activeThreads` doesn't have ownership of the threads? (e.g., we don't have to call `Join` on the threads removed)

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <summary>
+        /// Remove any threads that are no longer active.
+        /// </summary>
+        private void GarbageCollectThreads()
+        {
+            foreach (KeyValuePair<int, Thread> kvp in _activeThreads)
+            {
+                if (!kvp.Value.IsAlive)
+                {
+                    TryRemoveThread(kvp.Key);
```
imback82, line 78: I guess removing an element while iterating is fine in this case?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <returns>True if success, false if already added.</returns>
+        public bool TryAddThread(Thread thread) =>
+            _activeThreads.TryAdd(thread.ManagedThreadId, thread);
+
+        /// <summary>
+        /// Try to remove a thread.
+        /// </summary>
+        /// <param name="managedThreadId">The ID of the thread to remove.</param>
+        /// <returns>True if success, false if the thread cannot be found.</returns>
+        public bool TryRemoveThread(int managedThreadId)
```
imback82, line 58: Can we just use `threadId` (same as `BuildPayload`)?

imback82 (commented):
Just checked the C# side.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
imback82: Ah. I guess we can't do `>1` since the entire app can be run on a new thread.

imback82: But I think you should create a timer when `_activeThreads > 0` and destroy it when `_activeThreads == 0`.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
AFFogarty: I found the C++ runtime implementation of `IsAlive`:  https://github.com/dotnet/runtime/blob/master/src/coreclr/src/vm/comsynchronizable.cpp#L94

It looks like it will return false whenever the thread is in state `Thread::TS_ReportDead` or `Thread::TS_Dead` so I think we should be good.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
AFFogarty: Yup, it will be fine.  `_activeThreads.TryRemove` fails gracefully and returns `false` if the item has already been removed.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
AFFogarty: Changed.  I also renamed some things so that it's more clear that this class is basically just a garbage collector.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
AFFogarty: Added some comments explaining the bahaviour.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
AFFogarty: Changed so that the timer is only initialized once a thread is added, and automatically disposed if all threads are removed.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
imback82: I meant whether the iterator will be invalidated or not.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
AFFogarty: Where's the best place to put the config variable?  `SparkEnvironment.cs`?

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Run some code on a particular thread.
+   *
+   * @param threadId Integer id of the thread.
+   * @param task Function to run on the thread.
+   */
+  def run(threadId: Int, task: () => Unit): Unit = {
+    val executor = getOrCreateExecutor(threadId)
+    val future = executor.submit(new Runnable {
+      override def run(): Unit = task()
```
imback82, line 34: Question: if `task()` throws an exception, we get the same logging back to C# side, right?

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
```diff
@@ -0,0 +1,58 @@
...
+
+    future.get()
+  }
+
+  /**
+   * Delete a particular thread.
+   *
+   * @param threadId Integer id of the thread.
+   */
+  def deleteThread(threadId: Int): Unit = synchronized {
```
imback82, line 45: this is synchronizing with `getOrCreateExecutor`, right?

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  /**
+   * Get the executor if it exists, otherwise create a new one.
+   *
+   * @param id Integer id of the thread.
+   * @return The new or existing executor with the given id.
+   */
+  private def getOrCreateExecutor(id: Int): ExecutorService = synchronized {
+    executors.getOrElseUpdate(id, Executors.newSingleThreadExecutor)
```
imback82, line 56: I see the following documentation for `newSingleThreadExecutor`: "Creates an Executor that uses a single worker thread operating off an unbounded queue. (Note however that if this single thread terminates due to a failure during execution prior to shutdown, a new one will take its place if needed to execute subsequent tasks.)"

So, if I do the following on the C# side,

```csharp
try
{
   spark.Read.Parquet("invalidpath"); // should throw on the JVM side.
}
catch (...)
{
  // but I am catching on the C# side
  SparkSession.GetActiveSession(); // Will this be fine?
}
```

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
imback82: You can try this file for now: https://github.com/dotnet/spark/blob/master/src/csharp/Microsoft.Spark/Services/ConfigurationService.cs

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
AFFogarty: If `spark.read.parquet()` throws an exception on the JVM side, the executor will not crash because the executor is actually running `DotNetBackendHandler.handleMethodCall()` which has its own try/catch.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
AFFogarty: Yes, calling `synchronized { ...` is equivalent to calling  `this.synchronized { ...`, so both `synchronized` blocks are scoped to the `ThreadPool` object iself.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
imback82: so there is absolutely no scenario where `newSingleThreadExecutor` will spawn a new thread?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
AFFogarty: Moved the config to that file.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
AFFogarty: Added more documentation and explained that all method are thread-safe.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
AFFogarty: Changed the try/catch to catch `Throwable`, so I think we would only lose the executor if some hard crash caused the thread to die.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
AFFogarty: Yeah, all the logging and exception info is maintained on both sides.  `handleMethodCall()` still writes the exception info into the `DataOutputStream`.

```
Inner Exception 1:
JvmException: org.apache.spark.sql.AnalysisException: Path does not exist: file:/C:/Users/anfog/Source/sparknet/spark/artifacts/bin/Microsoft.Spark.Extensions.Delta/Debug/invalidpath;
	at org.apache.spark.sql.execution.datasources.DataSource$$anonfun$org$apache$spark$sql$execution$datasources$DataSource$$checkAndGlobPathIfNecessary$1.apply(DataSource.scala:558)
	at org.apache.spark.sql.execution.datasources.DataSource$$anonfun$org$apache$spark$sql$execution$datasources$DataSource$$checkAndGlobPathIfNecessary$1.apply(DataSource.scala:545)
	at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:241)
	at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:241)
	at scala.collection.immutable.List.foreach(List.scala:392)
...
```

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPool.cs
AFFogarty: Based on the [source code](https://referencesource.microsoft.com/#mscorlib/system/Collections/Concurrent/ConcurrentDictionary.cs,785) it looks like the iterator will return new elements if they are added to the dictionary while the iteration is in progress.

suhsteve (approved):

suhsteve (dismissed):

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -61,10 +61,40 @@ internal SparkSession(JvmObjectReference jvmObject)
...
-        /// that handles the requests is multi-threaded, any thread can invoke these APIs,
-        /// resulting in unexpected behaviors if different threads are used.
+        /// <summary>
+        /// Changes the SparkSession that will be returned in this thread when 
+        /// <see cref="Builder.GetOrCreate"/> is called. This can be used to ensure that a given
+        /// thread receives a SparkSession with an isolated session, instead of the global
+        /// (first created) context.
+        /// </summary>
+        /// <param name="session">SparkSession object</param>
+        public static void SetActiveSession(SparkSession session) =>
```
Niharikadutta, line 71: Can we add the version attribute `[Since(Versions.V3_0_0)]` for this API?

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -61,10 +61,40 @@ internal SparkSession(JvmObjectReference jvmObject)
...
+            session._jvmObject.Jvm.CallStaticJavaMethod(
+                s_sparkSessionClassName, "setActiveSession", session);
+
+        /// <summary>
+        /// Clears the active SparkSession for current thread. Subsequent calls to
+        /// <see cref="Builder.GetOrCreate"/> will return the first created context
+        /// instead of a thread-local override.
+        /// </summary>
+        public static void ClearActiveSession() =>
```
Niharikadutta, line 80: ditto

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -61,10 +61,40 @@ internal SparkSession(JvmObjectReference jvmObject)
...
+            SparkEnvironment.JvmBridge.CallStaticJavaMethod(
+                s_sparkSessionClassName, "clearActiveSession");
+
+        /// <summary>
+        /// Returns the active SparkSession for the current thread, returned by the builder.
+        /// </summary>
+        /// <returns>Return null, when calling this function on executors</returns>
+        public static SparkSession GetActiveSession()
```
Niharikadutta, line 88: ditto

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
Niharikadutta: Nevermind this comment, got confused by #633 and [Pyspark](https://github.com/apache/spark/blob/v3.0.1/python/pyspark/sql/session.py#L266) but I see this has been exposed by Spark since 2.0.0 so please ignore this comment.

Niharikadutta:
LGTM. Thanks @AFFogarty !

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
```diff
@@ -0,0 +1,110 @@
...
+                var thread = new Thread(() =>
+                {
+                    Assert.Null(SparkSession.GetActiveSession());
+
+                    SparkSession.SetActiveSession(
+                        SparkSession.Builder().AppName(appName).GetOrCreate());
+
+                    // Since we are in the child thread, GetActiveSession() should return the child
+                    // SparkSession.
+                    var activeSession = SparkSession.GetActiveSession();
```
imback82, line 45: nit: `SparkSession activeSession = `

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
```diff
@@ -0,0 +1,110 @@
...
+                    Assert.NotNull(activeSession);
+                    Assert.Equal(appName, activeSession.Conf().Get("spark.app.name", null));
+                });
+
+                thread.Start();
+                thread.Join();
+            }
+
+            for (var i = 0; i < 5; i++)
```
imback82, line 54: `for (int i = 0; i < 5; ++i)`

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            }
+
+            Assert.Null(SparkSession.GetActiveSession());
+        }
+
+        /// <summary>
+        /// Monitor a thread via the JvmThreadPoolGarbageCollector.
+        /// </summary>
+        [Fact]
+        public void TestMonitorThread()
```
imback82, line 66: Looks like you are testing `TryAddThread`.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            var threadPool = new JvmThreadPoolGarbageCollector(_jvmBridge, TimeSpan.FromMinutes(30));
```
imback82, line 68: `Dispose` this? Or put it under `using`.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// the JvmBridge.
+        /// </summary>
+        [Fact]
+        public void TestThreadRm()
+        {
+            // Create a thread and ensure that it is initialized in the JVM ThreadPool.
+            var thread = new Thread(() => _spark.Sql("SELECT TRUE"));
+            thread.Start();
+            thread.Join();
+            _jvmBridge.CallStaticJavaMethod("DotnetHandler", "rmThread", thread.ManagedThreadId);
```
imback82, line 91: If the thread id doesn't exist, it would throw an exception? If so, can you add the comment to the code?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            thread.Join();
+        }
+
+        /// <summary>
+        /// Create a Spark worker thread in the JVM ThreadPool then remove it directly through
+        /// the JvmBridge.
+        /// </summary>
+        [Fact]
+        public void TestThreadRm()
```
imback82, line 85: `TestRmThread`?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <summary>
+        /// Test that the JvmThreadGarbageCollectionInterval configuration defaults to 5 minutes,
+        /// and can be updated correctly by setting the environment variable.
+        /// </summary>
+        [Fact]
+        public void TestIntervalConfiguration()
+        {
+            // Default value is 5 minutes.
+            Assert.Null(Environment.GetEnvironmentVariable("DOTNET_THREAD_GC_INTERVAL"));
+            Assert.Equal(TimeSpan.FromMinutes(5), SparkEnvironment.ConfigurationService.JvmThreadGarbageCollectionInterval);
```
imback82, line 103: nit: can you check the char limit?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -35,6 +36,7 @@ internal sealed class JvmBridge : IJvmBridge
         private readonly ILoggerService _logger =
             LoggerServiceFactory.GetLogger(typeof(JvmBridge));
         private readonly int _portNumber;
+        private readonly JvmThreadPoolGarbageCollector _jvmThreadPool;
```
imback82, line 39: `_jvmThreadPoolGC`

src/csharp/Microsoft.Spark/Services/IConfigurationService.cs
```diff
@@ -2,13 +2,20 @@
...
+
 namespace Microsoft.Spark.Services
 {
     /// <summary>
     /// Helps getting config settings to be used in .NET runtime
     /// </summary>
     internal interface IConfigurationService
     {
+        /// <summary>
+        /// How often to run GC on JVM ThreadPool threads. Default value is 5 minutes.
```
imback82, line 15: No need to mention in the interface.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPoolGarbageCollector.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// In .NET for Apache Spark, we maintain a 1-to-1 mapping between .NET application threads
+    /// and corresponding JVM threads. When a .NET thread calls a Spark API, that call is executed
+    /// by its corresponding JVM thread. This functionality allows for multithreaded applications
+    /// with thread-local variables.
+    /// 
+    /// This class keeps track of the .NET application thread lifecycle. When a .NET application
+    /// thread is no longer alive, this class submits an rmThread command to the JVM backend to 
+    /// dispose of its corresponding JVM thread. All methods are thread-safe.
+    /// </summary>
+    internal class JvmThreadPoolGarbageCollector : IDisposable
```
imback82, line 22: Can we use `GC` instead of `GarbageCollector` throughout? I believe .NET also uses `GC`.

src/csharp/Microsoft.Spark/Services/ConfigurationService.cs
```diff
@@ -33,6 +33,18 @@ internal sealed class ConfigurationService : IConfigurationService
 
         private string _workerPath;
 
+        /// <summary>
+        /// How often to run GC on JVM ThreadPool threads.
```
imback82, line 37: You can mention default value here instead.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
AFFogarty: Added `using`.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark/Services/IConfigurationService.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPoolGarbageCollector.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark/Services/ConfigurationService.cs
AFFogarty: Added.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
AFFogarty: `rmThread` won't fail, it will just be a silent no-op.  Perhaps I could add a `WARN` log on the Jvm side?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
imback82: Yea, I think it will be good to add a log since we are supposed to call it once per thread id. Also, can you make it return boolean so we can do a proper testing?

imback82: Or, I think we can let `rmThread` return boolean and do the logging on the C# side. WDYT?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGCTests.cs
```diff
@@ -0,0 +1,114 @@
...
+                    SparkSession activeSession = SparkSession.GetActiveSession();
+                    Assert.NotNull(activeSession);
+                    Assert.Equal(appName, activeSession.Conf().Get("spark.app.name", null));
+                });
+
+                thread.Start();
+                thread.Join();
+            }
+
+            for (var i = 0; i < 5; ++i)
```
imback82, line 54: nit: `int i`

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGCTests.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <summary>
+        /// Test that the GC interval configuration defaults to 5 minutes, and can be updated
+        /// correctly by setting the environment variable.
+        /// </summary>
+        [Fact]
+        public void TestIntervalConfiguration()
+        {
+            // Default value is 5 minutes.
+            Assert.Null(Environment.GetEnvironmentVariable("DOTNET_THREAD_GC_INTERVAL"));
```
imback82, line 102: `DOTNET_JVM_THREAD_GC_INTERVAL` matches better with `JvmThreadGCInterval`

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPoolGC.cs
```diff
@@ -0,0 +1,136 @@
...
+            return returnValue;
+        }
+
+        /// <summary>
+        /// Try to remove a thread from the pool. If the removal is successful, then the
+        /// corresponding JVM thread will also be disposed.
+        /// </summary>
+        /// <param name="threadId">The ID of the thread to remove.</param>
+        /// <returns>True if success, false if the thread cannot be found.</returns>
+        private bool TryRemoveAndDisposeThread(int threadId)
```
imback82, line 97: nit: How about `TryDisposeJvmThread`? This is doing `TryRemoveDotnetThreadAndDisposeJvmThread` but it's mouthful 😄.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/ThreadPool.scala
imback82: thanks for verifying this!

imback82 (commented):
Few nit/minor comments, but looking good to me!

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPoolGC.cs
AFFogarty: Changed.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGCTests.cs
AFFogarty: Changed.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGCTests.cs
AFFogarty: Fixed.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/JvmThreadPoolGarbageCollectorTests.cs
AFFogarty: Made `rmThread` return bool and added logging on C# side.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPoolGC.cs
```diff
@@ -0,0 +1,152 @@
...
+            {
+                // Dispose of the timer if there are no threads to monitor.
+                if (_activeThreadGCTimer != null && _activeThreads.IsEmpty)
+                {
+                    _activeThreadGCTimer.Dispose();
+                    _activeThreadGCTimer = null;
+                }
+            }
+
+            _loggerService.LogDebug("JVM thread GC complete.");
```
imback82, line 149: Can we remove this log and line 130? (We don't have a good logger service, so this will always be printed). I think line 111 should be good enough.

imback82 (approved):
One nit comment, but LGTM, thanks @AFFogarty!

src/csharp/Microsoft.Spark/Interop/Ipc/JvmThreadPoolGC.cs
AFFogarty: Removed.  Also fixed a bug where that function was always returning false.

imback82 (approved):

PR #650, opened 2020-09-03
xxhash64 - function for Spark 3.0 Readiness #633

GoEddie:
From Spark 3.0 Readiness #633 - this is one of the functions, I wasn't 100% sure on the name xxhash64 - I went with XXHash64 but let me know if that isn't right.


GoEddie:
this has already been done!

PR #348, opened 2019-11-22
Set up build pipeline for Spark-3.0.0

imback82:
This PR sets up CI for Spark-3.0.0.

There are few failing tests that were disabled and will be tracked by #657.

PR #655, opened 2020-09-05
TestSignaturesV2_3_X Tests for Spark-3.0.0

suhsteve:
This PR addresses the TestSignaturesV2_3_X Tests for Spark-3.0.0.

`ToLocalIterator` implementation has changed from Spark 2.4 to 3.0

This is a part of the effort to bring in CI for Spark 3.0: #348

### DIFF
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
@@ -698,7 +698,21 @@ namespace Microsoft.Spark.E2ETest.IpcTests
         [SkipIfSparkVersionIsLessThan(Versions.V3_0_0)]
         public void TestSignaturesV3_X_X()
         {
-            Assert.IsType<Row[]>(_df.ToLocalIterator(true).ToArray());
+            // Validate ToLocalIterator
+            var data = new List<GenericRow>
+            {
+                new GenericRow(new object[] { "Alice", 20}),
+                new GenericRow(new object[] { "Bob", 30})
+            };
+            var schema = new StructType(new List<StructField>()
+            {
+                new StructField("Name", new StringType()),
+                new StructField("Age", new IntegerType())
+            });
+            DataFrame df = _spark.CreateDataFrame(data, schema);
+            IEnumerable<Row> actual = df.ToLocalIterator(true).ToArray();
+            IEnumerable<Row> expected = data.Select(r => new Row(r.Values, schema));
+            Assert.Equal(expected, actual);
         }
     }
 }
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
@@ -691,5 +691,14 @@ namespace Microsoft.Spark.E2ETest.IpcTests
                 Assert.IsType<RelationalGroupedDataset>(df.Pivot(Col("age"), values));
             }
         }
+
+        /// <summary>
+        /// Test signatures for APIs introduced in Spark 3.*
+        /// </summary>
+        [SkipIfSparkVersionIsLessThan(Versions.V3_0_0)]
+        public void TestSignaturesV3_X_X()
+        {
+            Assert.IsType<Row[]>(_df.ToLocalIterator(true).ToArray());
+        }
     }
 }
--- a/src/csharp/Microsoft.Spark/Sql/DataFrame.cs
+++ b/src/csharp/Microsoft.Spark/Sql/DataFrame.cs
@@ -922,7 +922,7 @@ namespace Microsoft.Spark.Sql
         /// "collectToPython").
         /// </summary>
         /// <param name="funcName"></param>
-        /// <returns></returns>
+        /// <returns><see cref="Row"/> objects</returns>
         private IEnumerable<Row> GetRows(string funcName)
         {
             (int port, string secret) = GetConnectionInfo(funcName);
@@ -941,7 +941,7 @@ namespace Microsoft.Spark.Sql
         /// <see cref="ToLocalIterator(bool)"/>
         /// </summary>
         /// <param name="info">The object to extract the connection info from.</param>
-        /// <returns></returns>
+        /// <returns><see cref="Row"/> objects</returns>
         private IEnumerable<Row> GetRowsV3_0_0(object info)
         {
             var infos = (JvmObjectReference[])info;
--- a/src/csharp/Microsoft.Spark/Sql/DataFrame.cs
+++ b/src/csharp/Microsoft.Spark/Sql/DataFrame.cs
@@ -743,15 +743,13 @@ namespace Microsoft.Spark.Sql
         [Since(Versions.V3_0_0)]
         public IEnumerable<Row> ToLocalIterator(bool prefetchPartitions)
         {
-            var info =
-                (JvmObjectReference[])_jvmObject.Invoke("toPythonIterator", prefetchPartitions);
-            var port = (int)info[0].Invoke("intValue");
-            var secret = (string)info[1].Invoke("toString");
-            JvmObjectReference server = info[2];
-
+            (int port, string secret, JvmObjectReference server) =
+                ParseConnectionInfo(
+                    _jvmObject.Invoke("toPythonIterator", prefetchPartitions),
+                    true);
             using ISocketWrapper socket = SocketFactory.CreateSocket();
             socket.Connect(IPAddress.Loopback, port, secret);
-            foreach (Row row in new RowCollector().SynchronousCollect(socket, server))
+            foreach (Row row in new RowCollector().Collect(socket, server))
             {
                 yield return row;
             }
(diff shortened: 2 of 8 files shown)


src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -691,5 +691,14 @@ public void TestSignaturesV2_4_X()
...
             }
         }
+
+        /// <summary>
+        /// Test signatures for APIs introduced in Spark 3.*
+        /// </summary>
+        [SkipIfSparkVersionIsLessThan(Versions.V3_0_0)]
+        public void TestSignaturesV3_X_X()
+        {
+            Assert.IsType<Row[]>(_df.ToLocalIterator(true).ToArray());
```
imback82, line 701: Since we added more logic other than simply calling the function on the JVM side, can you add a correctness test as well?

src/csharp/Microsoft.Spark/Sql/DataFrame.cs
```diff
@@ -722,7 +722,39 @@ public IEnumerable<Row> Collect()
...
+        /// <param name="prefetchPartitions">
+        /// If Spark should pre-fetch the next partition before it is needed.
+        /// </param>
+        /// <returns>Row objects</returns>
+        [Since(Versions.V3_0_0)]
+        public IEnumerable<Row> ToLocalIterator(bool prefetchPartitions)
+        {
+            var info =
+                (JvmObjectReference[])_jvmObject.Invoke("toPythonIterator", prefetchPartitions);
+            var port = (int)info[0].Invoke("intValue");
```
imback82, line 748: Can we extend `ParseConnectionInfo` to handle three args?
For example:
```csharp
private (int, string, JvmObjectReference) ParseConnectionInfo(object info, bool parseServer)
```

src/csharp/Microsoft.Spark/Sql/DataFrame.cs
```diff
@@ -903,7 +935,7 @@ public void CreateOrReplaceGlobalTempView(string viewName)
         /// "collectToPython").
         /// </summary>
         /// <param name="funcName"></param>
-        /// <returns></returns>
+        /// <returns><see cref="Row"/> objects</returns>
```
imback82, line 938: Existing code: can you also fill out `funcName` above?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -37,5 +38,81 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
...
+            private int _readStatus = 1;
+            private IEnumerable<Row> _collectEnumerable = null;
+
+            internal SynchronousRowCollector(ISocketWrapper socket, JvmObjectReference server)
+            {
+                _socket = socket;
+                _server = server;
+            }
+
+            ~SynchronousRowCollector()
```
imback82, line 68: I believe an exception thrown in finalizer is fatal. Should we guard against that?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -37,5 +38,81 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
...
+        /// <param name="socket">Socket the get the stream from</param>
+        /// <param name="server">The jvm socket auth server</param>
+        /// <returns>Collection of row objects</returns>
+        public IEnumerable<Row> SynchronousCollect(ISocketWrapper socket, JvmObjectReference server) =>
+            new SynchronousRowCollector(socket, server);
+
+        /// <summary>
+        /// SynchronousRowCollector synchronously collects Row objects from a socket.
+        /// </summary>
+        private class SynchronousRowCollector : IEnumerable<Row>
```
imback82, line 54: How about `LocalIteratorFromSocket` (from pyspar) or `RowIteratorFromSocket`?

imback82: Also, can you add a comment with a link to pyspark implementation?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -37,5 +38,81 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
...
+            }
+
+            public IEnumerator<Row> GetEnumerator()
+            {
+                Stream inputStream = _socket.InputStream;
+                Stream outputStream = _socket.OutputStream;
+
+                while (_readStatus == 1)
+                {
+                    // Request next partition data from Java
```
imback82, line 92: nit: can you add "." to comments?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -37,5 +38,81 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
...
+                        // Load the partition data from stream and read each item
+                        _collectEnumerable = new RowCollector().Collect(_socket);
+                        foreach (Row row in _collectEnumerable)
+                        {
+                            yield return row;
+                        }
+                    }
+                    else if (_readStatus == -1)
+                    {
+                        // An error occurred, join serving thread and raise any exceptions from the JVM
```
imback82, line 109: `raise any exceptions from the JVM` => does this mean if `getResult` is called, exception would be propagated?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -37,5 +38,81 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
...
+                Stream inputStream = _socket.InputStream;
+                Stream outputStream = _socket.OutputStream;
+
+                while (_readStatus == 1)
+                {
+                    // Request next partition data from Java
+                    SerDe.Write(outputStream, 1);
+                    outputStream.Flush();
+
+                    // If response is 1 then there is a partition to read, if 0 then fully consumed
```
imback82, line 96: then should we add `else` and assert the `_readStatus` is `0` to detect any issues early?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -37,5 +38,81 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
...
+        /// <summary>
+        /// SynchronousRowCollector synchronously collects Row objects from a socket.
+        /// </summary>
+        private class SynchronousRowCollector : IEnumerable<Row>
+        {
+            private readonly ISocketWrapper _socket;
+            private readonly JvmObjectReference _server;
+
+            private int _readStatus = 1;
+            private IEnumerable<Row> _collectEnumerable = null;
```
imback82, line 60: `_currentRows` or `_currentPartitionRows`?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
```diff
@@ -37,5 +38,81 @@ public IEnumerable<Row> Collect(ISocketWrapper socket)
...
             }
         }
+
+        /// <summary>
+        /// Synchronously collects pickled row objects from the given socket.
+        /// </summary>
+        /// <param name="socket">Socket the get the stream from</param>
+        /// <param name="server">The jvm socket auth server</param>
+        /// <returns>Collection of row objects</returns>
+        public IEnumerable<Row> SynchronousCollect(ISocketWrapper socket, JvmObjectReference server) =>
```
imback82, line 48: This can be named as `Collect` as an overload? The existing `Collect` seems to be synchronous as well. And you can add more descriptive comments to summary that this API collects rows by partitions leveraging the other `Collect`, for example.

imback82 (commented):
Few minor comment, but generally looking good to me.

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
suhsteve: The exception will be thrown on the spark side and the exception stack trace will appear in the spark logs.  Do you think we should throw an exception in the dotnet side too ?

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
suhsteve: Added try/catch and ignoring the exceptions.

src/csharp/Microsoft.Spark/Sql/DataFrame.cs
```diff
@@ -922,28 +952,32 @@ private IEnumerable<Row> GetRows(string funcName)
         /// used for connecting with Spark to receive rows for this `DataFrame`.
         /// </summary>
         /// <returns>A tuple of port number and secret string</returns>
```
imback82, line 954: Can you update the comment?

src/csharp/Microsoft.Spark/Sql/DataFrame.cs
```diff
@@ -902,18 +932,18 @@ public void CreateOrReplaceGlobalTempView(string viewName)
...
-        /// <returns></returns>
+        /// <param name="funcName">
+        /// The name of the function to call, either "toPythonIterator" or "collectToPython".
+        /// </param>
+        /// <returns><see cref="Row"/> objects</returns>
         private IEnumerable<Row> GetRows(string funcName)
         {
-            (int port, string secret) = GetConnectionInfo(funcName);
-            using (ISocketWrapper socket = SocketFactory.CreateSocket())
+            (int port, string secret, JvmObjectReference _) = GetConnectionInfo(funcName);
```
imback82, line 941: just `(int port, string secret, _)`?

imback82 (dismissed):
LGTM (except for a couple of nit comments), thanks @suhsteve!

src/csharp/Microsoft.Spark/Sql/RowCollector.cs
imback82: Can you add that the exception stack trace will appear in the driver log to the comment?

imback82 (approved):

PR #656, opened 2020-09-06
Update azure-pipeline.yml to include comments on why E2E forward compatibility test does not run against Spark 2.4.6.

imback82:
It appears that E2E forward compatibility tests are missing against Spark 2.4.6. The reason is that 2.4.5 is the latest version that 0.9.0 was tested against. I am adding a comment to `azure-pipeline.yml` not to forget about this (I knew about this during review, but forgot about it after few months).


suhsteve (approved):
LGTM.

imback82:
This is actually by design. I instead added the comment.

suhsteve (approved):

PR #661, opened 2020-09-09
Return UTF-8 encoded string from JVM => CLR

suhsteve:
We were previously only returning ASCII strings from JVM=>CLR.  This PR supports encoding the string as UTF-8.


src/scala/microsoft-spark-3.0.x/src/main/scala/org/apache/spark/api/dotnet/SerDe.scala
```diff
@@ -320,9 +321,10 @@ object SerDe {
 
   // NOTE: Only works for ASCII right now
```
suhsteve, line 322:
```suggestion
```

imback82 (approved):
LGTM, thanks @suhsteve!

imback82 (approved):

PR #665, opened 2020-09-10
DataFrameReader.Table

GoEddie:
This add the Table method to DataFrameReader so tables written using DataFrameWriter.SaveAsTable can be read without having to use SparkSession.Sql.


src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameReaderTests.cs
```diff
@@ -75,6 +76,11 @@ public void TestSignaturesV2_3_X()
             string textFile = $"{TestEnvironment.ResourceDirectory}people.txt";
             Assert.IsType<DataFrame>(dfr.Text(textFile));
             Assert.IsType<DataFrame>(dfr.Text(textFile, textFile));
+
+            string randomName = $"table_{new Random().Next(100000, 999999).ToString()}";
```
suhsteve, line 80: why does this need to be a random table name ?

src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
```diff
@@ -255,6 +255,15 @@ public DataFrameReader Options(Dictionary<string, string> options)
         /// <param name="paths">Input paths</param>
         /// <returns>DataFrame object</returns>
         public DataFrame Orc(params string[] paths) => LoadSource("orc", paths);
+        
+        /// <summary>
+        /// Returns the specified table as a DataFrame. This is a table that is configured in the
```
suhsteve, line 260:
```suggestion
        /// Returns the specified table as a DataFrame. This is a table that is configured in the
```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameReaderTests.cs
GoEddie: If the files already exist then spark will fail, there used to be a workaround but it was removed with spark 3.0 spark.sql.legacy.allowCreatingManagedTableUsingNonemptyLocation and if the option is set in spark 3.0 it throws an exception.

src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
```diff
@@ -255,6 +255,15 @@ public DataFrameReader Options(Dictionary<string, string> options)
...
+        /// Hive metabase.
```
suhsteve, line 261: Is this always true?  This only works with tables that are configured in hive ?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameReaderTests.cs
suhsteve: `DataFrameReader.Table(..)` looks like an alias for `SparkSession.Table(..)`.  Can we do the same thing we do for that test?

```csharp
            _spark.Range(10).CreateOrReplaceTempView("testView");
            Assert.IsType<DataFrame>(_spark.Table("testView"));
```

Niharikadutta, line 80: Is this `ToString()` needed? I think this can be simplified.

GoEddie: I've copied the test from Spark.Table

GoEddie: I've copied it now

src/csharp/Microsoft.Spark/Sql/DataFrameReader.cs
GoEddie: I've removed it now

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameReaderTests.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
```
suhsteve, line 5: is this still needed ?

GoEddie: gone now!

suhsteve (approved):
LGTM.  Thanks @GoEddie !

Niharikadutta:
LGTM. Thanks @GoEddie !

imback82 (approved):
LGTM, thanks @GoEddie!

PR #678, opened 2020-09-21
Add timeout to Foreachbatch tests

suhsteve:
Add a timeout in case `ProcessAllAvailable` blocks forever.


Niharikadutta:
LGTM

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/sql/api/dotnet/SQLUtils.scala
```diff
@@ -34,4 +36,21 @@ object SQLUtils {
...
+   *                   property of the dotnet type.
+   * @return A typed MemoryStream.
+   */
+  def createMemoryStream(implicit sqlContext: SQLContext, streamType: String): MemoryStream[_] = {
+    import sqlContext.implicits._
+
+    streamType match {
+      case "Int32" => MemoryStream[Int]
+      case "String" => MemoryStream[String]
+      case _ => throw new Exception(s"$streamType not supported")
```
Niharikadutta, line 53: Why are `int` and `string` the only types of `MemoryStream` we are supporting?

suhsteve: It's only for unit testing.  If anyone feels like they need a different type, they can add support themselves.

Niharikadutta: Got it. Also, nit, but I'm wondering if we want to put this function in this file or a separate `TestUtils` file given how it is meant to be used for unit tests only.

imback82: Yea, can we move this out to `TestUtils`?

imback82: Also, we can put it under `org.apache.spark.sql.test`

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -71,44 +71,45 @@ public void TestSignaturesV2_3_X()
...
+                    df.Select(outerUdf(innerUdf(Col("value"))))
                         .Write()
                         .Csv(Path.Combine(dstTempDirectory.Path, id.ToString()));
                 });
 
             StreamingQuery sq = dsw.Start();
 
-            // Process until all available data in the source has been processed and committed
-            // to the ForeachBatch sink. 
-            sq.ProcessAllAvailable();
```
imback82, line 101: Can you update the PR description why this didn't work?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -71,44 +71,45 @@ public void TestSignaturesV2_3_X()
...
+            TimeSpan streamingTimeout = TimeSpan.FromSeconds(30);
+            try
+            {
+                memoryStream.AddData(Enumerable.Range(0, 10).ToArray());
+
+                // Process until all available data in the source has been processed and committed
+                // to the ForeachBatch sink.
+                FailAfter(streamingTimeout, () => sq.ProcessAllAvailable());
```
imback82, line 101: Do we also expect the timeout with this approach? If so, what's the benefit of doing this over the previous approach?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -128,7 +129,7 @@ public void TestForeachBatch()
                 .Sort("id");
 
             IEnumerable<int> actualIds = df.Collect().Select(r => r.GetAs<int>("id"));
-            Assert.True(Enumerable.Range(300, 20).SequenceEqual(actualIds));
+            Assert.True(Enumerable.Range(300, 1010).SequenceEqual(actualIds));
```
imback82, line 132: Can you add a comment on 1010?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
suhsteve: We can determine if it's this test that is hanging.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
suhsteve: I don't know if this was what was failing the tests, but this is something that some spark tests do when using `ProcessAllAvailable`

imback82: Oh then why not just put `FailAfter` with the current approach without `MemoryStream`?

suhsteve: Spark tests also use `MemoryStream`.  If this test is flaky, then hoping this will fix it.

imback82: Unless you are absolutely sure (you mentioned "some" spark tests were using this not all, so I am confused), can we first use `FailAfter` to monitor if this is causing the hang instead of brining in other changes?

imback82: Also, do you think we can try to repro this locally? (e.g., run this in a loop, etc.)

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -96,19 +96,26 @@ public void TestForeachBatch()
...
-            // Add new file to the source path. The spark stream will read any new files
-            // added to the source path.
-            // id column: [10, 11, ..., 19]
-            WriteCsv(10, 10, Path.Combine(srcTempDirectory.Path, "input2.csv"));
-
-            // Process until all available data in the source has been processed and committed
-            // to the ForeachBatch sink.
-            sq.ProcessAllAvailable();
-            sq.Stop();
+            TimeSpan streamingTimeout = TimeSpan.FromSeconds(30);
```
AFFogarty, line 99: Should this value be configurable somehow?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamWriterTests.cs
```diff
@@ -273,6 +282,15 @@ private void WriteCsv(int start, int count, string path)
             }
         }
 
+        private static void FailAfter(TimeSpan timeout, Action func)
```
AFFogarty, line 285: Should this function be private to this class, or treated as a shared utility function?

imback82:
@suhsteve Can you keep retriggering the tests until you see the failures and see if this catches the failures?

suhsteve:
> @suhsteve Can you keep retriggering the tests until you see the failures and see if this catches the failures?

Most recent one has it failing with timeout.

imback82:
Closing this as the mitigation work has been committed (#714).

PR #549, opened 2020-06-17
Expose DataStreamWriter.ForeachBatch API

suhsteve:
This PR exposes the `DataStreamWriter.ForeachBatch` API

#208 

Users can call this API by using an `Action<DataFrame, long>` delegate.  For example

```csharp
spark
    .ReadStream()
    .Schema("id INT")
    .Csv("/path/to/csv")
    .WriteStream()
    .ForeachBatch((df, id) => 
    {
        df.Write().Csv("/unique/write/path"));
    }
```

Logs produced on the dotnet side:
```
[2020-06-21T20:55:22.0427094Z] [Info] [CallbackServer] Starting CallbackServer.
[2020-06-21T20:55:22.0581917Z] [Info] [CallbackServer] Started CallbackServer on 127.0.0.1:61344
[2020-06-21T20:55:26.1706082Z] [Info] [CallbackConnection] [1] Connected with RemoteEndPoint: 127.0.0.1:61350
[2020-06-21T20:55:26.1754979Z] [Info] [CallbackServer] Pool snapshot: [NumThreads:1], [NumConnections:1]
[2020-06-21T20:55:26.1779739Z] [Info] [CallbackConnection] [1] Received request for callback id: 1, callback handler: Microsoft.Spark.Interop.Ipc.ForeachBatchCallbackHandler
[2020-06-21T20:55:28.7991929Z] [Debug] [CallbackConnection] [1] Received END_OF_STREAM signal.
[2020-06-21T20:55:29.9014111Z] [Info] [CallbackConnection] [1] Received request for callback id: 1, callback handler: Microsoft.Spark.Interop.Ipc.ForeachBatchCallbackHandler
[2020-06-21T20:55:30.6256825Z] [Debug] [CallbackConnection] [1] Received END_OF_STREAM signal.
```

Logs produced on the JVM side:
```
20/06/21 13:55:22 INFO DotnetBackendHandler: Connecting to a callback server at 127.0.0.1:61344
20/06/21 13:55:26 INFO CallbackConnection: Calling callback [callback id = 1] ...
20/06/21 13:55:26 INFO CallbackConnection: Signaling END_OF_STREAM.
20/06/21 13:55:28 INFO CallbackConnection: Received END_OF_STREAM signal. Calling callback [callback id = 1] successful.
20/06/21 13:55:29 INFO CallbackConnection: Calling callback [callback id = 1] ...
20/06/21 13:55:29 INFO CallbackConnection: Signaling END_OF_STREAM.
[...]

PR #681, opened 2020-09-24
CallbackClient - add connection back to pool

suhsteve:
Add connection back to connection pool after successfully calling callback.  Was accidently removed in commit https://github.com/dotnet/spark/pull/549/commits/95fe7dbe0e0e8a47c4803845fab81871bc3dc29e


imback82 (approved):
LGTM, thanks @suhsteve!

Issue #683, opened 2020-09-24
[FEATURE REQUEST]: PicklingUdfWrapper cache type information


suhsteve:
Cache parameter type information to determine if converting/casting is necessary.  This will help speed up the call if the Type only needs to be casted.

suhsteve:
Relevant PR #689 merged.

PR #689, opened 2020-09-25
PicklingUdfWrapper cache type information

suhsteve:
Cache parameter type information to determine if converting/casting is necessary. This will help speed up the call if the Type only needs to be casted.

#683


src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
```diff
@@ -274,24 +390,52 @@ internal object Execute(int splitIndex, object[] input, int[] argOffsets)
...
-                TypeConverter.ConvertTo<T9>(input[argOffsets[8]]));
+                (_sameT1 ??= param1 is T1) ? (T1)param1 : TypeConverter.ConvertTo<T1>(param1),
+                (_sameT2 ??= param2 is T2) ? (T2)param2 : TypeConverter.ConvertTo<T2>(param2),
+                (_sameT3 ??= param3 is T3) ? (T3)param3 : TypeConverter.ConvertTo<T3>(param3),
+                (_sameT4 ??= param4 is T4) ? (T4)param4 : TypeConverter.ConvertTo<T4>(param4),
+                (_sameT5 ??= param5 is T5) ? (T5)param5 : TypeConverter.ConvertTo<T5>(param5),
+                (_sameT6 ??= param6 is T6) ? (T6)param6 : TypeConverter.ConvertTo<T6>(param6),
+                (_sameT7 ??= param7 is T7) ? (T7)param7 : TypeConverter.ConvertTo<T7>(param7),
+                (_sameT8 ??= param8 is T8) ? (T8)param8 : TypeConverter.ConvertTo<T8>(param8),
+                (_sameT9 ??= param9 is T9) ? (T9)param9 : TypeConverter.ConvertTo<T9>(param9));
```
Niharikadutta, line 438: I am wondering if this logic can be easily abstracted into a function that is fed the `input` and `argOffsets` and returns the `_func`

src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
```diff
@@ -274,24 +390,52 @@ internal object Execute(int splitIndex, object[] input, int[] argOffsets)
...
+        private bool? _sameT5 = null;
+        [NonSerialized]
+        private bool? _sameT6 = null;
+        [NonSerialized]
+        private bool? _sameT7 = null;
+        [NonSerialized]
+        private bool? _sameT8 = null;
+        [NonSerialized]
+        private bool? _sameT9 = null;
+
```
imback82, line 411: Any reason for not using an array? I guess b/c it's not serialized (requires allocation)? Maybe we can just serialize and use array?

src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
suhsteve: added

src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
suhsteve: changed to array.

src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
```diff
@@ -4,9 +4,16 @@
...
 using System;
 using Microsoft.Spark.Utils;
+using static Microsoft.Spark.Sql.PicklingUdfWrapperHelper;
 
 namespace Microsoft.Spark.Sql
 {
+    internal static class PicklingUdfWrapperHelper
+    {
+        internal static T CastOrConvertTo<T>(ref bool? same, object param) =>
+            (same ??= param is T) ? (T)param : TypeConverter.ConvertTo<T>(param);
```
imback82, line 14: Would this be inlined? The whole purpose of this PR was perf, so can you check if this wouldn't cause regression?

src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
```diff
@@ -35,16 +42,19 @@ internal object Execute(int splitIndex, object[] input, int[] argOffsets)
     [UdfWrapper]
     internal class PicklingUdfWrapper<T, TResult>
     {
+        [NonSerialized]
+        private readonly bool?[] _sameT = new bool?[1];
```
imback82, line 46: No need to create an array?

src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
```diff
@@ -21,7 +28,7 @@ internal PicklingUdfWrapper(Func<TResult> func)
             _func = func;
         }
 
-        internal object Execute(int splitIndex, object[] input, int[] argOffsets)
+        internal object Execute(int _, object[] input, int[] argOffsets)
```
imback82, line 31: Can we either have a comment about this param or revert the change?

src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
suhsteve: Removed fn call.

src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
suhsteve: all values in the array will have default values.  The default value of a `Nullable` is `null`

src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
suhsteve: readded

src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
imback82: Oh I meant that this has only one element, so no need to create an array.

imback82 (commented):
LGTM, thanks @suhsteve!

src/csharp/Microsoft.Spark/Sql/PicklingUdfWrapper.cs
suhsteve: ah got it.  Changed it to use a non array for the 1 param case.

imback82 (approved):

Issue #618, opened 2020-08-11
[FEATURE REQUEST]: Delta Lake & Hyperspace version static analyzer


AFFogarty:
**Is your feature request related to a problem? Please describe.**
As we add support for newer versions of Delta Lake and Hyperspace, we need to track which APIs are available in which versions of those libraries.

For example, some Delta Lake APIs may be available only in Delta Lake 0.7.0+ (targeting Spark 3.0+).  If a .NET for Apache Spark user is running Spark 2.4.x and Delta Lake 0.6.1, then they should be able to see which Delta Lake APIs are actually available to them to use.

**Describe the solution you'd like**
We should add annotations to all APIs in `Microsoft.Spark.Extensions.Delta` and `Microsoft.Spark.Extensions.Hyperspace` detailing the required version of Delta Lake and Hyperspace, respectively.  With those annotations present, we can add static analysis to validate correctness of API usage at compile time.

PR #696, opened 2020-10-02
Add Hyperspace version annotations to `Microsoft.Spark.Extensions.Hyperspace` APIs.

AFFogarty:
This PR adds Hyperspace version annotations to all Hyperspace APIs in `Microsoft.Spark.Extensions.Hyperspace`. 

This PR contributes to #618 .

### DIFF
--- a/src/csharp/Extensions/Microsoft.Spark.Extensions.Hyperspace/Hyperspace.cs
+++ b/src/csharp/Extensions/Microsoft.Spark.Extensions.Hyperspace/Hyperspace.cs
@@ -12,6 +12,7 @@ namespace Microsoft.Spark.Extensions.Hyperspace
     /// <summary>
     /// .Net for Spark binding for Hyperspace index management APIs.
     /// </summary>
+    [HyperspaceSince(HyperspaceVersions.V0_0_1)]
     public class Hyperspace : IJvmObjectReferenceProvider
     {
         private static readonly string s_hyperspaceClassName =
@@ -20,6 +21,7 @@ namespace Microsoft.Spark.Extensions.Hyperspace
         private readonly IJvmBridge _jvmBridge;
         private readonly JvmObjectReference _jvmObject;
 
+        [HyperspaceSince(HyperspaceVersions.V0_0_1)]
         public Hyperspace(SparkSession spark)
         {
             _spark = spark;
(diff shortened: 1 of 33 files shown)


src/csharp/Extensions/Microsoft.Spark.Extensions.Hyperspace/Hyperspace.cs
```diff
@@ -20,6 +21,7 @@ public class Hyperspace : IJvmObjectReferenceProvider
         private readonly IJvmBridge _jvmBridge;
         private readonly JvmObjectReference _jvmObject;
 
+        [HyperspaceSince(HyperspaceVersions.V0_0_1)]
```
Niharikadutta, line 24: nit: One naive question, if the version attribute has been added on the class level, does it need to be mentioned for every class API?

src/csharp/Extensions/Microsoft.Spark.Extensions.Hyperspace/HyperspaceVersions.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Spark.Extensions.Hyperspace
+{
+    internal static class HyperspaceVersions
+    {
+        internal const string V0_0_1 = "0.0.1";
+        internal const string V0_0_2 = "0.0.2";
```
Niharikadutta, line 10: I don't see this version being attributed anywhere, is there a reason we have included it here?

src/csharp/Extensions/Microsoft.Spark.Extensions.Hyperspace/Hyperspace.cs
AFFogarty: Yeah. For example, a later version of Hyperspace could add some new functions to an existing class.

src/csharp/Extensions/Microsoft.Spark.Extensions.Hyperspace/HyperspaceVersions.cs
AFFogarty: I included it just for completeness.  @suhsteve Do you think we should include only the necessary versions?

suhsteve: We can just add new versions as necessary, ie when we need to use it.

AFFogarty: Sounds good, removed `0.0.2`.

suhsteve (approved):
LGTM.  Thanks @AFFogarty

imback82 (approved):
LGTM, thanks @AFFogarty!

Issue #654, opened 2020-09-04
[FEATURE REQUEST]: Support GroupedMapUdf in Spark-3.0.0


suhsteve:
Spark has changed the expected contents of an `Arrow RecordBatch` from 2.4 and 3.0.  Spark 3.0 expects the results of the GroupedMap to be a column of StructType instead of separating each column from the DataFrame.

[SPARK-27163](https://issues.apache.org/jira/browse/SPARK-27163)

### Spark2.4
- https://github.com/apache/spark/blob/807e0a484d1de767d1f02bd8a622da6450bdf940/python/pyspark/serializers.py#L219-L261
- https://github.com/apache/spark/blob/807e0a484d1de767d1f02bd8a622da6450bdf940/sql/core/src/main/scala/org/apache/spark/sql/execution/python/FlatMapGroupsInPandasExec.scala#L140-L148

### Spark3.0
- https://github.com/apache/spark/blob/3fdfce3120f307147244e5eaf46d61419a723d50/python/pyspark/sql/pandas/serializers.py#L135-L193
- https://github.com/apache/spark/blob/3fdfce3120f307147244e5eaf46d61419a723d50/sql/core/src/main/scala/org/apache/spark/sql/execution/python/FlatMapGroupsInPandasExec.scala#L94
- https://github.com/apache/spark/blob/3fdfce3120f307147244e5eaf46d61419a723d50/sql/core/src/main/scala/org/apache/spark/sql/execution/python/PandasGroupUtils.scala#L38-L54

To support these changes we need to use Arrow's `StructArray` and `StructType`.  Unfortunately it is currently unsupported [ARROW-6972](https://issues.apache.org/jira/browse/ARROW-6972)

rapoth:
@eerhardt Would you be able to help with [ARROW-6972](https://issues.apache.org/jira/browse/ARROW-6972)? I'm not sure if anyone from the Arrow community is already working on it. Looks like we cannot support Spark 3.0 in .NET without that functionality.

eerhardt:
> @eerhardt Would you be able to help with ARROW-6972?

I should be able to look into it later this week.

rapoth:
Great, thank you so much in advance!

pgovind:
Just adding a note that I'll be working on this starting tomorrow or Monday :) I'll update this post as I make progress!

rapoth:
I'm closing this issue considering it is done. Thank you all!

PR #711, opened 2020-10-05
Changes to support Arrow 2.0 and Spark 3.0

pgovind:
Fix for https://github.com/dotnet/spark/issues/654

Apache arrow changes are here: https://github.com/apache/arrow/pull/8348

Alright, this is ready to review now!

cc @eerhardt @suhsteve


eerhardt:
> we need a minor change in MDA

Does this mean we will need to ship a new version of MDA? Is it possible to get that minor change in now?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -343,10 +345,12 @@ private static RecordBatch ArrowBasedCountCharacters(RecordBatch records)
 
             int characterCount = 0;
 
+            ArrowBuffer.BitmapBuilder nullBitmap = new ArrowBuffer.BitmapBuilder();
             for (int i = 0; i < nameColumn.Length; ++i)
             {
                 string current = nameColumn.GetString(i);
                 characterCount += current.Length;
+                nullBitmap.Append(current == null ? false : true);
```
eerhardt, line 353: `current` will never be `null`. If it was `current.Length` above would throw.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -20,6 +20,8 @@
 using DataFrame = Microsoft.Spark.Sql.DataFrame;
 using FxDataFrame = Microsoft.Data.Analysis.DataFrame;
 using Int32Type = Apache.Arrow.Types.Int32Type;
+using ArrowStructType = Apache.Arrow.Types.StructType;
+using System.Diagnostics;
```
pgovind, line 24: remove this

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -794,20 +794,33 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
             {
                 FxDataFrame dataFrame = FxDataFrame.FromArrowRecordBatch(input);
                 FxDataFrame resultDataFrame = worker.Func(dataFrame);
+
                 IEnumerable<RecordBatch> recordBatches = resultDataFrame.ToArrowRecordBatches();
 
                 foreach (RecordBatch result in recordBatches)
                 {
-                    stat.NumEntriesProcessed += result.Length;
+                    ArrowBuffer.BitmapBuilder validityBitmapBuilder = new ArrowBuffer.BitmapBuilder();
```
pgovind, line 802: Just testing spark 3.0 CI at the moment. It should pass.

Once it does, we may need to exclude this code for versions < 3.0

src/csharp/Microsoft.Spark.Worker.UnitTest/CommandExecutorTests.cs
```diff
@@ -981,15 +981,18 @@ static ArrowStringDataFrameColumn ConvertStrings(ArrowStringDataFrameColumn stri
             RecordBatch outputBatch = await arrowReader.ReadNextRecordBatchAsync();
 
             Assert.Equal(numRows, outputBatch.Length);
-            Assert.Equal(2, outputBatch.ColumnCount);
+            Assert.Equal(1, outputBatch.ColumnCount);
```
eerhardt, line 984: Does this need to be different on Spark 3.0+ vs. < 3.0?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -737,6 +738,25 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
...
         }
 
+        private RecordBatch WrapArrowRecordBatchColumnsInAStruct(RecordBatch batch)
+        {
+            if (SparkEnvironment.SparkVersion >= new Version(Versions.V3_0_0))
+            {
+                ArrowBuffer.BitmapBuilder validityBitmapBuilder = new ArrowBuffer.BitmapBuilder();
+                for (int i = 0; i < batch.Length; i++)
+                {
+                    validityBitmapBuilder.Append(true);
```
eerhardt, line 748: If all rows are non-null, this bitmap buffer can be empty - `ArrowBuffer.Empty`. That saves the allocations and data.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -355,9 +357,11 @@ private static RecordBatch ArrowBasedCountCharacters(RecordBatch records)
             // Return 1 record, if we were given any. 0, otherwise.
             int returnLength = records.Length > 0 ? 1 : 0;
 
+            ArrowStructType structType = new ArrowStructType(new List<Field> { ageField, new Field("name_CharCount", Int32Type.Default, true) });
+
             return new RecordBatch(
                 new Schema.Builder()
-                    .Field(ageField)
+                    .Field(ageField).Field(new Field("Ret Struct", structType, true))
```
pgovind, line 364: Delete this.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -737,6 +738,25 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
...
+                }
+                ArrowBuffer validityBitmap = validityBitmapBuilder.Build();
+
+                StructType structType = new StructType(batch.Schema.Fields.Select((KeyValuePair<string, Field> pair) => pair.Value).ToList());
```
eerhardt, line 752: Can we not use Linq here? It feels like it isn't worth the extra runtime cost.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -794,20 +815,22 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
...
+                    stat.NumEntriesProcessed += final.Length;
 
                     if (writer == null)
                     {
                         writer =
-                            new ArrowStreamWriter(outputStream, result.Schema, leaveOpen: true, ipcOptions);
+                            new ArrowStreamWriter(outputStream, final.Schema, leaveOpen: true, ipcOptions);
                     }
 
                     // TODO: Remove sync-over-async once WriteRecordBatch exists.
```
eerhardt, line 832: Just an FYI - now that Arrow 2.0.0 shipped, we can remove this sync-over-async call since @suhsteve added the sync overloads for the writer.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -737,6 +738,25 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
             return ExecuteArrowGroupedMapCommand(inputStream, outputStream, commands);
         }
 
+        private RecordBatch WrapArrowRecordBatchColumnsInAStruct(RecordBatch batch)
+        {
+            if (SparkEnvironment.SparkVersion >= new Version(Versions.V3_0_0))
```
eerhardt, line 743: I _think_ the unit tests are failing on `SparkEnvironment.SparkVersion`. Maybe it isn't available during unit tests??

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
pgovind: yup yup, was going to clean that up on the next pass

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
pgovind: Yea I think so too. I just pinged @suhsteve about it. Maybe there's another way to get the version, I'm not sure

pgovind: I commented it out for now to test CI. I'll turn it back on once I hear from @suhsteve or @imback82

suhsteve: Can you use `_version` instead ?

pgovind:
Done, I believe this is ready to review now. We're handling the spark version internally in the Spark worker, so no UDFs have to change.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -737,6 +737,23 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
             return ExecuteArrowGroupedMapCommand(inputStream, outputStream, commands);
         }
 
+        private RecordBatch WrapArrowRecordBatchColumnsInAStruct(RecordBatch batch)
+        {
+            if (_version >= new Version(Versions.V3_0_0))
+            {
+                List<Field> fields = new List<Field>(batch.Schema.Fields.Count);
```
eerhardt, line 744: (nit) why not an array?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -794,20 +812,21 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
...
                     if (writer == null)
                     {
                         writer =
-                            new ArrowStreamWriter(outputStream, result.Schema, leaveOpen: true, ipcOptions);
+                            new ArrowStreamWriter(outputStream, final.Schema, leaveOpen: true, ipcOptions);
                     }
 
-                    // TODO: Remove sync-over-async once WriteRecordBatch exists.
-                    writer.WriteRecordBatchAsync(result).GetAwaiter().GetResult();
+                    writer.WriteRecordBatch(final);
```
eerhardt, line 829: If you are going to fix this one here, can you also fix the TODO above? on line 786

`// TODO: Remove sync-over-async once WriteRecordBatch exists.`

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
```diff
@@ -37,10 +37,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Content Include="..\..\scala\microsoft-spark-*\target\microsoft-spark-*.jar"
-             Link="jars\%(Filename)%(Extension)"
-             Pack="true"
-             PackagePath="jars\%(Filename)%(Extension)" />
+    <Content Include="..\..\scala\microsoft-spark-*\target\microsoft-spark-*.jar" Link="jars\%(Filename)%(Extension)" Pack="true" PackagePath="jars\%(Filename)%(Extension)" />
```
eerhardt, line 40: On one hand, this change is unnecessary and should be reverted. On the other hand, I don't like fighting VS....

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
pgovind: Fixed that and a couple more that I found.

src/csharp/Microsoft.Spark/Microsoft.Spark.csproj
pgovind: Read my mind :)

eerhardt (approved):
Changes look good to me. Thanks for the great work here, @pgovind.

You may want to change the PR from "Draft" now.

pgovind:
> You may want to change the PR from "Draft" now.

I'm not sure I can. I think only the repo owners can do that. @rapoth @suhsteve @imback82

src/csharp/Microsoft.Spark.Worker.UnitTest/CommandExecutorTests.cs
```diff
@@ -874,15 +874,28 @@ Int64Array ConvertInt64s(Int64Array int64s)
...
+            if (sparkVersion < new Version(Versions.V3_0_0))
+            {
+                Assert.Equal(2, outputBatch.ColumnCount);
+                stringArray = (StringArray)outputBatch.Column(0);
+                longArray = (Int64Array)outputBatch.Column(1);
+            }
+            else
+            {
+                Assert.Equal(1, outputBatch.ColumnCount);
+                StructArray structArray = (StructArray)outputBatch.Column(0);
```
suhsteve, line 888:
```suggestion
                var structArray = (StructArray)outputBatch.Column(0);
```

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -737,6 +735,23 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
...
+                {
+                    fields[i] = batch.Schema.GetFieldByIndex(i);
+                }
+                StructType structType = new StructType(fields);
+                StructArray structArray = new StructArray(structType, batch.Length, batch.Arrays.Cast<Apache.Arrow.Array>(), ArrowBuffer.Empty);
+                Schema schema = new Schema.Builder().Field(new Field("Struct", structType, false)).Build();
+                return new RecordBatch(schema, new[] { structArray }, batch.Length);
+            }
+            return batch;
+        }
```
suhsteve, line 753:
```suggestion
            if (_version >= new Version(Versions.V3_0_0))
            {
                var fields = new Field[batch.Schema.Fields.Count];
                for (int i = 0; i < batch.Schema.Fields.Count; ++i)
                {
                    fields[i] = batch.Schema.GetFieldByIndex(i);
                }

                var structType = new StructType(fields);
                var structArray = new StructArray(
                    structType,
                    batch.Length,
                    batch.Arrays.Cast<Apache.Arrow.Array>(),
                    ArrowBuffer.Empty);
                Schema schema = new Schema.Builder().Field(new Field("Struct", structType, false)).Build();
                return new RecordBatch(schema, new[] { structArray }, batch.Length);
            }

            return batch;
        }
```

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -737,6 +735,23 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
             return ExecuteArrowGroupedMapCommand(inputStream, outputStream, commands);
         }
 
+        private RecordBatch WrapArrowRecordBatchColumnsInAStruct(RecordBatch batch)
```
suhsteve, line 738: I think we should either bring the if statement out to the caller or rename this method ?

suhsteve (approved):
LGTM.  Just a few minor comments.  Thanks @pgovind !

suhsteve:
@pgovind would it be possible to add a E2E test that uses chained Arrow/FxDataFrame UDFs ?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
pgovind: Why? I'm not opposed to it. What would you rename the method to? I don't have any strong opinions here

pgovind:
> would it be possible to add a E2E test that uses chained Arrow/FxDataFrame UDFs ?

I think they exist already right? I see chained UDFs being tested in E2ETests/IpcTests/Sql/DataFrameTests.cs?

suhsteve:
> > would it be possible to add a E2E test that uses chained Arrow/FxDataFrame UDFs ?
> 
> I think they exist already right? I see chained UDFs being tested in E2ETests/IpcTests/Sql/DataFrameTests.cs?

Great, I see them now.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
suhsteve: To me the method name seems to imply that the RecordBatch will be wrapped in a Struct.  But that will happen if the version is >= 3.0, otherwise it is not wrapped in a struct.

suhsteve: We can keep the name as is and @imback82 can provide feedback if needed.

imback82: How about `WrapColumnsAsStructIfApplicable`?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -737,6 +735,27 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
...
+        {
+            if (_version >= new Version(Versions.V3_0_0))
+            {
+                var fields = new Field[batch.Schema.Fields.Count];
+                for (int i = 0; i < batch.Schema.Fields.Count; i++)
+                {
+                    fields[i] = batch.Schema.GetFieldByIndex(i);
+                }
+                var structType = new StructType(fields);
```
suhsteve, line 747:
```suggestion
                }

                var structType = new StructType(fields);
```

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -737,6 +735,27 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
...
+                var structArray = new StructArray(
+                    structType,
+                    batch.Length,
+                    batch.Arrays.Cast<Apache.Arrow.Array>(),
+                    ArrowBuffer.Empty);
+                Schema schema = new Schema.Builder().Field(new Field("Struct", structType, false)).Build();
+                return new RecordBatch(schema, new[] { structArray }, batch.Length);
+            }
+            return batch;
```
suhsteve, line 756:
```suggestion
            }

            return batch;
```

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -794,20 +813,21 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
...
                 FxDataFrame dataFrame = FxDataFrame.FromArrowRecordBatch(input);
                 FxDataFrame resultDataFrame = worker.Func(dataFrame);
+
                 IEnumerable<RecordBatch> recordBatches = resultDataFrame.ToArrowRecordBatches();
 
-                foreach (RecordBatch result in recordBatches)
+                foreach (RecordBatch batch in recordBatches)
                 {
-                    stat.NumEntriesProcessed += result.Length;
+                    RecordBatch final = WrapArrowRecordBatchColumnsInAStruct(batch);
```
suhsteve, line 821: the name of this var is `final` and its counterpart in `ExecuteArrowGroupedMapCommand` is named `result`.  Should we name them the same for consistency ?

suhsteve (commented):
some minor nits.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
pgovind: Good catch. Fixed it

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -737,6 +735,29 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
             return ExecuteArrowGroupedMapCommand(inputStream, outputStream, commands);
         }
 
+        private RecordBatch WrapColumnsInStructIfApplicable(RecordBatch batch)
+        {
+            if (_version >= new Version(Versions.V3_0_0))
```
suhsteve, line 740: maybe this instead ?

```suggestion
            if (_version.Major >= 3)
```

pgovind: Ha, that's actually what I had first, until I saw this:  https://github.com/dotnet/spark/blob/1bcd37b501eb08df72259e2770da6cf23e8b946a/src/csharp/Microsoft.Spark.E2ETest/Utils/VersionBasedFacts.cs#L14. 

Are you worried about the `new` ? I don't mind changing it!!

suhsteve: just a very micro optimization.  We can keep, I don't mind either.  👍

suhsteve (approved):

pgovind:
Just so you know, I don't have merge permissions here :) So one of you will need to merge this PR

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -737,6 +735,29 @@ internal ArrowOrDataFrameGroupedMapCommandExecutor(Version version)
...
+            {
+                var fields = new Field[batch.Schema.Fields.Count];
+                for (int i = 0; i < batch.Schema.Fields.Count; i++)
```
imback82, line 743:
```suggestion
                for (int i = 0; i < batch.Schema.Fields.Count; ++i)
```

imback82 (approved):
LGTM, thanks for the great work @pgovind!

PR #714, opened 2020-10-06
JvmBridge/Netty blocking connection deadlock mitigation

### LINKED ISSUE #701: [BUG] Re-enable DataStreamWriterTests.TestForeachBatch E2E test.
`DataStreamWriterTests.TestForeachBatch` is disabled due to the flakiness by #700.

Once the root cause is found, the test can be reenabled. This issue can be used to track the root cause.

cc: @suhsteve

suhsteve:
This PR attempts to limit the number of connections to the JVM backend. Netty is configured to use a set number of threads to process incoming connections. Each new connection is delegated to these threads in a round robin fashion. A deadlock can occur if a new connection is scheduled on a blocked thread.

Fixes #701

### DIFF
--- a/src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
+++ b/src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
@@ -54,17 +54,17 @@ namespace Microsoft.Spark.Interop.Ipc
             _jvmThreadPoolGC = new JvmThreadPoolGC(
                 _logger, this, SparkEnvironment.ConfigurationService.JvmThreadGCInterval);
 
-            int backendThreads = SparkEnvironment.ConfigurationService.GetBackendThreads();
-            int maxSockets = backendThreads;
-            if (backendThreads >= (2 * SocketBufferThreshold))
+            int numBackendThreads = SparkEnvironment.ConfigurationService.GetNumBackendThreads();
+            int maxNumSockets = numBackendThreads;
+            if (numBackendThreads >= (2 * SocketBufferThreshold))
             {
                 // Set the max number of concurrent sockets to be less than the number of
                 // JVM backend threads to allow some buffer.
-                maxSockets -= SocketBufferThreshold;
+                maxNumSockets -= SocketBufferThreshold;
             }
-            _logger.LogInfo($"JVM backend threads is {backendThreads}. JvMBridge max " +
-                $"concurrent sockets is set to {maxSockets}.");
-            _socketSemaphore = new SemaphoreSlim(maxSockets, maxSockets);
+            _logger.LogInfo($"The number of JVM backend thread is set to {numBackendThreads}. " +
+                $"The max number of concurrent sockets in JvmBridge is set to {maxNumSockets}.");
+            _socketSemaphore = new SemaphoreSlim(maxNumSockets, maxNumSockets);
         }
 
         private ISocketWrapper GetConnection()
--- a/src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
+++ b/src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
@@ -31,6 +31,8 @@ namespace Microsoft.Spark.Interop.Ipc
         [ThreadStatic]
         private static MemoryStream s_payloadMemoryStream;
 
+        private readonly SemaphoreSlim _socketSemaphore =
+            new SemaphoreSlim(SparkEnvironment.ConfigurationService.GetBackendThreads());
         private readonly ConcurrentQueue<ISocketWrapper> _sockets =
             new ConcurrentQueue<ISocketWrapper>();
         private readonly ILoggerService _logger =
@@ -54,6 +56,11 @@ namespace Microsoft.Spark.Interop.Ipc
 
         private ISocketWrapper GetConnection()
         {
+            // Limit the number of connections to the JVM backend. Netty is configured
+            // to use a set number of threads to process incoming connections. Each
+            // new connection is delegated to these threads and in a round robin fashion.
+            // A deadlock can occur if a new connection is scheduled on a blocking thread.
+            _socketSemaphore.Wait();
             if (!_sockets.TryDequeue(out ISocketWrapper socket))
             {
                 socket = SocketFactory.CreateSocket();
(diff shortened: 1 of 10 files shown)


src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -241,6 +248,10 @@ private ISocketWrapper GetConnection()
                 socket?.Dispose();
                 throw;
             }
+            finally
+            {
+                _socketSemaphore.Release();
+            }
```
suhsteve, line 254: I think in rare cases we may still hit this deadlock.  Say if max backend threads is 10.  And if we are currently using 10 connections (0 in the _sockets queue) we hit this exception.  The `socket?.Dispose()` will not requeue this socket and we will release the semaphore.  Then in the next thread (assuming the other 9 connections are still busy), a new connection will be made to the backend and this connection may be scheduled on the blocked netty thread.

imback82: > I think in rare cases we may still hit this deadlock.

Can we fix it?

imback82: Or not? If not, can you add the comment to the code?

imback82: Should we reuse the `socket` if it's still "valid"?

imback82:
Can you re-enable the disabled test?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
suhsteve: I think we should just leave a comment in the code and see if netty 5 will prevent this from happening.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -31,6 +31,8 @@ internal sealed class JvmBridge : IJvmBridge
         [ThreadStatic]
         private static MemoryStream s_payloadMemoryStream;
 
+        private readonly SemaphoreSlim _socketSemaphore =
+            new SemaphoreSlim(SparkEnvironment.ConfigurationService.GetBackendThreads());
```
imback82, line 35: This should be less than `DOTNET_NUM_BACKEND_THREADS`, no? I still see the tests hanging.

suhsteve: It was working fine leaving it as the same number in my local test.  I can try something less than though.

suhsteve: Should we do a flat number less than the setting ?  ie, -3 ?  Also should we make sure that this setting has a minimum value set?  Like >= 10 ?  If no minimum value, then what should we do in cases that the value is set to something low, like 1-5 ?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
suhsteve: Requeuing the socket if we received a JvmException from the DotnetBackendHandler.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
imback82: We could do -3 and set the minimum as 5, and let's log the final number.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -54,6 +56,11 @@ internal JvmBridge(int portNumber)
 
         private ISocketWrapper GetConnection()
         {
+            // Limit the number of connections to the JVM backend. Netty is configured
+            // to use a set number of threads to process incoming connections. Each
+            // new connection is delegated to these threads in a round robin fashion.
+            // A deadlock can occur if a new connection is scheduled on a blocked thread.
```
imback82, line 62: let's add that this deadlock happens on the JVM side.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/DotnetBackend.scala
```diff
@@ -30,8 +32,8 @@ class DotnetBackend extends Logging {
   private[this] var bossGroup: EventLoopGroup = _
 
   def init(portNumber: Int): Int = {
-    // need at least 3 threads, use 10 here for safety
-    bossGroup = new NioEventLoopGroup(10)
+    val conf = Option(SparkEnv.get).map(_.conf).getOrElse(new SparkConf())
+    bossGroup = new NioEventLoopGroup(conf.get(DOTNET_NUM_BACKEND_THREADS))
```
imback82, line 36: Can we log this number?

imback82 (commented):
Few minor points, but looks good to me!

imback82:
> We could do -3 and set the minimum as 5, and let's log the final number.

Instead of minimum as 5, let's not apply "-3" if x - 3 < 3. So basically only apply if x > 2*threshold. WDYT?

imback82:
This is just a buffer and if you think it's unnecessary, we can just respect the env var value as long as we log them.

suhsteve:
I think having a buffer would be pretty nice to have just in case.

src/scala/microsoft-spark-2.3.x/src/main/scala/org/apache/spark/api/dotnet/DotnetBackend.scala
```diff
@@ -30,8 +32,10 @@ class DotnetBackend extends Logging {
   private[this] var bossGroup: EventLoopGroup = _
 
   def init(portNumber: Int): Int = {
-    // need at least 3 threads, use 10 here for safety
-    bossGroup = new NioEventLoopGroup(10)
+    val conf = Option(SparkEnv.get).map(_.conf).getOrElse(new SparkConf())
+    val numBackendThreads = conf.get(DOTNET_NUM_BACKEND_THREADS)
+    logInfo(s"Backend threads set to ${DOTNET_NUM_BACKEND_THREADS.key}=$numBackendThreads")
```
imback82, line 37: ```s"The number of DotnetBackend threads is set to $numBackendThreads."``` (`DOTNET_NUM_BACKEND_THREADS` may not be set by the user)

imback82 (approved):
LGTM, thanks @suhsteve!

PR #715, opened 2020-10-06
Get Microsoft.Spark and Microsoft.Spark.Worker assembly version information

### LINKED ISSUE #713: [FEATURE REQUEST]: .NET for Apache Spark Driver/Worker Version method
Expose a method that returns the current Spark .NET assembly version running on the Driver and try to make a `best effort` in determining the version of the Spark .NET Worker version.  It will be `best effort` unless there's a way to ensure that every worker in a cluster will run an action.

suhsteve:
Method added that returns a DataFrame that contains the current `Microsoft.Spark` assembly version running on the `Driver` and tries to make a `best effort` attempt in determining the assembly version of the `Microsoft.Spark.Worker`.

There is no guarantee that a Spark Executor will be run on all the nodes in a cluster. To increase the likelyhood, the spark conf `spark.executor.instances` and the `numPartitions` (parameter to `GetAssemblyInfo(...)` should be a adjusted to a reasonable number relative to the number of nodes in the Spark cluster.

```csharp
spark.GetAssemblyInfo().Show(20, 0);
```
```pwsh
+----------------------+---------------+--------+
|AssemblyName          |AssemblyVersion|HostName|
+----------------------+---------------+--------+
|Microsoft.Spark       |0.12.1.0       |hostname|
|Microsoft.Spark.Worker|0.12.1.0       |hostname|
+----------------------+---------------+--------+
```

Building `Microsoft.Spark.Worker` with an updated version and rerunning the above yields
```pwsh
+----------------------+---------------+--------+
|AssemblyName          |AssemblyVersion|HostName|
+----------------------+---------------+--------+
|Microsoft.Spark       |0.12.1.0       |hostname|
|Microsoft.Spark.Worker|0.13.1.0       |hostname|
+----------------------+---------------+--------+
```

Fixes #713

### DIFF
--- a/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
+++ b/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
@@ -4,11 +4,10 @@
 
 using System;
 using System.Linq;
-using Microsoft.Spark.Experimental.Utils;
 using Microsoft.Spark.Sql;
 using Microsoft.Spark.Sql.Types;
-using static Microsoft.Spark.Sql.Functions;
 using static Microsoft.Spark.Experimental.Utils.AssemblyInfoProvider;
+using static Microsoft.Spark.Sql.Functions;
 
 namespace Microsoft.Spark.Experimental.Sql
 {
--- a/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
+++ b/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Spark.Experimental.Sql
         /// </returns>
         public static DataFrame GetAssemblyInfo(this SparkSession session, int numPartitions = 10)
         {
-            StructType schema = AssemblyInfoProvider.AssemblyInfo.s_schema;
+            StructType schema = AssemblyInfoProvider.Schema();
 
             DataFrame driverAssmeblyInfoDf = session.CreateDataFrame(
                 new GenericRow[] { AssemblyInfoProvider.MicrosoftSparkAssemblyInfo().ToGenericRow() },
--- a/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
+++ b/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
@@ -4,16 +4,17 @@
 
 using System;
 using System.Linq;
+using Microsoft.Spark.Experimental.Utils;
 using Microsoft.Spark.Sql;
 using Microsoft.Spark.Sql.Types;
-using Microsoft.Spark.Utils;
+using static Microsoft.Spark.Sql.Functions;
 
 namespace Microsoft.Spark.Experimental.Sql
 {
     public static class SparkSessionExtensions
     {
         /// <summary>
-        /// Get the <see cref="VersionSensor.VersionInfo"/> for the Microsoft.Spark assembly
+        /// Get the <see cref="VersionSensor.AssemblyInfo"/> for the Microsoft.Spark assembly
         /// running on the Spark Driver and make a "best effort" attempt in determining the
         /// version of Microsoft.Spark.Worker assembly on the Spark Executors.
         /// 
@@ -25,32 +26,32 @@ namespace Microsoft.Spark.Experimental.Sql
         /// <param name="session">The <see cref="SparkSession"/></param>
         /// <param name="numPartitions">Number of partitions</param>
         /// <returns>
-        /// A <see cref="DataFrame"/> containing the <see cref="VersionSensor.VersionInfo"/>
+        /// A <see cref="DataFrame"/> containing the <see cref="VersionSensor.AssemblyInfo"/>
         /// </returns>
         public static DataFrame Version(this SparkSession session, int numPartitions = 10)
         {
-            StructType schema = VersionSensor.VersionInfo.s_schema;
+            StructType schema = VersionSensor.AssemblyInfo.s_schema;
 
-            DataFrame sparkInfoDf = session.CreateDataFrame(
+            DataFrame driverAssmeblyInfoDf = session.CreateDataFrame(
                 new GenericRow[] { VersionSensor.MicrosoftSparkVersion().ToGenericRow() },
                 schema);
 
-            Func<Column, Column> workerInfoUdf = Functions.Udf<int>(
+            Func<Column, Column> executorAssemblyInfoUdf = Udf<int>(
                 i => VersionSensor.MicrosoftSparkWorkerVersion().ToGenericRow(),
                 schema);
             DataFrame df = session.CreateDataFrame(Enumerable.Range(0, 10 * numPartitions));
 
-            string tempColName = "WorkerVersionInfo";
-            DataFrame workerInfoDf = df
+            string tempColName = "ExecutorAssemblyInfo";
+            DataFrame executorAssemblyInfoDf = df
                 .Repartition(numPartitions)
-                .WithColumn(tempColName, workerInfoUdf(df["_1"]))
+                .WithColumn(tempColName, executorAssemblyInfoUdf(df["_1"]))
                 .Select(
-                    schema.Fields.Select(f => Functions.Col($"{tempColName}.{f.Name}")).ToArray());
+                    schema.Fields.Select(f => Col($"{tempColName}.{f.Name}")).ToArray());
 
-            return sparkInfoDf
-                .Union(workerInfoDf)
+            return driverAssmeblyInfoDf
+                .Union(executorAssemblyInfoDf)
                 .DropDuplicates()
-                .Sort(schema.Fields.Select(f => Functions.Col(f.Name)).ToArray());
+                .Sort(schema.Fields.Select(f => Col(f.Name)).ToArray());
         }
     }
 }
--- a/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
+++ b/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
@@ -14,9 +14,10 @@ namespace Microsoft.Spark.Experimental.Sql
     public static class SparkSessionExtensions
     {
         /// <summary>
-        /// Get the <see cref="VersionSensor.AssemblyInfo"/> for the Microsoft.Spark assembly
-        /// running on the Spark Driver and make a "best effort" attempt in determining the
-        /// version of Microsoft.Spark.Worker assembly on the Spark Executors.
+        /// Get the <see cref="AssemblyInfoProvider.AssemblyInfo"/> for the "Microsoft.Spark"
+        /// assembly running on the Spark Driver and make a "best effort" attempt in determining
+        /// the <see cref="AssemblyInfoProvider.AssemblyInfo"/> of "Microsoft.Spark.Worker"
+        /// assembly on the Spark Executors.
         /// 
         /// There is no guarantee that a Spark Executor will be run on all the nodes in
         /// a cluster. To increase the likelyhood, the spark conf `spark.executor.instances`
@@ -26,18 +27,18 @@ namespace Microsoft.Spark.Experimental.Sql
         /// <param name="session">The <see cref="SparkSession"/></param>
         /// <param name="numPartitions">Number of partitions</param>
         /// <returns>
-        /// A <see cref="DataFrame"/> containing the <see cref="VersionSensor.AssemblyInfo"/>
+        /// A <see cref="DataFrame"/> containing the <see cref="AssemblyInfoProvider.AssemblyInfo"/>
         /// </returns>
         public static DataFrame Version(this SparkSession session, int numPartitions = 10)
         {
-            StructType schema = VersionSensor.AssemblyInfo.s_schema;
+            StructType schema = AssemblyInfoProvider.AssemblyInfo.s_schema;
 
             DataFrame driverAssmeblyInfoDf = session.CreateDataFrame(
-                new GenericRow[] { VersionSensor.MicrosoftSparkVersion().ToGenericRow() },
+                new GenericRow[] { AssemblyInfoProvider.MicrosoftSparkAssemblyInfo().ToGenericRow() },
                 schema);
 
             Func<Column, Column> executorAssemblyInfoUdf = Udf<int>(
-                i => VersionSensor.MicrosoftSparkWorkerVersion().ToGenericRow(),
+                i => AssemblyInfoProvider.MicrosoftSparkWorkerAssemblyInfo().ToGenericRow(),
                 schema);
             DataFrame df = session.CreateDataFrame(Enumerable.Range(0, 10 * numPartitions));
 
@@ -45,8 +46,7 @@ namespace Microsoft.Spark.Experimental.Sql
             DataFrame executorAssemblyInfoDf = df
                 .Repartition(numPartitions)
                 .WithColumn(tempColName, executorAssemblyInfoUdf(df["_1"]))
-                .Select(
-                    schema.Fields.Select(f => Col($"{tempColName}.{f.Name}")).ToArray());
+                .Select(schema.Fields.Select(f => Col($"{tempColName}.{f.Name}")).ToArray());
 
             return driverAssmeblyInfoDf
                 .Union(executorAssemblyInfoDf)
--- a/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
+++ b/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Spark.Experimental.Sql
         /// </returns>
         public static DataFrame GetAssemblyInfo(this SparkSession session, int numPartitions = 10)
         {
-            StructType schema = new StructType(new StructField[]
+            var schema = new StructType(new StructField[]
             {
                 new StructField("AssemblyName", new StringType(), isNullable: false),
                 new StructField("AssemblyVersion", new StringType(), isNullable: false),
--- a/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
+++ b/src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Spark.Experimental.Sql
         /// 
         /// There is no guarantee that a Spark Executor will be run on all the nodes in
         /// a cluster. To increase the likelyhood, the spark conf `spark.executor.instances`
-        /// and the <paramref name="numPartitions"/> settings should be a adjusted to a
+        /// and the <paramref name="numPartitions"/> settings should be adjusted to a
         /// reasonable number relative to the number of nodes in the Spark cluster.
         /// </summary>
         /// <param name="session">The <see cref="SparkSession"/></param>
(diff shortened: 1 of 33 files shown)


src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -389,6 +390,39 @@ public void Dispose()
...
+
+            Func<Column, Column> workerInfoUdf =
+                Functions.Udf<int>(
+                    i => VersionSensor.MicrosoftSparkWorkerVersion().ToGenericRow(),
+                    VersionSensor.VersionInfo.s_schema);
+            DataFrame df = CreateDataFrame(Enumerable.Range(0, 1000));
+
+            string tempColName = "WorkerVersionInfo";
+            DataFrame workerInfoTempDf = df
+                .Repartition(1000)
```
suhsteve, line 416: the `1000` is too big for local settings.  Is there some spark setting we can leverage to set this value ?  Or what do you think are safe defaults ?  we can also pass this in as a parameter and have a safe default of 10 or something ?

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -389,6 +390,42 @@ public void Dispose()
         /// </summary>
         public void Stop() => _jvmObject.Invoke("stop");
 
+        /// <summary>
+        /// Get the <see cref="VersionSensor.VersionInfo"/> for the Microsoft.Spark assembly
+        /// running on the Spark Driver and make a "best effort" attempt in determining the version
+        /// of Microsoft.Spark.Worker assembly on the Spark Executors.
+        /// </summary>
```
Niharikadutta, line 397: Missing param type

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -389,6 +390,42 @@ public void Dispose()
...
+        /// <returns>
+        /// A <see cref="DataFrame"/> containing the <see cref="VersionSensor.VersionInfo"/>
+        /// </returns>
+        public DataFrame Version(int numPartitions = 10)
```
Niharikadutta, line 401: Just curious, why have we chosen `numPartitions` as 10 here?

suhsteve: Any number is fine, just some safe default to run for local / small clusters.  Do you have a recommendation on what value we can use ?

Niharikadutta: No 10 is fine, I was just wondering if there was a specific reason.

Niharikadutta: Do we want to add a test covering this since it is a public API?

Niharikadutta: Also I am wondering if asking for `numPartitions` is really necessary for finding the version from the user perspective. Would having it as a constant inside the function pose a problem?

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -389,6 +390,43 @@ public void Dispose()
...
+
+            DataFrame sparkInfoDf =
+                CreateDataFrame(
+                    new GenericRow[] { VersionSensor.MicrosoftSparkVersion().ToGenericRow() },
+                    schema);
+
+            Func<Column, Column> workerInfoUdf =
+                Functions.Udf<int>(
+                    i => VersionSensor.MicrosoftSparkWorkerVersion().ToGenericRow(),
+                    schema);
```
Niharikadutta, line 414:
```suggestion
                    i => VersionSensor.MicrosoftSparkWorkerVersion().ToGenericRow(), schema);
```

src/csharp/Microsoft.Spark/Utils/VersionSensor.cs
```diff
@@ -0,0 +1,72 @@
...
+using System;
+using System.Linq;
+using System.Net;
+using System.Reflection;
+using Microsoft.Spark.Sql;
+using Microsoft.Spark.Sql.Types;
+
+namespace Microsoft.Spark.Utils
+{
+    internal class VersionSensor
```
Niharikadutta, line 14: Can we add a summary for this class?

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
suhsteve: What would be the pros and cons of hardcoding the numbers instead ?

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
suhsteve: This suggestion against the example in [csharp-coding-style](https://github.com/dotnet/spark/blob/master/docs/coding-guidelines/csharp-coding-style.md)

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
```diff
@@ -389,6 +390,41 @@ public void Dispose()
...
+        /// <summary>
+        /// Get the <see cref="VersionSensor.VersionInfo"/> for the Microsoft.Spark assembly
+        /// running on the Spark Driver and make a "best effort" attempt in determining the version
+        /// of Microsoft.Spark.Worker assembly on the Spark Executors.
+        /// </summary>
+        /// <param name="numPartitions">Number of partitions</param>
+        /// <returns>
+        /// A <see cref="DataFrame"/> containing the <see cref="VersionSensor.VersionInfo"/>
+        /// </returns>
+        public DataFrame Version(int numPartitions = 10)
```
suhsteve, line 402: Maybe we should name this `AssemblyVersion` instead just in case Spark ever adds a method called `Version` ?

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
Niharikadutta: Pros I think would be usability, since it might be hard to intuitively understand what number of partitions a user should enter and what it has to do with getting version information. Cons, I guess include not being able to trigger every worker node and if some nodes have a different worker version on them then that would go uncaught. But that could happen even with taking `numPartitions` as an argument so not sure what the benefit there is, unless we document it in more depth so as to advise the user of the importance of estimating a 'correct' `numPartitions` value, maybe explain in detail what `best effort` means etc. Thoughts @imback82 ?

src/csharp/Microsoft.Spark/Sql/SparkSession.cs
imback82, line 402: How about making this an extension method under `Experimental` namespace?

imback82, line 402: How about the assembly version on the driver?

suhsteve: Which assembly version on the driver ?  We are getting the Microsoft.Spark assembly version on the driver and the Microsoft.Spark.Worker version on whichever machines the spark executors are spinning up on.

suhsteve: Keep in same `Microsoft.Spark` project just a different namespace, right ?

suhsteve: Moved to `Microsoft.Spark.Experimental.Sql`

src/csharp/Microsoft.Spark/Utils/VersionSensor.cs
```diff
@@ -0,0 +1,68 @@
...
+                new StructField[]
+                {
+                    new StructField("AssemblyName", new StringType(), isNullable: false),
+                    new StructField("AssemblyVersion", new StringType(), isNullable: false),
+                    new StructField("HostName", new StringType(), isNullable: false)
+                });
+
+            internal string AssemblyName { get; set; }
+            internal string AssemblyVersion { get; set; }
+            internal string HostName { get; set; }
```
suhsteve, line 62: We can add BuildDate if we think it'd be useful/nice to have.

src/csharp/Microsoft.Spark/Utils/VersionSensor.cs
```diff
@@ -0,0 +1,68 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Linq;
+using System.Net;
+using System.Reflection;
+using Microsoft.Spark.Sql;
+using Microsoft.Spark.Sql.Types;
+
+namespace Microsoft.Spark.Utils
```
imback82, line 12: Can you move this under Experimental as well?

src/csharp/Microsoft.Spark/Utils/VersionSensor.cs
imback82: We can iterate on this as we get feedbacks

src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// </returns>
+        public static DataFrame Version(this SparkSession session, int numPartitions = 10)
+        {
+            StructType schema = VersionSensor.VersionInfo.s_schema;
+
+            DataFrame sparkInfoDf = session.CreateDataFrame(
+                new GenericRow[] { VersionSensor.MicrosoftSparkVersion().ToGenericRow() },
+                schema);
+
+            Func<Column, Column> workerInfoUdf = Functions.Udf<int>(
```
imback82, line 38: Can we do `using static Microsoft.Spark.Sql.Functions;`?

src/csharp/Microsoft.Spark/Utils/VersionSensor.cs
```diff
@@ -0,0 +1,68 @@
...
+            AssemblyName asmName = assembly.GetName();
+            return new VersionInfo
+            {
+                AssemblyName = asmName.Name,
+                AssemblyVersion = asmName.Version.ToString(),
+                HostName = Dns.GetHostName()
+            };
+        }
+
+        internal class VersionInfo
```
imback82, line 50: Sounds more like `AssemblyInfo` to me.

src/csharp/Microsoft.Spark/Utils/VersionSensor.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        private static readonly Lazy<VersionInfo> s_microsoftSparkWorkerVersionInfo =
+            new Lazy<VersionInfo>(() => CreateVersionInfo(MicrosoftSparkWorkerAssemblyName));
+
+        internal static VersionInfo MicrosoftSparkVersion() => s_microsoftSparkVersionInfo.Value;
+
+        internal static VersionInfo MicrosoftSparkWorkerVersion() =>
+            s_microsoftSparkWorkerVersionInfo.Value;
+
+        private static VersionInfo CreateVersionInfo(string simpleAssemblyName)
```
imback82, line 34: just `assemblyName`? why simple?

src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <param name="session">The <see cref="SparkSession"/></param>
+        /// <param name="numPartitions">Number of partitions</param>
+        /// <returns>
+        /// A <see cref="DataFrame"/> containing the <see cref="VersionSensor.VersionInfo"/>
+        /// </returns>
+        public static DataFrame Version(this SparkSession session, int numPartitions = 10)
+        {
+            StructType schema = VersionSensor.VersionInfo.s_schema;
+
+            DataFrame sparkInfoDf = session.CreateDataFrame(
```
imback82, line 34: Should we name this `driverAssmeblyInfoDf` and `executorAssemblyInfoDf` to match with the coment?

src/csharp/Microsoft.Spark/Utils/VersionSensor.cs
suhsteve: `AssemblyName` has the `Name` property which is the simple name of the assembly and the `FullName` property which is the full name of the assembly.  But I'll change this var name to what you've suggested.

src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+                i => VersionSensor.MicrosoftSparkWorkerVersion().ToGenericRow(),
+                schema);
+            DataFrame df = session.CreateDataFrame(Enumerable.Range(0, 10 * numPartitions));
+
+            string tempColName = "ExecutorAssemblyInfo";
+            DataFrame executorAssemblyInfoDf = df
+                .Repartition(numPartitions)
+                .WithColumn(tempColName, executorAssemblyInfoUdf(df["_1"]))
+                .Select(
+                    schema.Fields.Select(f => Col($"{tempColName}.{f.Name}")).ToArray());
```
imback82, line 49:
```suggestion
                .Select(schema.Fields.Select(f => Col($"{tempColName}.{f.Name}")).ToArray());
```

src/csharp/Microsoft.Spark/Experimental/Utils/VersionSensor.cs
```diff
@@ -0,0 +1,68 @@
...
+using Microsoft.Spark.Sql;
+using Microsoft.Spark.Sql.Types;
+
+namespace Microsoft.Spark.Experimental.Utils
+{
+    /// <summary>
+    /// Gets the <see cref="AssemblyInfo"/> for the "Microsoft.Spark" and "Microsoft.Spark.Worker"
+    /// assemblies if they exist within the current execution context of this application domain.
+    /// </summary>
+    internal class VersionSensor
```
imback82, line 18: `internal static class`

src/csharp/Microsoft.Spark/Experimental/Utils/VersionSensor.cs
```diff
@@ -0,0 +1,68 @@
...
+        private const string MicrosoftSparkAssemblyName = "Microsoft.Spark";
+        private const string MicrosoftSparkWorkerAssemblyName = "Microsoft.Spark.Worker";
+
+        private static readonly Lazy<AssemblyInfo> s_microsoftSparkVersionInfo =
+            new Lazy<AssemblyInfo>(() => CreateVersionInfo(MicrosoftSparkAssemblyName));
+
+        private static readonly Lazy<AssemblyInfo> s_microsoftSparkWorkerVersionInfo =
+            new Lazy<AssemblyInfo>(() => CreateVersionInfo(MicrosoftSparkWorkerAssemblyName));
+
+        internal static AssemblyInfo MicrosoftSparkVersion() => s_microsoftSparkVersionInfo.Value;
```
imback82, line 29: Should be renamed?

src/csharp/Microsoft.Spark/Experimental/Utils/VersionSensor.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        internal static AssemblyInfo MicrosoftSparkWorkerVersion() =>
+            s_microsoftSparkWorkerVersionInfo.Value;
+
+        private static AssemblyInfo CreateVersionInfo(string assemblyName)
+        {
+            Assembly assembly = AppDomain
+                .CurrentDomain
+                .GetAssemblies()
+                .SingleOrDefault(asm => asm.GetName().Name == assemblyName);
```
imback82, line 39: Should be `Single`?

src/csharp/Microsoft.Spark/Experimental/Utils/VersionSensor.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        private static readonly Lazy<AssemblyInfo> s_microsoftSparkWorkerVersionInfo =
+            new Lazy<AssemblyInfo>(() => CreateVersionInfo(MicrosoftSparkWorkerAssemblyName));
+
+        internal static AssemblyInfo MicrosoftSparkVersion() => s_microsoftSparkVersionInfo.Value;
+
+        internal static AssemblyInfo MicrosoftSparkWorkerVersion() =>
+            s_microsoftSparkWorkerVersionInfo.Value;
+
+        private static AssemblyInfo CreateVersionInfo(string assemblyName)
```
imback82, line 34: rename?

src/csharp/Microsoft.Spark/Experimental/Utils/VersionSensor.cs
imback82: Should be `AssemblyInfoProvider`?

src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
suhsteve: updated

src/csharp/Microsoft.Spark/Experimental/Utils/VersionSensor.cs
suhsteve: sounds good.  changed.

src/csharp/Microsoft.Spark/Experimental/Utils/VersionSensor.cs
suhsteve: renamed, thanks.

src/csharp/Microsoft.Spark/Experimental/Utils/VersionSensor.cs
suhsteve: Good idea.  We can have it throw exception if it doesn't find it.

src/csharp/Microsoft.Spark/Experimental/Utils/VersionSensor.cs
suhsteve: renamed.

imback82 (dismissed):
LGTM, thanks @suhsteve!

src/csharp/Microsoft.Spark/Experimental/Sql/SparkSessionExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// a cluster. To increase the likelyhood, the spark conf `spark.executor.instances`
+        /// and the <paramref name="numPartitions"/> settings should be adjusted to a
+        /// reasonable number relative to the number of nodes in the Spark cluster.
+        /// </summary>
+        /// <param name="session">The <see cref="SparkSession"/></param>
+        /// <param name="numPartitions">Number of partitions</param>
+        /// <returns>
+        /// A <see cref="DataFrame"/> containing the <see cref="AssemblyInfoProvider.AssemblyInfo"/>
+        /// </returns>
+        public static DataFrame Version(this SparkSession session, int numPartitions = 10)
```
imback82, line 32: Oh, I missed this. Should we also rename this?

imback82: maybe `GetAssemblyInfo`?

suhsteve: Updated.

src/csharp/Microsoft.Spark/Experimental/Utils/AssemblyInfoProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                AssemblyName = asmName.Name,
+                AssemblyVersion = asmName.Version.ToString(),
+                HostName = Dns.GetHostName()
+            };
+        }
+
+        internal class AssemblyInfo
+        {
+            internal static readonly StructType s_schema = new StructType(
```
suhsteve, line 52: should we make this lazy ?

imback82: Good point. Can you just move this and `ToGenericRow` to `GetAssemblyInfo`? I will probably move this to non-experimental in my next PR since this class will be more general after moving it out.

suhsteve: moved to `SparkSessionExtensions`

imback82 (approved):

PR #718, opened 2020-10-08
Update version check logic on the worker

imback82:
From 1.0.0, we enforce [semantic versioning](https://semver.org/), meaning the worker will be compatible only within the same major version.

This PR changes the behavior of the version checking logic on the worker, following the semantic versioning.

### DIFF
--- a/src/csharp/Microsoft.Spark/Versions.cs
+++ b/src/csharp/Microsoft.Spark/Versions.cs
@@ -13,10 +13,5 @@ namespace Microsoft.Spark
         internal const string V2_4_0 = "2.4.0";
         internal const string V2_4_2 = "2.4.2";
         internal const string V3_0_0 = "3.0.0";
-
-        // The following is used to check the compatibility of UDFs between
-        // the driver and worker side. This needs to be updated only when there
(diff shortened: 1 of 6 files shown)


src/csharp/Microsoft.Spark/Versions.cs
```diff
@@ -16,7 +16,7 @@ internal static class Versions
 
         // The following is used to check the compatibility of UDFs between
         // the driver and worker side. This needs to be updated only when there
-        // is a breaking change on the UDF contract.
-        internal const string CurrentVersion = "0.9.0";
+        // is a breaking change (major version change) on the UDF contract.
+        internal const string CurrentVersion = "1.0.0";
```
suhsteve, line 20: Is this version going to be tightly coupled with the Microsoft.Spark assembly version?  If so then can we just get the version of Microsoft.Spark loaded in the the current `AppDomain` ?

imback82: Yes, good point. Now that we are going with semantic versioning, we don't need this. I will rework on this after #715 is in.

suhsteve (approved):
LGTM.  Thanks @imback82 !

PR #719, opened 2020-10-08
Split E2E test build pipeline

imback82:
I see occasional timeouts in build due to E2E tests taking longer than one hour if they are run on a SKU that's slow.

This PR splits E2E tests based on Spark versions (2.3, 2.4, 3.0, for example).

I didn't attempt to create a template for this because we cannot reuse it when compatibility tests are brought back. We need to revisit creating a template with compatibility tests in mind.


imback82 (commented):
The build now finishes in ~25 mins. Merging to master!

suhsteve:
I wonder if we can have each version test be its own stage and have all version e2e tests run in parallel.

imback82:
I think it would be an overkill since it will acquire a pool per test. You can try it.

PR #724, opened 2020-10-09
Parallelize all E2E tests in build pipeline

suhsteve:
This PR runs all the E2E tests in parallel.


imback82:
Can we wait on this until we introduce the compatibility tests? Do you plan to have one pool per test for compatibility tests as well?

suhsteve:
> Can we wait on this until we introduce the compatibility tests? Do you plan to have one pool per test for compatibility tests as well?

We can either have separate stages for all the forward/backward similar to what we're doing with the e2e one (all e2e running in parallel).  Or we can add the forward/backward to the template, so it'll do e2e, forward, backward in one stage?

imback82:
OK sounds good, forward compat may not always work if new Spark is released, but I guess we can come back to this later. Merging to master!

imback82 (approved):
LGTM, thanks @suhsteve!

PR #728, opened 2020-10-09
Update microsoft-spark JARs to include Scala version

### LINKED ISSUE #293: [FEATURE REQUEST]: Show Scala version in microsoft-spark*.jar filename
Currently, scala version is not embedded in the microsoft-spark JAR name. This will be useful for the user when Spark 3.0 is released which uses 2.12 scala by default.

imback82:
Fixes #293 

This PR proposes to include Scala version in the JAR name (in preparation for publishing these to the central repository) following the similar approach taken by Spark: https://github.com/apache/spark/blob/master/external/kafka-0-10/pom.xml


suhsteve (approved):
LGTM.  Thanks @imback82 !

PR #735, opened 2020-10-11
Shutdown the JVM thread pool to fix the hanging issue while exiting.

imback82:
The JVM thread is not properly cleaned up, so there could be executor services still running when the JVM process tries to exit.

If there are executors running, this will prevent the JVM process to exit (behavior only under Ubuntu), leading to a hanging behavior at the end of the run.

This PR proposes to fix this issue by shutting down the thread pool.

### DIFF
--- a/src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
+++ b/src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
@@ -16,7 +16,7 @@ import java.util.concurrent.{Semaphore, TimeUnit}
 import org.apache.commons.io.FilenameUtils
 import org.apache.hadoop.fs.Path
 import org.apache.spark
-import org.apache.spark.api.dotnet.{DotnetBackend, ThreadPool}
+import org.apache.spark.api.dotnet.DotnetBackend
 import org.apache.spark.deploy.{PythonRunner, SparkHadoopUtil}
 import org.apache.spark.internal.Logging
 import org.apache.spark.util.dotnet.{Utils => DotnetUtils}
@@ -143,11 +143,6 @@ object DotnetRunner extends Logging {
         } else {
           logInfo(s".NET application exited successfully")
         }
-
-        // Shutdown the thread pool whose executors could still be running,
-        // blocking the JVM process to exit. Since .NET process has already exited,
-        // the thread pool will never be cleaned up by the .NET process.
-        ThreadPool.shutdown()
 
         // TODO: The following is causing the following error:
         // INFO ApplicationMaster: Final app status: FAILED, exitCode: 16,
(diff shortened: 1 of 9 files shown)


src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
```diff
@@ -187,11 +193,10 @@ object DotnetRunner extends Logging {
...
         .find(path => Files.isRegularFile(path) && path.getFileName.toString == dotnetExecutable) match {
-          case Some(path) => path.toAbsolutePath.toString
-          case None =>
-            throw new IllegalArgumentException(
-              s"Failed to find $dotnetExecutable under" +
-                s" ${dir.getAbsolutePath}")
+        case Some(path) => path.toAbsolutePath.toString
+        case None =>
+          throw new IllegalArgumentException(
+            s"Failed to find $dotnetExecutable under ${dir.getAbsolutePath}")
```
imback82, line 199: scalafmt changes

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
```diff
@@ -255,7 +260,8 @@ object DotnetRunner extends Logging {
       returnCode = dotnetProcess.waitFor()
     } catch {
       case _: InterruptedException =>
-        logInfo("Thread interrupted while waiting for graceful close. Forcefully closing .NET process")
+        logInfo(
+          "Thread interrupted while waiting for graceful close. Forcefully closing .NET process")
```
imback82, line 264: scalafmt changes

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
```diff
@@ -143,6 +143,12 @@ object DotnetRunner extends Logging {
         } else {
           logInfo(s".NET application exited successfully")
         }
+
+        // Shutdown the thread pool whose executors could still be running,
```
suhsteve, line 147: should this be moved in DotnetBackend.close()` ?

imback82: That works with the current code, but the level is a bit different. If we allow multiple `DotnetBackend` objects, we shouldn't shutdown the ThreadPool (singleton), and that's why I put it outside.

But, I think it's already broken with `DotnetBackend.shutdownCallbackClient()`, so I will move it so that we can fix it at the same time if needed.

src/scala/microsoft-spark-3-0/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
```diff
@@ -187,11 +187,10 @@ object DotnetRunner extends Logging {
         .iterator()
         .asScala
         .find(path => Files.isRegularFile(path) && path.getFileName.toString == dotnetExecutable) match {
-          case Some(path) => path.toAbsolutePath.toString
```
imback82, line 190: To reviewers. DotnetRunner.scala contains only the format changes.

suhsteve (approved):
LGTM

PR #736, opened 2020-10-12
Isolate all dotnet-interactive statements to use the same jvm thread

suhsteve:
Each code submission processed by `dotnet-interactive` run on an available thread in its thread pool.  This means that if two separate code submissions were submitted, each may run on a different thread in the CLR, which in turn be processed on a different thread in the JVM.

This PR prevents this from occurring by forcing the same JVM thread to process all incoming requests from the CLR when running a `dotnet-interactive` repl session.

### DIFF
--- a/src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive/AssemblyKernelExtension.cs
+++ b/src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive/AssemblyKernelExtension.cs
@@ -103,7 +103,7 @@ namespace Microsoft.Spark.Extensions.DotNet.Interactive
 
         private bool TryGetSparkSession(out SparkSession sparkSession)
         {
-            sparkSession = SparkSession.GetDefaultSession();
+            sparkSession = SparkSession.GetActiveSession();
             return sparkSession != null;
         }
 
--- a/src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
+++ b/src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
@@ -11,7 +11,6 @@ using System.Text;
 using System.Threading;
 using Microsoft.Spark.Network;
 using Microsoft.Spark.Services;
-using Microsoft.Spark.Utils;
 
 namespace Microsoft.Spark.Interop.Ipc
 {
@@ -33,6 +32,7 @@ namespace Microsoft.Spark.Interop.Ipc
         private static MemoryStream s_payloadMemoryStream;
 
         private const int SocketBufferThreshold = 3;
+        private const int ThreadIdForRepl = 1;
 
         private readonly SemaphoreSlim _socketSemaphore;
         private readonly ConcurrentQueue<ISocketWrapper> _sockets =
@@ -208,7 +208,7 @@ namespace Microsoft.Spark.Interop.Ipc
                 PayloadHelper.BuildPayload(
                     payloadMemoryStream,
                     isStatic,
-                    thread == null ? 1 : thread.ManagedThreadId,
+                    thread == null ? ThreadIdForRepl : thread.ManagedThreadId,
                     classNameOrJvmObjectReference,
                     methodName,
                     args);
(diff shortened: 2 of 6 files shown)


src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive/AssemblyKernelExtension.cs
```diff
@@ -103,7 +103,7 @@ private bool TryEmitAssembly(CSharpKernel kernel, string dstPath, out string ass
 
         private bool TryGetSparkSession(out SparkSession sparkSession)
         {
-            sparkSession = SparkSession.GetDefaultSession();
+            sparkSession = SparkSession.GetActiveSession();
```
suhsteve, line 106: Should we update this to `GetActiveSession` or keep it as `GetDefaultSession` ?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -184,13 +189,13 @@ private ISocketWrapper GetConnection()
...
             {
-                Thread thread = Thread.CurrentThread;
+                Thread thread = _isRunningRepl ? null : Thread.CurrentThread;
                 MemoryStream payloadMemoryStream = s_payloadMemoryStream ??= new MemoryStream();
                 payloadMemoryStream.Position = 0;
                 PayloadHelper.BuildPayload(
                     payloadMemoryStream,
                     isStatic,
-                    thread.ManagedThreadId,
+                    thread == null ? 1 : thread.ManagedThreadId,
```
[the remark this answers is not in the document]
suhsteve: I think we should just not support multiple threads in repl scenarios.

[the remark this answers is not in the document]
suhsteve: Although I think we may be in trouble if we're using `ForeachBatch` in the repl session.

src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive/AssemblyKernelExtension.cs
imback82: Is this called only during the load? Then it can be either one, I think.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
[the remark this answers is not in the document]
imback82: So, `ForeachBatch` does not work in repl now with this change?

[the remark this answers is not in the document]
suhsteve: If on the JVM side, everything will be scheduled on one ExecutorService, then I don't think it'll work.  Let's say you set up the `StreamingQuery` and you call either `AwaitTermination` or `ProcessAllAvailable`.  This will be scheduled on the `ExecutorService` and will be blocking.  Then a spark thread will spin up and call the `ForeachBatch` callback.  On the dotnet side, the `ForeachBatch` callback will run on a new thread, but when the dataframe methods are called, it will be scheduled on the same `ExecutorService`  that is currently blocked by `AwaitTermination`/`ProcessAllAvailable` ?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -54,6 +56,9 @@ internal JvmBridge(int portNumber)
             _jvmThreadPoolGC = new JvmThreadPoolGC(
                 _logger, this, SparkEnvironment.ConfigurationService.JvmThreadGCInterval);
 
+            _isRunningRepl =
+                EnvironmentUtils.GetEnvironmentVariableAsBool(Constants.RunningREPLEnvVar);
```
imback82, line 60: Can this be moved to `ConfigurationService`?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -184,13 +189,13 @@ private ISocketWrapper GetConnection()
             ISocketWrapper socket = null;
             try
             {
-                Thread thread = Thread.CurrentThread;
+                Thread thread = _isRunningRepl ? null : Thread.CurrentThread;
```
imback82, line 192: Can you add a comment why we are doing this?

imback82 (commented):
Few minor comments, but PR itself looks fine to me.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
imback82: Along with limitations? (foreach batch, etc.)

imback82 (approved):
LGTM, thanks @suhsteve!

PR #737, opened 2020-10-12
Run E2E tests on Linux in build pipeline and add Backward/Forward E2E tests

suhsteve:
This PR runs the E2E tests on the Linux and adds Backward/Forward E2E tests

Skip 2.3.0 E2E tests running on Linux due to issue #753


azure-pipelines.yml
```diff
@@ -11,8 +11,6 @@ variables:
   _SignType: real
   _TeamName: DotNetSpark
   MSBUILDSINGLELOADCONTEXT: 1
-  ArtifactPath: '$(Build.ArtifactStagingDirectory)\Microsoft.Spark.Binaries'
```
imback82, line 14: I think this is used for internal build path?

suhsteve: good catch.  readded

suhsteve:
@Niharikadutta  can you look why the 2.3.0 linux test is hanging?  Looking at the other tests, ie 2.3.1/2.3.2 the next test that should be running is a test that throws that contains  `Caused by: java.io.IOException: org.apache.spark.SparkException: Attempted to use Broadcast(86) after it was destroyed (destroy at NativeMethodAccessorImpl.java:0) `

Niharikadutta:
> @Niharikadutta can you look why the 2.3.0 linux test is hanging? Looking at the other tests, ie 2.3.1/2.3.2 the next test that should be running is a test that contains an exception messages related to broadcast `Caused by: java.io.IOException: org.apache.spark.SparkException: Attempted to use Broadcast(86) after it was destroyed (destroy at NativeMethodAccessorImpl.java:0) `

Sure I'll look into it

Niharikadutta:
I see this warning after the test file has been triggered:
`[Warn] [JvmBridge] rmThread returned false for JVM thread 6. Either thread does not exist or has already been GC'd.
`
Is this benign?

Niharikadutta:
I removed all Broadcast tests and saw the E2E tests for Spark 2.3.0 on Ubuntu still hanging as per #748 , so I don't think it is hanging because of Broadcast. Do we have any other ideas?

imback82:
Can you help finding the offending test suite? Thanks!

AFFogarty:
> I see a bunch of these warnings after the test file has been triggered:
> `[Warn] [JvmBridge] rmThread returned false for JVM thread 6. Either thread does not exist or has already been GC'd. `
> Is this benign? cc @AFFogarty

@Niharikadutta can we run the tests with Debug logging to verify that the thread GC is happening?

suhsteve:
@imback82 , @Niharikadutta , @AFFogarty Skipping 2.3.0 E2E tests on Linux due to issue #753

Niharikadutta (commented):
LGTM. Thanks @suhsteve !

imback82:
Let me check #753 first and see if it is something that can be fixed or not.

suhsteve:
@elvaliuliuliu  can you review?

elvaliuliuliu:
LGTM, thanks @suhsteve!

azure-pipelines.yml
```diff
@@ -12,13 +12,43 @@ variables:
...
+  backwardCompatibleTestOptions_Linux_2_3: ""
+  forwardCompatibleTestOptions_Linux_2_3: ""
+
+  backwardCompatibleTestOptions_Windows_2_4: ""
+  forwardCompatibleTestOptions_Windows_2_4: ""
+  backwardCompatibleTestOptions_Linux_2_4: ""
+  forwardCompatibleTestOptions_Linux_2_4: "--filter \
+  (FullyQualifiedName!=Microsoft.Spark.Extensions.Hyperspace.E2ETest.HyperspaceTests.TestExplainAPI)&\
+  (FullyQualifiedName!=Microsoft.Spark.Extensions.Hyperspace.E2ETest.HyperspaceTests.TestIndexCreateAndDelete)&\
+  (FullyQualifiedName!=Microsoft.Spark.Extensions.Hyperspace.E2ETest.HyperspaceTests.TestSignatures)"
```
imback82, line 30: Do you know why these are filtered? I see no changes from 1.0.0?

azure-pipelines.yml
```diff
@@ -173,18 +203,142 @@ stages:
...
-    - '3.0.1'
+    backwardCompatibleRelease: $(backwardCompatibleRelease)
+    forwardCompatibleRelease: $(forwardCompatibleRelease)
+    tests:
+    - version: '2.3.0'
+      jobOptions:
+      - pool: 'Hosted VS2017'
+        testOptions: ''
+        backwardCompatibleTestOptions: $(backwardCompatibleTestOptions_Windows_2_3)
+        forwardCompatibleTestOptions: $(forwardCompatibleTestOptions_Windows_2_3)
```
imback82, line 214: Can you add a comment that this is disabled on ubuntu and add a link to the issue?

suhsteve: Comment added

azure-pipelines.yml
suhsteve: This is filtered due to the issue you found in this PR.

1.0.0 codebase is using `.Csv("Resources\\people.csv");`

Which doesn't work on linux which requires `.Csv($"{TestEnvironment.ResourceDirectory}people.csv");` to be OS agnostic.

imback82: Can you add a comment that this is due to the test issue, not the functionality changes? And you can add a link to this PR for the fix.

imback82 (approved):
LGTM, thanks @suhsteve!

imback82:
@Niharikadutta could you add this to 1.0.1 release note?

PR #745, opened 2020-10-15
Fix flaky CallbackTests.TestCallbackHandlers Test

suhsteve:
Instead of checking that reading from the socket throws an exception, check `CallbackConnection` to verify if it is running instead.

Cancelling the `CancellationToken` changes the _isRunning flag on the `CallbackConnection` from true to false.  Testing that this flag has been set correctly is the right thing to verify versus checking communication states/exceptions while trying to read the socket streams.

### DIFF
--- a/src/csharp/Microsoft.Spark.UnitTest/CallbackTests.cs
+++ b/src/csharp/Microsoft.Spark.UnitTest/CallbackTests.cs
@@ -164,6 +164,15 @@ namespace Microsoft.Spark.UnitTest
             {
                 using ISocketWrapper serverSocket = serverListener.Accept();
                 WriteAndReadTestData(serverSocket, callbackHandler, inputToHandler);
+
+                if (callbackHandler.Throws)
+                {
+                    Assert.False(callbackConnection.IsRunning);
+                }
+                else
+                {
+                    Assert.True(callbackConnection.IsRunning);
+                }
             }
         }
 
--- a/src/csharp/Microsoft.Spark.UnitTest/CallbackTests.cs
+++ b/src/csharp/Microsoft.Spark.UnitTest/CallbackTests.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Spark.UnitTest
                 clientSockets[i] = clientSocket;
                 clientSocket.Connect(ipEndpoint.Address, ipEndpoint.Port);
 
-                WriteAndReadTestData(clientSocket, callbackHandler, i, new CancellationToken());
+                WriteAndReadTestData(clientSocket, callbackHandler, i);
             }
 
             Assert.Equal(connectionNumber, callbackServer.CurrentNumConnections);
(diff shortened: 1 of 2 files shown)


imback82:
> Instead of checking that reading from the socket throws an exception, check `CallbackConnection` to verify if it is running instead.

Can you update the PR description why the previous behavior could be flakey?

suhsteve:
> > Instead of checking that reading from the socket throws an exception, check `CallbackConnection` to verify if it is running instead.
> 
> Can you update the PR description why the previous behavior could be flakey?

Updated.

imback82:
> checking communication states/exceptions while trying to read the socket streams.

Do you know why this was flaky?

suhsteve:
> > checking communication states/exceptions while trying to read the socket streams.
> 
> Do you know why this was flaky?

No I wasn't able to pinpoint why exactly the socket wasn't throwing an exception.

src/csharp/Microsoft.Spark.UnitTest/CallbackTests.cs
```diff
@@ -153,17 +153,33 @@ public void TestCallbackHandlers()
             clientSocket.Connect(ipEndpoint.Address, ipEndpoint.Port);
 
             var callbackConnection = new CallbackConnection(0, clientSocket, callbackHandlersDict);
-            Task.Run(() => callbackConnection.Run(token));
+            Task task = Task.Run(() => callbackConnection.Run(token));
 
-            using ISocketWrapper serverSocket = serverListener.Accept();
-            WriteAndReadTestData(serverSocket, callbackHandler, inputToHandler, token);
+            if (token.IsCancellationRequested)
```
Niharikadutta, line 158: How have we verified this change removes the flakiness of these tests? Do we need to run this multiple times to confirm?

suhsteve: Yes, this has already been verified.

src/csharp/Microsoft.Spark.UnitTest/CallbackTests.cs
```diff
@@ -153,17 +153,33 @@ public void TestCallbackHandlers()
...
+                using ISocketWrapper serverSocket = serverListener.Accept();
+                WriteAndReadTestData(serverSocket, callbackHandler, inputToHandler);
+
+                if (callbackHandler.Throws)
+                {
+                    Assert.False(callbackConnection.IsRunning);
+                }
+                else
+                {
+                    Assert.True(callbackConnection.IsRunning);
```
imback82, line 174: Question. When is this running `callbackConnection` stopped?

suhsteve: The callbackConnection can be stopped when the JVM side closes the connection cleanly, with a "CallbackFlags.CLOSE` request, or when reading the stream the socket is closed unexpectedly, or when the cancellation token is triggerd (ie, when the callback server shutsdown).

imback82: I mean in this function?

suhsteve: It'll should stop once the GC cleans up the sockets.

suhsteve: Or if we can manually trigger the cancellation token after each test.

imback82: Yea, can you make the clean up deterministic please?

suhsteve: CallbackConnections stopped.

imback82 (approved):
LGTM, thanks @suhsteve!

PR #759, opened 2020-10-28
Check whether file is found before trying to dereference it

### LINKED ISSUE #758: [BUG]: UDF SerDe cannot get to FileNotFoundException because of null reference exception
**Describe the bug**

When using a UDF if you do not have the dll in the working directory or have not configured DOTNET_ASSEMBLY_SEARCH_PATHS (or have configured it incorrectly) you get a Null Reference Exception in "/_/src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs:line 168"

https://stackoverflow.com/questions/61496962/dotnet-apache-spark-object-reference-not-set-to-an-instance-of-an-object

```csharp
private static Type DeserializeType(TypeData typeData) =>
            s_typeCache.GetOrAdd(
                typeData,
                td =>
                {
                    Type type = AssemblyLoader.LoadAssembly(
                        td.AssemblyName,
                        td.AssemblyFileName).GetType(td.Name);
                    if (type == null)
                    {
                        throw new FileNotFoundException(
                            string.Format(
                                "Assembly '{0}' file not found '{1}'",
                                td.AssemblyName,
                                td.AssemblyFileName));
                    }

                    return type;
                });

```

```csharp
Type type = AssemblyLoader.LoadAssembly(
                        td.AssemblyName,
                        td.AssemblyFileName).GetType(td.Name);  <-- assembly can be null so this GetType can fail
                    if (type == null)  

```

**To Reproduce**

Call a UDF but don't have the dll in the working directory or misconfigure DOTNET_ASSEMBLY_SEARCH_PATHS

**Expected behavior**

Instead of "Object reference not set to an instance of an object." throw the FileNotFound exception which should help people to troubleshoot why the assembly is not being loaded.

GoEddie:
Fixes #758 

The current code causes a NullReferenceException when the assembly cannot be found so this switches the logic slightly and also adds some additional detail to describe how the DOTNET_ASSEMBLY_SEARCH_PATHS environment variable is set to help users understand why their dll's are not being loaded.

I wasn't sure on writing a test for this, if it needs a test let me know, I did think about adding a check for a FileNotFoundException for a file that doesn't exist and call `UdfSerDe.Deserialize(UdfData udfData)`.

I was also unsure what to throw if the assembly had been loaded but the type couldn't for some reason, I went with TypeLoadException but let me know if you think something else is more appropriate.

The point of this really is that I have seen this pop up a few times and the NullReferenceException hides a genuinely useful FileNotFoundException.

Ed

### DIFF
--- a/src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
+++ b/src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
@@ -262,34 +262,18 @@ namespace Microsoft.Spark.Utils
                 typeData,
                 td =>
                 {
-                    Assembly assembly = AssemblyLoader.LoadAssembly(
+                    Type type = AssemblyLoader.LoadAssembly(
                         td.AssemblyName,
-                        td.AssemblyFileName);
-                    if (assembly == null)
-                    {
-                        string searchPath =
-                            Environment.GetEnvironmentVariable("DOTNET_ASSEMBLY_SEARCH_PATHS");
-                        
-                        if (String.IsNullOrEmpty(searchPath))
-                        {
-                            searchPath = "Empty";
-                        }
-                        
-                        throw new FileNotFoundException(
-                            string.Format(
-                                "Assembly '{0}' file not found '{1}'. " 
-                                + "Current DOTNET_ASSEMBLY_SEARCH_PATHS '{2}'",
-                                td.AssemblyName,
-                                td.AssemblyFileName,
-                                searchPath));
-                    }
-
-                    Type type = assembly.GetType(td.Name);
+                        td.AssemblyFileName)?.GetType(td.Name);
                     if (type == null)
                     {
-                        throw new TypeLoadException(
-                            string.Format("Unable to load Type '{0}' from Assembly '{1}'"));
+                        throw new FileNotFoundException(
+                            string.Format(
+                                "Assembly '{0}' file not found '{1}'",
+                                td.AssemblyName,
+                                td.AssemblyFileName));
                     }
+
                     return type;
                 });
     }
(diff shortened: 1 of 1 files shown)


imback82:
Thanks @GoEddie. I think this can be useful. @suhsteve Can you help with the review? Thanks!

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -262,18 +262,34 @@ private static TypeData SerializeType(Type type)
                 typeData,
                 td =>
                 {
-                    Type type = AssemblyLoader.LoadAssembly(
+                    Assembly assembly = AssemblyLoader.LoadAssembly(
```
suhsteve, line 265: I think we can just add the null conditional operator `?.` instead when calling `GetType`.    `AssemblyLoader.LoadAssembly` should already log a warning indicating the assembly filename and the different paths searched.

GoEddie: I have switched and it throws the FileNotFound and I can see the warning that logs all of the paths before the exceptions.

src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs
```diff
@@ -264,7 +264,7 @@ private static TypeData SerializeType(Type type)
                 {
                     Type type = AssemblyLoader.LoadAssembly(
                         td.AssemblyName,
-                        td.AssemblyFileName).GetType(td.Name);
+                        td.AssemblyFileName)?.GetType(td.Name);
```
suhsteve, line 267: Maybe we can also use the `GetType` method that throws on error as well.
```suggestion
                        td.AssemblyFileName)?.GetType(td.Name, true);
```

suhsteve (approved):
LGTM.  Thanks @GoEddie !

imback82 (approved):
LGTM, thanks @GoEddie!

Issue #328, opened 2019-11-07
[FEATURE REQUEST]: support for something like dbutils to move/copy/delete data easily


fxd1111359:
**Is your feature request related to a problem? Please describe.**
There is a request to save one single file in ADLS Gen2 as the output rather than a folder containing several files with partition GUID. In order to do it, in scala, we can play a trick by using dutils: coalesce the files into one and save the folder, copy the merged csv file and rename it, remove the temp folder. 
A sample code is like below,

	val fileprefix= "abfss://filesystem@storageaccount.dfs.core.windows.net/PageView/V1"

	agg_pageView_df.coalesce(1)
	  .write
	  .format("com.databricks.spark.csv")
	  .mode("overwrite")
	  .option("header", "true")
	  .option("inferSchema", "true")
	  .save(fileprefix + ".tmp")

	val partition_path = dbutils.fs.ls(fileprefix+".tmp/")
		 .filter(file=>file.name.endsWith(".csv"))(0).path

	dbutils.fs.cp(partition_path,fileprefix+".csv")
	dbutils.fs.rm(fileprefix+".tmp",recurse=true)

There is no support like dbutils in Spark.Net. We may need to use a separate ADLS client to connect to the storage and move/delete/copy the data.

That will be great if Spark.Net support this kind of data processing.

**Describe the solution you'd like**
Add support for something like dbutils, then you don't need to connect to the storage again with an independent client (like ALDS client) as spark already connect to it.

rapoth:
Unfortunately, dbutils is closed source and there is not much we can do to add support (we are open to contributions, however!). This may change if Databricks decides to open source dbutils or adds support themselves. Have you considered making a feature request to Databricks?

bigdatamoore:
Honestly for this, I think Spark should just have a built-in way of outputting a single file instead of a directory with part files.  This use case comes up so much that somebody should just add it to the Spark APIs.  Since that is not the case, the best way I have found for doing this is the copymerge method that used to be part of Hadoop. You can find a good pyspark implementation in this stack post: https://stackoverflow.com/questions/29025147/how-can-i-merge-spark-results-files-without-repartition-and-copymerge   

However I don't know how you would do this in dotnet.

imback82:
[...]

PR #787, opened 2020-12-02
FileSystem extension skeleton

AFFogarty:
This PR provides the skeleton for an extension that implements `FileSystem`.  This PR only implements the `Delete()` API, which allows users to delete files.  The rest of the APIs are intentionally omitted so that they can be implemented in future PRs.

This PR also provides the skeleton for `Configuration`, which is returned by a new method `SparkContext.HadoopConfiguration()`.  This `Configuration` skeleton does not currently implement any methods, but can be passed to `FileSystem.Get()` to get the `FileSystem` for a given `SparkContext`.

Example:

```
SparkSession spark = ...

FileSystem fs = FileSystem.Get(spark.SparkContext.HadoopConfiguration())
```

This PR relates to #328.


src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    /// <summary>
+    /// An abstract base class for a fairly generic filesystem. It may be implemented as a distributed filesystem, or
+    /// as a "local" one that reflects the locally-connected disk. The local version exists for small Hadoop instances
+    /// and for testing.
+    /// 
+    /// All user code that may potentially use the Hadoop Distributed File System should be written to use an FileSystem
+    /// object. The Hadoop DFS is a multi-machine system that appears as a single disk.It's useful because of its fault
+    /// tolerance and potentially very large capacity.
+    /// </summary>
```
Niharikadutta, line 19: Please reformat to keep each line within the 110 character limit

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
```diff
@@ -0,0 +1,51 @@
...
+using Microsoft.Spark.Interop.Ipc;
+
+namespace Microsoft.Spark.Extensions.FileSystem
+{
+    /// <summary>
+    /// An abstract base class for a fairly generic filesystem. It may be implemented as a distributed filesystem, or
+    /// as a "local" one that reflects the locally-connected disk. The local version exists for small Hadoop instances
+    /// and for testing.
+    /// 
+    /// All user code that may potentially use the Hadoop Distributed File System should be written to use an FileSystem
```
Niharikadutta, line 16:
```suggestion
    /// All user code that may potentially use the Hadoop Distributed File System should be written to use a FileSystem
```

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// object. The Hadoop DFS is a multi-machine system that appears as a single disk.It's useful because of its fault
```
Niharikadutta, line 17:
```suggestion
    /// object. The Hadoop DFS is a multi-machine system that appears as a single disk. It's useful because of its fault
```

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// </summary>
+        /// <param name="sparkContext">The SparkContext whose configuration will be used.</param>
+        /// <returns>The FileSystem.</returns>
+        public static FileSystem Get(SparkContext sparkContext)
+        {
+            // TODO: Expose hadoopConfiguration as a .NET class and add an override for Get() that takes it.
+            JvmObjectReference hadoopConfiguration = (JvmObjectReference)
+                ((IJvmObjectReferenceProvider)sparkContext).Reference.Invoke("hadoopConfiguration");
+
+            return new JvmReferenceFileSystem(
```
Niharikadutta, line 33: Why do we need `JvmReferenceFileSystem` to encapsulate the JVM object, why can't we do that within `FileSystem` itself?

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
```diff
@@ -0,0 +1,51 @@
...
+                    "org.apache.hadoop.fs.FileSystem",
+                    "get",
+                    hadoopConfiguration));
+        }
+
+        /// <summary>
+        /// Delete a file.
+        /// </summary>
+        /// <param name="path">The path to delete.</param>
+        /// <param name="recursive">If path is a directory and set to true, the directory is deleted else throws an
```
Niharikadutta, line 44: Exceeding 110 character limit.

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/JvmReferenceFileSystem.cs
```diff
@@ -0,0 +1,44 @@
...
+            _jvmObject = jvmObject;
+        }
+
+        JvmObjectReference IJvmObjectReferenceProvider.Reference => _jvmObject;
+
+        /// <summary>
+        /// Delete a file.
+        /// </summary>
+        /// <param name="path">The path to delete.</param>
+        /// <param name="recursive">If path is a directory and set to true, the directory is deleted else throws an
```
Niharikadutta, line 28: Exceeding 110 character limit

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/JvmReferenceFileSystem.cs
```diff
@@ -0,0 +1,44 @@
...
+            JvmObjectReference pathObject =
+                SparkEnvironment.JvmBridge.CallConstructor("org.apache.hadoop.fs.Path", path);
+
+            return (bool)_jvmObject.Invoke("delete", pathObject, recursive);
+        }
+
+        public override void Dispose()
+        {
+            _jvmObject.Invoke("close");
+        }
```
Niharikadutta, line 42:
```suggestion
        public override void Dispose() => _jvmObject.Invoke("close");
```

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem.E2ETest/Constants.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Spark.Extensions.FileSystem.E2ETest
+{
+    /// <summary>
+    /// Constants related to the FileSystem test suite.
+    /// </summary>
+    internal class Constants
```
Niharikadutta, line 10: Is this file needed? What else are we expecting to be added here?

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem.E2ETest/FileSystemTests.cs
```diff
@@ -0,0 +1,39 @@
...
+            using FileSystem fs = FileSystem.Get(_spark.SparkContext);
+
+            using var tempDirectory = new TemporaryDirectory();
+            string path = Path.Combine(tempDirectory.Path, "temp-table");
+            _spark.Range(25).Write().Format("parquet").Save(path);
+
+            Assert.True(fs.Delete(path, true));
+            Assert.False(fs.Delete(path, true));
+        }
+    }
```
Niharikadutta, line 38: Can we add another test that validates the `Delete` API by checking if the file got deleted from the file system?

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem.E2ETest/Microsoft.Spark.Extensions.FileSystem.E2ETest.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <IsPackable>false</IsPackable>
```
Niharikadutta, line 5: Why are we setting this to false?

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem.E2ETest/Microsoft.Spark.Extensions.FileSystem.E2ETest.csproj
```diff
@@ -0,0 +1,13 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\Microsoft.Spark.E2ETest\Microsoft.Spark.E2ETest.csproj" />
+    <ProjectReference Include="..\..\Microsoft.Spark\Microsoft.Spark.csproj" />
+    <ProjectReference Include="..\Microsoft.Spark.Extensions.FileSystem\Microsoft.Spark.Extensions.FileSystem.csproj" />
+  </ItemGroup>
```
Niharikadutta, line 12: nit: add empty line

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// tolerance and potentially very large capacity.
+    /// </summary>
+    public abstract class FileSystem : IDisposable
+    {
+        /// <summary>
+        /// Returns the configured FileSystem implementation.
+        /// </summary>
+        /// <param name="sparkContext">The SparkContext whose configuration will be used.</param>
+        /// <returns>The FileSystem.</returns>
+        public static FileSystem Get(SparkContext sparkContext)
```
imback82, line 27: This is weird. Is this like a factory? How can I create a new type of FileSystem?

Why not just expose the Hadoop FileSystem directly?

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
AFFogarty: In the JVM implementation, `FileSystem` is an abstract class.  I wanted to keep that same pattern here.

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem.E2ETest/Microsoft.Spark.Extensions.FileSystem.E2ETest.csproj
AFFogarty: Hmmm, I just copied this from the other extension `E2ETest` projects.  I will remove it everywhere.

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
AFFogarty: Oops, I thought we increased it to 120 for some reason.  Fixed.

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
AFFogarty: Fixed.

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem.E2ETest/Constants.cs
AFFogarty: This is the pattern we're using in all the test projects.

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem.E2ETest/FileSystemTests.cs
AFFogarty: Changed `TestSignatures()` to only test the signatures, and moved the functional testing into a new test that validates that the file is properly deleted.

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem.E2ETest/Microsoft.Spark.Extensions.FileSystem.E2ETest.csproj
AFFogarty: Fixed.

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
AFFogarty: This is the pattern from the Hadoop `FileSystem` class:  https://hadoop.apache.org/docs/current/api/org/apache/hadoop/fs/FileSystem.html#get-org.apache.hadoop.conf.Configuration-

FileSystem` is an abstract class with static `get()` factory methods that return concrete implementations based on the configuration parameters.

For my .NET implementation, I've added an override of `Get()` that takes `SparkContext` so that we don't have to expose `SparkContext.hadoopConfiguration`.

imback82: If you are mimicking the hadoop `FileSystem`, shall we follow the same signature (expose minimal `Configuration`)? Also, we should add `Hadoop` to the namespace? (and let's add the link to the comment as well)

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/JvmReferenceFileSystem.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using Microsoft.Spark.Interop;
+using Microsoft.Spark.Interop.Ipc;
+
+namespace Microsoft.Spark.Extensions.FileSystem
+{
+    /// <summary>
+    /// <see cref="FileSystem"/> implementation that wraps a corresponding FileSystem object in the JVM.
+    /// </summary>
+    public class JvmReferenceFileSystem : FileSystem, IJvmObjectReferenceProvider
```
imback82, line 13: Why is `JvmReference` public?

imback82: I am thinking that we should just put APIs into `FileSystem` since `get` handles the getting the right concrete class. In what scenarios, do you see we need a concrete implementation of `FileSystem` other than hadoop filesystem?

imback82: And just name the package as `Microsoft.Spark.Extensions.Hadoop.FileSystem` so that there is no confusion; i.e., we are just wrapping `org.apache.hadoop.fs.FileSystem` more or less.

AFFogarty: Sure, I'll just make `FileSystem` concrete.  I was just trying to keep it as similar to the JVM implementation as possible.  But I suppose we won't have any POCO `FileSystem` implementations.

AFFogarty: Made `FileSystem` concrete and removed `JvmReferenceFileSystem`.

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
AFFogarty: Added `Hadoop` to the namespace.

For this PR, I just wanted to provide an MVP skeleton so that it would be easy for community members to contribute APIs in additional PRs.  I'm thinking that we can invite others to contribute `Configuration` if they want to.  Thoughts?

imback82: Looks like we cannot define `Configuration` class in the extension package, since `SparkContext.hadoopConfiguration` will be inside `Microsoft.Spark`..

What if we add `Hadoop` directory under https://github.com/dotnet/spark/tree/master/src/csharp/Microsoft.Spark and add `FileSystem.cs` and `Configuration.cs`? Note that we don't have to expose any of the APIs for `Configuration`.  I just want to be able to create `FileSystem` by `FileSystem.Get(sparkContext.HadoopConfiguration)`. Since we are at 1.0, we want to avoid breaking public APIs if possible.

AFFogarty: Thoughts @rapoth? I know you wanted to put `FileSystem` in an extension.

src/csharp/Extensions/Microsoft.Spark.Extensions.Hadoop.FileSystem/FileSystem.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Microsoft.Spark.Extensions.Hadoop.FileSystem
+{
+    /// <summary>
+    /// A fairly generic filesystem. It may be implemented as a distributed filesystem, or as a "local" one
+    /// that reflects the locally-connected disk. The local version exists for small Hadoop instances and for
+    /// testing.
+    /// 
+    /// All user code that may potentially use the Hadoop Distributed File System should be written to use a FileSystem
+    /// object. The Hadoop DFS is a multi-machine system that appears as a single disk. It's useful because of its fault
```
Niharikadutta, line 17: Exceeding 110 character limit.

src/csharp/Extensions/Microsoft.Spark.Extensions.Hadoop.FileSystem/FileSystem.cs
```diff
@@ -0,0 +1,66 @@
...
+            _jvmObject = jvmObject;
+        }
+
+        JvmObjectReference IJvmObjectReferenceProvider.Reference => _jvmObject;
+
+        /// <summary>
+        /// Returns the configured FileSystem implementation.
+        /// </summary>
+        /// <param name="sparkContext">The SparkContext whose configuration will be used.</param>
+        /// <returns>The FileSystem.</returns>
```
Niharikadutta, line 35: nit: The FileSystem object.

Niharikadutta (commented):
A few small nits but otherwise LGTM. Thanks @AFFogarty !

src/csharp/Extensions/Microsoft.Spark.Extensions.Hadoop.FileSystem/FileSystem.cs
AFFogarty: Made a reference.

src/csharp/Extensions/Microsoft.Spark.Extensions.Hadoop.FileSystem/FileSystem.cs
AFFogarty: Fixed.

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/FileSystem.cs
AFFogarty: Synced with @rapoth.  We will go with @imback82 's approach.

src/csharp/Extensions/Microsoft.Spark.Extensions.FileSystem/JvmReferenceFileSystem.cs
AFFogarty: Changed implementation to use `Configuration`.

src/csharp/Microsoft.Spark/SparkContext.cs
```diff
@@ -312,6 +313,13 @@ public Broadcast<T> Broadcast<T>(T value)
             return new Broadcast<T>(this, value);
         }
 
+        /// <summary>
+        /// A default Hadoop Configuration for the Hadoop code (e.g. file systems) that we reuse.
+        /// </summary>
+        /// <returns>The Hadoop Configuration.</returns>
+        public Configuration HadoopConfiguration() =>
```
AFFogarty, line 320: @imback82 Not sure if this should be a function or a property.

imback82: Since this is `def` on the Scala side, this will be a method on C# side. So the current approach is good.

src/csharp/Microsoft.Spark/Hadoop/FS/FileSystem.cs
```diff
@@ -0,0 +1,58 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Microsoft.Spark.Hadoop.Conf;
+using Microsoft.Spark.Interop;
+using Microsoft.Spark.Interop.Ipc;
+
+namespace Microsoft.Spark.Hadoop.FS
```
imback82, line 10:
```suggestion
namespace Microsoft.Spark.Hadoop.Fs
```
? Since we use `Ipc` instead of `IPC`?

src/csharp/Microsoft.Spark/Hadoop/FS/FileSystem.cs
```diff
@@ -0,0 +1,58 @@
...
+        JvmObjectReference IJvmObjectReferenceProvider.Reference => _jvmObject;
+
+        /// <summary>
+        /// Returns the configured <see cref="FileSystem"/>.
+        /// </summary>
+        /// <param name="conf">The configuration to use.</param>
+        /// <returns>The FileSystem.</returns>
+        public static FileSystem Get(Configuration conf) =>
+            new FileSystem((JvmObjectReference)SparkEnvironment.JvmBridge.CallStaticJavaMethod(
+                "org.apache.hadoop.fs.FileSystem", "get", conf));
```
imback82, line 39: nit: I believe you need to break for each param.

src/csharp/Microsoft.Spark.E2ETest/Hadoop/FileSystemTests.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Test that methods return the expected signature.
+        /// </summary>
+        [Fact]
+        public void TestSignatures()
+        {
+            using var tempDirectory = new TemporaryDirectory();
+
+            using FileSystem fs = Assert.IsAssignableFrom<FileSystem>(
```
imback82, line 32: you can remove `IsAssignableFrom` here since it's not going to compile if it doesn't return assignable `FileSystem`.

imback82 (commented):
LGTM (few minor comments), thanks @AFFogarty!

imback82:
@AFFogarty Can you also update the title/description now that this PR provides bindings to Hadoop Filesystem?

src/csharp/Microsoft.Spark.E2ETest/Hadoop/FileSystemTests.cs
AFFogarty: Removed.

src/csharp/TestApp/Program.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Microsoft.Spark.Extensions.Hadoop.FileSystem;
+using Microsoft.Spark.Sql;
+using System;
+
```
imback82, line 4: Can we remove this file?

AFFogarty: Oops, didn't mean to commit that. Removed.

src/csharp/Microsoft.Spark/Hadoop/FS/FileSystem.cs
AFFogarty: Fixed

AFFogarty:
Addressed nits and updated description.

imback82:
@AFFogarty can you check the failed tests if it's transient?

AFFogarty:
@imback82 I merged with `master` and now the tests are passing.

imback82 (approved):
LGTM, thanks @AFFogarty!

PR #793, opened 2020-12-09
 Using (processId, threadId) as key to mantain threadpool executor instead of only threadId

gaoyangxiaozhu:
- [x] There's a descriptive title that will make sense to other developers some time from now. 
- [x] There's associated issues. All PR's should have issue(s) associated - unless a trivial self-evident change such as fixing a typo. You can use the format `Fixes #nnnn` in your description to cause GitHub to automatically close the issue(s) when your PR is merged.
- [x] Your change description explains what the change does, why you chose your approach, and anything else that reviewers should know.
- [x] You have included any necessary tests in the same PR.

This PR attempts to fix the issue [792](https://github.com/dotnet/spark/issues/792) , instead of let multiple `DotnetBackendHandler` instances share one global standalone threadPool object to maintain the relationship between thread and thread id, it always new an new threadPool instance for each `DotnetBackendHandler` instance.

I have already verified and tested end to end in my side.

Issue #792, opened 2020-12-09
[BUG]: dotnet-interactive will hang due to deadlock if recursively call Spark Dotnet method


gaoyangxiaozhu:
#### for eaxmple:

we have two csharp notebooks - `Notebook1` and `Notebook2`,

where the content of `Notebook1` is something like below 
```csharp
Notebook.Run('Notebook2');
``` 
and the contnet of `Notebook2` is something like below
```csharp
spark.Range(5).Show();
```

In our java code, we will first initialize a dotnet-interactive instance just like what `livy` does (let's name it `dotnet1`) to run dotnet code from `Notebook1`,

and what the code `Notebook.Run('Notebook2')` does is actually executing something like `JvmBridge.CallConstructor("com.microsoft.spark.notebook.run", "Notebook2")`
,  sparkDotNet will eventually submit related JVM code to `DotnetBackendHandler` to run.

What the  JVM code `com.microsoft.spark.notebook.run("Notebook2")` does is actually initialize another dotnet instance (let's name it `dotnet2`) to run code from `Notebook2`.

Here, `dotnet1` instance is connected to one `DotnetBackendHandler` instance to run the related jvm code `com.microsoft.spark.notebook.run("Notebook2")`, 

`dotnet2` is connnected to another `DotnetBackendHandler` instance to run the related jvm code `spark.range(5).show()`.

However, since  sparkDotnet limit the threadId for `dotnet1` and `dotnet2` both are `1` if run in dotnet-interactive. And in one jvm process, for different `DotnetBackendHandler` instances, the current dotnetbackend code always use standtone threadPool object to maintain the map of threadId and thread, cause the jvm code `spark.range(5).show` has no chance to be executed due to deadlock.

Simplily,  the jvm code `com.microsoft.spark.notebook.run("Notebook2")` will be submit to thread with threadId is `1` to run, while this jvm code will eventuall try to submit another jvm code `spark.range(5).show` to same thread to run, then deadlock happened. And what we see from the `dotnet1` is it hang with no any results return.

**To Reproduce**
For my case, it is actually a corner case, only happened if you recursively initialize dotnet-interactive instances to run sparkdotnet code,  and the jvm code related to one dotnet-interactive is blocked to wait another jvm code submitted by another dotnet-interactive to run finished.   Hard for me to provide a repo way for you, you can just concat me (gayangya@microsoft.com) to have a sync.

**Expected behavior**
instead of hang with no any result, just return the code result to me.

[...]

Issue #15, opened 2019-04-24
Extract common classes from src/scala/microsoft-spark-<version>.


imback82:
PR #190 creates multiple jars, but the implementation is copied from one version to another.

Create a `common` directory and extract common classes from duplicate code. Note that even if class/code is exactly the same, you cannot pull out to a common class if it depends on spark.

spzSource:
Hi @imback82 
I'm happy to be a volunteer for working on that ticket.

imback82:
That will be great, thanks @spzSource!

spzSource:
Hi @imback82 

Before start work I just want to confirm if I correctly understand suggested approach.

Am I correct saying that the intention is to create separate maven project (for instance, `microsoft-spark-common`), which should compile into separate `.jar` file?

imback82:
Yea, I think that's one way to do it. But we have to make sure `microsoft-spark` is a fat JAR so that we don't break the existing customers' pipeline. Or we can put the common files into a common folder that different projects can build (not sure if this is doable). Does this make sense?

spzSource:
Hi @imback82 

Looks like I'm stuck right after creating common maven module. Almost all classes inherit from `org.apache.spark.util.Utils.Logging` which doesn't allow to move such classes into common module due to dependence on specific spark version.

Am I correctly understand that removing `Logging` inheritance is not the option? In any way I'm happy for any ideas how to mitigate the  problem.

PR #801, opened 2020-12-18
Fix for memory leak in JVMObjectTracker

### LINKED ISSUE #799: [BUG]: Memory Leak when running scheduled job using DotnetRunner
**Description**
There is massive memory leak when running scheduled jobs using `org.apache.spark.deploy.dotnet.DotnetRunner`. As a result it causes Spark Driver node crash and restart due to GC overhead or heap size exception.

**To Reproduce**
1. Create simple batch job
  ```cs
public static void Main(string[] args)
{
    var spark = SparkSession
        .Builder()
        .GetOrCreate();
    var dataFrame = spark.Read().Format("delta").Load("/delta/source");

    dataFrame = dataFrame
        .WithColumn("RequestDate", dataFrame.Col("RequestDate").Cast("timestamp"))
        .Filter(dataFrame.Col("RequestDate")
            .Gt(Functions.DateSub(Functions.CurrentTimestamp(), 7)));
    dataFrame
        .Write()
        .Format("delta")
        .Mode(SaveMode.Overwrite)
        .Save("/delta/destination");
}
  ```
_Important note_: delta folders do not contain any data. It doesn't matter for that particular case.

2. Create databricks job definition running by schedule
image

3. Wait a few hours and take a look at memory consumption metrics

**Actual Result**

image
Note: memory drops on the graph indicate driver crashes

Heap statistics gathered using `jmap` tool right before driver crash (take a look at `PS Old Generation` metric):
```
Attaching to process ID 2284, please wait...
Debugger attached successfully.
Server compiler detected.
JVM version is 25.265-b01

using thread-local object allocation.
Parallel GC with 4 thread(s)

Heap Configuration:
   MinHeapFreeRatio         = 0
   MaxHeapFreeRatio         = 100
   MaxHeapSize              = 6564085760 (6260.0MB)
   NewSize                  = 2187853824 (2086.5MB)
   MaxNewSize               = 2187853824 (2086.5MB)
   OldSize                  = 4376231936 (4173.5MB)
   NewRatio                 = 2
   SurvivorRatio            = 8
   MetaspaceSize            = 21807104 (20.796875MB)
   CompressedClassSpaceSize = 1073741824 (1024.0MB)
   MaxMetaspaceSize         = 17592186044415 MB
   G1HeapRegionSize         = 0 (0.0MB)

Heap Usage:
PS Young Generation
Eden Space:
   capacity = 877133824 (836.5MB)
   used     = 302337464 (288.3314743041992MB)
   free     = 574796360 (548.1685256958008MB)
   34.46879549362812% used
From Space:
   capacity = 479723520 (457.5MB)
   used     = 0 (0.0MB)
   free     = 479723520 (457.5MB)
   0.0% used
To Space:
   capacity = 675807232 (644.5MB)
   used     = 0 (0.0MB)
   free     = 675807232 (644.5MB)
   0.0% used
PS Old Generation
   capacity = 4376231936 (4173.5MB)
   used     = 4375784088 (4173.072898864746MB)
   free     = 447848 (0.42710113525390625MB)
   99.98976635593019% used

47640 interned Strings occupying 5093352 bytes.
```

GC logs:

```
2020-12-16T13:41:51.773+0000: [Full GC (Ergonomics) [PSYoungGen: 1068544K->959362K(1612800K)] [ParOldGen: 4273219K->4273218K(4273664K)] 5341763K->5232581K(5886464K), [Metaspace: 261859K->261859K(1339392K)], 3.8349085 secs] [Times: user=14.02 sys=0.05, real=3.83 secs] 
2020-12-16T13:41:55.904+0000: [Full GC (Ergonomics) [PSYoungGen: 1068544K->967217K(1612800K)] [ParOldGen: 4273218K->4273218K(4273664K)] 5341762K->5240435K(5886464K), [Metaspace: 261859K->261859K(1339392K)], 3.7493484 secs] [Times: user=13.92 sys=0.03, real=3.75 secs] 
2020-12-16T13:41:59.922+0000: [Full GC (Ergonomics) [PSYoungGen: 1068544K->975163K(1612800K)] [ParOldGen: 4273218K->4273218K(4273664K)] 5341762K->5248382K(5886464K), [Metaspace: 261861K->261861K(1339392K)], 3.8089805 secs] [Times: user=14.32 sys=0.03, real=3.80 secs] 
...
```

Analyzing memory dump it's visible that there are a lot of `HashSetEntry` and `string` objects living in Old Gen which never get garbage collected. 
image

Going deeper through object graph you can see that `JVMObjectTracker` (which is static class) keeps holding references to SparkContext and hence to aforementioned "rubbish" objects (like query execution plans and datasets) even after `DotnetBackend` stopped.

image

**Expected behavior**
- Job running by schedule using DotnetRunner must not cause Spark Driver crashes. 
- Memory consumption must not increase over time. Old gen size must be within predicted range.

**Additional context**

Affected versions: 
- microsoft-spark-2-3_2.11
- microsoft-spark-2-4_2.11
- microsoft-spark-3-0_2.12

Unfortunately I cannot attach memory dump here, as it is about 8 GB size.

**Development notes**

I've already done a few experiments with patched version, where `JVMObjectTracker` is cleaned when DotetBacked shutdown. This does help, so after applying the fix the the same job runs smoothly without any memory issues.

If everybody agree that described behavior is a bug, I'll be happy to create PR.

spzSource:
Fixes #799 

As per explanation in the linked issue there is memory leak in `JVMObjectTracker`. Because of `JVMObjectTracker` is singleton by it's nature, it retains references to JVM objects throughout the entire life cycle of spark driver node.
 
All these in turn causes massive memory leak when running `DotnetRunner` multiple time on the same driver node. For instance this happens when using [SetJar ](https://docs.microsoft.com/en-us/dotnet/spark/how-to-guides/databricks-deploy-methods#deploy-using-set-jar) approach in Databricks.

To fix the memory leak `JVMObjectTracker` is cleaned up right before `DotnetBackend` shutdown, so it releases references to tracked JVM objects, which results for successful garbage collection against these objects.

Heap statistics using `jmap` tool after the fix applied:
```
Attaching to process ID 887, please wait...
Debugger attached successfully.
Server compiler detected.
JVM version is 25.265-b11

using thread-local object allocation.
Parallel GC with 4 thread(s)

Heap Configuration:
   MinHeapFreeRatio         = 0
   MaxHeapFreeRatio         = 100
   MaxHeapSize              = 6564085760 (6260.0MB)
   NewSize                  = 2187853824 (2086.5MB)
   MaxNewSize               = 2187853824 (2086.5MB)
   OldSize                  = 4376231936 (4173.5MB)
   NewRatio                 = 2
   SurvivorRatio            = 8
   MetaspaceSize            = 21807104 (20.796875MB)
   CompressedClassSpaceSize = 1073741824 (1024.0MB)
   MaxMetaspaceSize         = 17592186044415 MB
   G1HeapRegionSize         = 0 (0.0MB)

Heap Usage:
PS Young Generation
Eden Space:
   capacity = 1568145408 (1495.5MB)
   used     = 1502102368 (1432.5164489746094MB)
   free     = 66043040 (62.983551025390625MB)
   95.78846198426007% used
From Space:
   capacity = 303038464 (289.0MB)
   used     = 232644176 (221.8667755126953MB)
   free     = 70394288 (67.13322448730469MB)
   76.77051055802606% used
To Space:
   capacity = 309854208 (295.5MB)
   used     = 0 (0.0MB)
   free     = 309854208 (295.5MB)
   0.0% used
PS Old Generation
   capacity = 4376231936 (4173.5MB)
   used     = 532408504 (507.7443161010742MB)
   free     = 3843823432 (3665.755683898926MB)
   12.16591149157959% used

70301 interned Strings occupying 8281616 bytes.
```
Usage of Old gen space is 12%, comparing with 99,9% before the fix.

**Point to discuss**:
My first attempt was to make  `JVMObjectTracker` non-static, which potentially may provide better protection from leaks in the future. But making it non-static causes multiple file being modified, moreover it looks like it will be pretty hard to make `DotnetForeachBatchHelper` work with non-static version of `JVMObjectTracker`.

Assuming all mentioned above I've decided to implement an easiest possible solution, but I absolutely do not mind discussing other approaches.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/spark?pullRequest=801) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/spark?pullRequest=801) before we can accept your contribution.<br/><br/>:x: spzSource [sign now](https://cla.dotnetfoundation.org/dotnet/spark?pullRequest=801)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/spark?pullRequest=801) it.</sub>

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/api/dotnet/DotnetBackendTest.scala
```diff
@@ -0,0 +1,19 @@
+package org.apache.spark.api.dotnet
```
suhsteve, line 1: can you add license headers to each of the new files ?

suhsteve:
@imback82 I think the fix this PR addresses and the fix for issue #792 / PR #793 are similar.  We can either leave `ThreadPool` and `JVMObjectTracker` as singletons and add the appropriate fixes, or change them to classes and each instance of `DotnetBackend` can have their own instances of them.

This PR takes the first approach, and #793 does the second, but I think we should try to be consistent with both fixes.  I'm partial to the 2nd approach and `CallbackClient` will also need to be revisited with a similar fix to allow multiple `DotnetBackend` instances running.

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/api/dotnet/DotnetBackendTest.scala
spzSource: Sure. Headers have been added.

spzSource:
> @imback82 I think the fix this PR addresses and the fix for issue #792 / PR #793 are similar. We can either leave `ThreadPool` and `JVMObjectTracker` as singletons and add the appropriate fixes, or change them to classes and each instance of `DotnetBackend` can have their own instances of them.
> 
> This PR takes the first approach, and #793 does the second, but I think we should try to be consistent with both fixes. I'm partial to the 2nd approach and `CallbackClient` will also need to be revisited with a similar fix when running multiple `DotnetBackend` instances.

Hi @suhsteve and @imback82 
Just to let you know. The fix I've created solves issue with memory leak, but causes job isolation issues, when all tracking object for one  job are cleaned during shutdown of another job.

The only sensible way to fix that is to make `JVMObjectTracker` non-static, so that each new `DotnetBackend` will have it's own instance of `JVMObjectTracker`. I'm currently working on that.

suhsteve:
>Hi @suhsteve and @imback82
>
>Just to let you know. The fix I've created solves memory leak issue, but causes job isolation issues, when all tracking objects for one job may be cleaned up during shutdown of another job.
>
>The only sensible way to fix that is to make JVMObjectTracker non-static, so that each new DotnetBackend will have it's own instance of JVMObjectTracker. I'm currently working on that.

Sounds like a good plan.  Looking forward to the updated PR.

spzSource:
Hi @suhsteve 

PR was updated, so currently `JVMObjectTracker` is created per `DotnetBackend`. The changes were applied for all three versions 2.3, 2.4, and 3.0. Since there is no common part which is shared across versions, some code duplication is taking place. (Probably it would be better to increase priority of #15 for sake of reducing code duplication)

Looking forward for any feedback.

spzSource:
FYI I've done a little regression testing on my QA environment using latest fixes from current PR. All batch jobs, which previously suffered by memory leak, currently are running correctly without any memory issues.

imback82:
Thanks @spzSource for the update. I will get to this PR this week, sorry for the delay.

gaoyangxiaozhu:
looking forword the PR go to master.

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/api/dotnet/SerDeTest.scala
```diff
@@ -0,0 +1,386 @@
...
+import org.junit.Assert._
+import org.junit.function.ThrowingRunnable
+import org.junit.{Before, Test}
+
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, DataInputStream, DataOutputStream}
+import java.sql.Date
+import scala.collection.JavaConverters.{mapAsJavaMapConverter, seqAsJavaListConverter}
+
+@Test
+class SerDeTest {
```
imback82, line 20: Thanks for writing tests!

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/api/dotnet/Extensions.scala
```diff
@@ -0,0 +1,19 @@
...
+ * See the LICENSE file in the project root for more information.
+ */
+
+package org.apache.spark.api.dotnet
+
+import java.io.DataInputStream
+
+object Extensions {
+  implicit class DataInputStreamExt(stream: DataInputStream) {
+    def readN(n: Int): Array[Byte] = {
```
imback82, line 13: nit: `readNBytes`. Since this is only used inside `SerDeTests`, could you just define a helper function inside the test suite instead? (When I saw `in.readN`, I thought it was a public API)

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/api/dotnet/DotnetBackendTest.scala
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  @After
+  def after(): Unit = {
+    sut.close()
+  }
+
+  @Test
+  def shouldNotResetCallbackClient(): Unit = {
+    // specifying port = 0 to select port dynamically
```
imback82, line 31:
```suggestion
    // Specifying port = 0 to select port dynamically.
```

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/api/dotnet/DotnetBackend.scala
```diff
@@ -30,6 +30,10 @@ class DotnetBackend extends Logging {
   private[this] var channelFuture: ChannelFuture = _
   private[this] var bootstrap: ServerBootstrap = _
   private[this] var bossGroup: EventLoopGroup = _
+  private[this] val objectsTracker = new JVMObjectTracker
```
imback82, line 33: nit: `objectTracker` to match the class name?

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/api/dotnet/DotnetBackend.scala
```diff
@@ -64,6 +68,23 @@ class DotnetBackend extends Logging {
...
+  private[dotnet] def setCallbackClient(address: String, port: Int): Unit = {
+    callbackClient = callbackClient match {
+      case Some(_) => throw new Exception("Callback client already set.")
+      case None =>
+        logInfo(s"Connecting to a callback server at $address:$port")
+        Some(new CallbackClient(new SerDe(objectsTracker), address, port))
+    }
+  }
+
+  private[dotnet] def shutdownCallbackClient(): Unit = {
```
imback82, line 80: Should we make `setCallbackClient` and `shutdownCallbackClient` `synchronized` since there is no guaranteed that this will be called in a single thread from the Scala point of view?

imback82 (commented):
Few minor/nit comments, but generally looking great to me. Thanks @spzSource for working on this!

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/api/dotnet/SerDe.scala
```diff
@@ -18,7 +18,7 @@ import scala.collection.JavaConverters._
  * Functions to serialize and deserialize between CLR & JVM.
  * This implementation of methods is mostly identical to the SerDe implementation in R.
  */
-object SerDe {
+class SerDe(var tracker: JVMObjectTracker) {
```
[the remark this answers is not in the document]
spzSource: Good catch. Changed to `val`. Thanks.

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/api/dotnet/SerDe.scala
```diff
@@ -18,7 +18,7 @@ import scala.collection.JavaConverters._
  * Functions to serialize and deserialize between CLR & JVM.
```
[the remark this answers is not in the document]
spzSource: Done as suggested

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/api/dotnet/SerDeTest.scala
```diff
@@ -0,0 +1,386 @@
...
+  private var sut: SerDe = _
```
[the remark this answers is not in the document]
spzSource: Agree. Changes were applied for all places where `sut` used.

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/api/dotnet/DotnetBackendTest.scala
spzSource: Applied as suggested

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/api/dotnet/DotnetBackend.scala
spzSource: Agree. Field `objectsTracker` was renamed to `objectTracker `

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/api/dotnet/DotnetBackend.scala
spzSource: Agree. `synchronized` was initially applied for 3.0, but for some reason I've forgot to add it for 2.3 and 2.4. I've applied as suggested for all three versions.

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/api/dotnet/Extensions.scala
spzSource: Done as suggested for 2.3.
In 2.4 and 3.0 there are multiple usages of `readN` -> `readNBytes`:
- SerDeTest
- DotnetBackendHandlerTest
So for these versions I've took the liberty of leaving this extension method as is, but with making it visible only for `dotnet` package.
Anyway I'm happy to hear you thoughts about that.

spzSource:
Seems that latest build failed due to nuget feed issue:
> Unable to load the service index for source https://dotnet.myget.org/F/dotnet-try/api/v3/index.json.

Do anybody have idea what has happened with this feed?

imback82:
> Seems that latest build failed due to nuget feed issue:
> 
> > Unable to load the service index for source https://dotnet.myget.org/F/dotnet-try/api/v3/index.json.
> 
> Do anybody have idea what has happened with this feed?

Looks like the feed got deprecated recently. I will create a PR to fix this. Thanks!

imback82:
The build issue is fixed in #807. I pushed the changes to your branch.

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/api/dotnet/Extensions.scala
imback82: Ah, in that case, let's leave it as an extension method even for 2.3 so that it's easier to maintain.

spzSource: I've added the extension method back for 2.3.

spzSource:
All comments addressed, I think everything is ready for final review and approval :crossed_fingers:

imback82 (approved):
LGTM, thanks @spzSource!

Issue #381, opened 2019-12-30
[FEATURE REQUEST]: Implement ML Features 


GoEddie:
This is to track implementation of the ML-Features: https://spark.apache.org/docs/latest/ml-features

Bucketizer has been implemented in https://github.com/dotnet/spark/pull/378 but there are more features that should be implemented.

- [ ] Feature Extractors
  - [ ] TF-IDF
  - [ ] Word2Vec
  - [ ] CountVectorizer
  - [ ] FeatureHasher
- [ ] Feature Transformers
  - [ ] Tokenizer
  - [ ] StopWordsRemover
  - [ ] n-gram
  - [ ] Binarizer
  - [ ] PCA
  - [ ] PolynormalExpansion
  - [ ] Dicrete Cosine Transform (DCT)
  - [ ] StringIndexer
  - [ ] IndexToString
  - [ ] OneHotEncoderEstimator
  - [ ] VectorIndexer
  - [ ] Normalizer
  - [ ] StandardScaler
  - [ ] MinMaxScaler
  - [ ] MaxAbsScaler
  - [X] Bucketizer
  - [ ] ElementwiseProduct
  - [ ] SQLTransformer
  - [ ] VectorAssembler
  - [ ] VectorSizeHint
  - [ ] QuantileDiscretizer
  - [ ] Imputer
- [ ] Feature Selectors
  - [ ] VectorSlicer
  - [ ] RFormula
  - [ ] ChiSqSelector
- [ ] Locality Sensitive Hashing
  - [ ] LSH Operations
    - [ ] Feature Transformation
    - [ ] Approximate Similarity Join
    - [ ] Approximate Nearest Neighbour Search
  - [ ] LSH Algorithms
    - [ ] Bucketed Random Projection for Euclidean Distance
    - [ ] MinHash for Jaccard Distance

I'll try and carry on implementing theses (will wait until pr #378 has been completed to check that the code will go in Microsoft.Spark and not somewhere else etc.).

If anyone else is going to implement probably best to put a comment here or something??

imback82:
Thanks @GoEddie for compiling the list. 

> If anyone else is going to implement probably best to put a comment here or something??

This sounds good to me. We will review #378 in the next few days (some of us are out on vacation).

GoEddie:
Great thanks, @imback82 I didn’t expect anyone to reply until the new year :)

Niharikadutta:
[...]

PR #804, opened 2021-01-06
Included StringIndexer and StringIndexerModel along with related test…

ramanathanv:
Hi,

I am creating this pull request that implements the basic methods in StringIndexer and StringIndexerModel class along with Test cases.

This is related to "Implement ML Features" #381

### DIFF
--- a/src/csharp/Microsoft.Spark/ML/Feature/StringIndexer.cs
+++ b/src/csharp/Microsoft.Spark/ML/Feature/StringIndexer.cs
@@ -1,0 +1,174 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Spark.Interop;
+using Microsoft.Spark.Interop.Ipc;
+using Microsoft.Spark.Sql;
+using Microsoft.Spark.Sql.Types;
+
+namespace Microsoft.Spark.ML.Feature
+{
+    /// <summary>
+    /// <see cref="StringIndexer"/> encodes a string column of labels to a column of label indices.
+    /// </summary>
+    public class StringIndexer : FeatureBase<StringIndexer>, IJvmObjectReferenceProvider
+    {
+        private static readonly string s_StringIndexerClassName =
+            "org.apache.spark.ml.feature.StringIndexer";
+
+        /// <summary>
+        /// Create a <see cref="StringIndexer"/> without any parameters.
+        /// </summary>
+        public StringIndexer() : base(s_StringIndexerClassName)
+        {
+        }
+
+        /// <summary>
+        /// Create a <see cref="StringIndexer"/> with a UID that is used to give the
+        /// <see cref="StringIndexer"/> a unique ID.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


src/csharp/Microsoft.Spark/ML/Feature/StringIndexer.cs
```diff
@@ -0,0 +1,174 @@
...
+            new StructType(
+                (JvmObjectReference)_jvmObject.Invoke(
+                    "transformSchema",
+                    DataType.FromJson(_jvmObject.Jvm, value.Json)));
+
+        /// <summary>
+        /// Executes the <see cref="StringIndexer"/> and fits a model to the input data.
+        /// </summary>
+        /// <param name="source"></param>
+        /// <returns></returns>
```
Niharikadutta, line 59: Could you please add the `param` and `returns` description?

Niharikadutta:
@ramanathanv Is this PR ready to review? I see some `StringIndexerTests` tests failing in the pipeline.

ramanathanv:
> @ramanathanv Is this PR ready to review? I see some `StringIndexerTests` tests failing in the pipeline.

Hi @Niharikadutta ,

In the test case, I am trying to compare two List objects using Assert.Equal(). I notice in the test results that the Lists have the same content (see image below). Still there is an error.
image

I assume that there must be a different way to compare the Lists. Can you please suggest me a way an alternate way to do this comparison?

Thanks.

imback82:
> I notice in the test results that the Lists have the same content (see image below).

I see "..." in the output. Can you compare elements one by one to see where the difference comes from?

ramanathanv:
> > I notice in the test results that the Lists have the same content (see image below).
> 
> I see "..." in the output. Can you compare elements one by one to see where the difference comes from?

Hi @imback82 ,
I noticed that even the raw log file provides only 3 dots "...". Is there another way for me to check this?

imback82:
What I meant was if you are comparing two lists, you can compare elements in those lists one by one instead of comparing them at the list level. Does it make sense?

GoEddie:
In StringIndexer there are a few calls which are missing the get* at the beginning - if I change them all to get* then the test passes for me.

```
public string GetStringOrderType() => (string)_jvmObject.Invoke("stringOrderType");
```
should be
```
public string GetStringOrderType() => (string)_jvmObject.Invoke("getStringOrderType");
```

```
public string[] GetInputCols() => (string[])_jvmObject.Invoke("inputCols");
```
should be
```
public string[] GetInputCols() => (string[])_jvmObject.Invoke("getInputCols");
```

```
public string GetInputCol() => (string)_jvmObject.Invoke("inputCol");
```
should be
```
public string GetInputCol() => (string)_jvmObject.Invoke("getInputCol");
```

```
public string GetHandleInvalid() => (string)_jvmObject.Invoke("handleInvalid");
```
should be
```
public string GetHandleInvalid() => (string)_jvmObject.Invoke("getHandleInvalid");
```

If you change those then the tests should pass :)

ramanathanv:
> 
> If you change those then the tests should pass :)

Thanks @GoEddie for identifying the issue . I have fixed the property names. But the test fails for unknown reasons. Can you please re-initiate the build/tests?

PR #808, opened 2021-01-13
Update dotnet-try feed for install-interactive-notebook.sh

imback82:
The old feed for `dotnet-try` is deprecated. This PR proposes to update the feed for `install-interactive-notebook.sh`.


imback82 (commented):
Merging as the change only touches a script file, which is not a part of the build.

PR #810, opened 2021-01-14
[WIP] Sparse vector

GoEddie:
Implements SparseVector which is used by many of the spark ml objects.

This is an implementation for SparseVector, there are a couple of oddities to point out!

First, the constructor for SparseVector is private so I needed to call Vectors.sparse in the .NET SparseVector constructor.

The second thing is that when passing a SparseVector from Spark back to .NET (collect or udf) the scala code returns an ArrayList with four values for the SpaerseVector and I couldn't see any opportunity to intercept the SparseVector before it was serialized as the ArrayList. If anyone knows how to tell from the scala backend that the type being written is a SparseVector rather than an ArrayList with four values then I could include a new type on the serializer. As a half-way measure I have created a SparseVector.FromRow which takes the Row and returns a new SparseVector (including writing it back over the pickling interface) and added an example of how people can access the underlying values if they don't want to deal with the performance hit.

Let me know if this is ok or if it is possible to tell on the scala side what the type is so I can introduce a new type on the SerDe (only De side needs it)

Lastly, there is a new DataType SparseVectorUDT which includes the schema that we need to pass the SparseVector across, if this isn't the right approach let me know! If this approach is good there are three more user defined types (UDT's) to add - DenseVector/DenseVectorUDT, SparseMatrix/SparseMatrixUDT, and SparseMatrix/SparseMatrixUDT.

ed

***** Build Failures *****

I had a look at the build failures and it looks like there is a set of tests which run against the last published version of Microsoft.Spark.Worker (E2E Backward Compatibility Tests) - as this type isn't in that version it will always fail - do I need to check for the version of the spark worker anywhere or do anything else?


GoEddie:
also the build is failing because it looks like the worker isn't using the version of microsoft.spark.dll that includes the changes in the code?

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
```diff
@@ -322,4 +322,69 @@ internal override object FromInternal(object obj)
...
+    /// </summary>
+    public class SparseVectorUDT : DataType
+    {
+        /// <summary>
+        /// Returns JSON object describing this type.
+        /// </summary>
+        internal override object JsonValue =>
+            new JObject(
+                new JProperty("type", "udt"),
+                new JProperty("class", "org.apache.spark.ml.linalg.VectorUDT"),
```
imback82, line 337: Can you define this class under `Microsoft.Spark.ML.Linalg`?

imback82: How is this type used? I don't see this being used in this PR?

imback82:
> I had a look at the build failures and it looks like there is a set of tests which run against the last published version of Microsoft.Spark.Worker (E2E Backward Compatibility Tests) - as this type isn't in that version it will always fail - do I need to check for the version of the spark worker anywhere or do anything else?

You can filter out tests. For exmple: https://github.com/dotnet/spark/blob/d2822c33e5835e66863b39dc0a015d190852e76d/azure-pipelines.yml#L36

src/csharp/Microsoft.Spark/Sql/Types/ComplexTypes.cs
GoEddie: Will move it -or- it is used when calling CreateDataFrame to map the int,int,int[],double[] into a single column - a similar version appears in pyspark (not exactly the same) - if you think it is better for users to create it themselves I can take it out and document it instead?

imback82: I think we can move it. Could you also add a test?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/ML/Linalg/SparseVectorTests.cs
```diff
@@ -0,0 +1,84 @@
...
+        public SparseVectorTests(SparkFixture fixture)
+        {
+            _spark = fixture.Spark;
+        }
+
+        /// <summary>
+        /// Test that we can create SparseVector, collect it back to .NET and use it in a UDF
+        /// </summary>
+        [Fact]
+        public void TestPassingBetweenApacheSpark()
```
imback82, line 28: Can we name this to `TestSparseVectorE2EScenario`?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/ML/Linalg/SparseVectorUDTTests.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.Spark.ML.Linalg;
+using Microsoft.Spark.Sql;
+using Microsoft.Spark.Sql.Types;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace Microsoft.Spark.E2ETest.IpcTests.ML.Linalg
+{
+    [Collection("Spark E2E Tests")]
+    public class SparseVectorUDTTests
```
imback82, line 16: Can we combine this with `SparseVectorTests` and just add `TestSparseVectorUDT`?

imback82 (commented):
@Niharikadutta can you help reviewing this as well?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/ML/Linalg/SparseVectorTests.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            var schema = new StructType(new List<StructField>()
+            {
+                new StructField("name", new StringType()),
+                new StructField("hectorTheVector", new SparseVectorUDT())
+            });
+
+            var dataFrame =
+                _spark.CreateDataFrame(new List<GenericRow>() { new(new object[] { "hector", vector }) },
+                    schema);
```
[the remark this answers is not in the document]
GoEddie: have added it - the odd thing is rider put GenericRow in grey over the top of the text so it looked like it was there on mine

azure-pipelines.yml
```diff
@@ -16,27 +16,33 @@ variables:
...
-
-  backwardCompatibleTestOptions_Windows_2_4: ""
-  forwardCompatibleTestOptions_Windows_2_4: ""
-  backwardCompatibleTestOptions_Linux_2_4: ""
+  # Filter Microsoft.Spark.E2ETest.IpcTests.ML.Linalg.SparseVectorTests.TestSparseVectorE2EScenario
+  # Please see https://github.com/dotnet/spark/pull/810
+  backwardCompatiableTestIgnore_PR_810: "FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.ML.Linalg.SparseVectorTests.TestSparseVectorE2EScenario"
+
+  backwardCompatibleTestOptions_Windows_2_3: "--filter ($(backwardCompatiableTestIgnore_PR_810))"
+  forwardCompatibleTestOptions_Windows_2_3: "--filter ($(backwardCompatiableTestIgnore_PR_810))"
```
imback82, line 24: Weird, it shouldn't break the forward compatible tests since this new test shouldn't exist. (we are doing "Checkout dotnet/spark@refs/tags/v1.0.0 to s$(PATH_SEPARATOR)dotnet-spark-1.0.0" and build/test unless it's broken somehwere).

Could you share the log for the failures?

GoEddie: I had another look and the forward tests were breaking because maven was getting a reset connection: https://dev.azure.com/dnceng/public/_build/results?buildId=959021&view=logs&j=752a2194-d368-5dec-19f4-d06166099277&t=b757f20e-e579-55f4-086e-3f4cbc799e47

I’ll put them back in and see what happens.

imback82: I re-triggered two failing tests due to maven, and tests are all passing now.

imback82: Oh wait, it seems that we still have the filter for forward compatibility tests?

GoEddie: I was on my phone when I said i'd change it - have changed it now :).

GoEddie:
I've removed the filter from the forward tests and they passed once, will re-trigger and see if they still pass because the merge caused a failed build.

azure-pipelines.yml
```diff
@@ -35,10 +39,11 @@ variables:
   # tests due to https://github.com/dotnet/spark/pull/711
   backwardCompatibleTestOptions_Windows_3: "--filter \
   (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestDataFrameGroupedMapUdf)&\
-  (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestGroupedMapUdf)"
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestGroupedMapUdf)&\
+  ($(backwardCompatiableTestIgnore_PR_810))"
   forwardCompatibleTestOptions_Windows_3: ""
   backwardCompatibleTestOptions_Linux_3: $(backwardCompatibleTestOptions_Windows_3)
-  forwardCompatibleTestOptions_Linux_3: $(forwardCompatibleTestOptions_Linux_2_4)
```
imback82, line 41: So this is an existing bug? @suhsteve do you know anything about this why this flag was here?

suhsteve: This is safe to remove.  Seems like the Test Attribute skips Spark 3.0 tests.  https://github.com/dotnet/spark/blob/b16b4f64e6978432f0c3b7976f7a6153300c7bd9/src/csharp/Extensions/Microsoft.Spark.Extensions.Hyperspace.E2ETest/HyperspaceTests.cs#L57-L58

imback82: Cool, thanks.

imback82:
@Niharikadutta Could you take one more look?

src/csharp/Microsoft.Spark/ML/Linalg/SparseVectorUDT.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// Returns JSON object describing this type.
+        /// </summary>
+        internal override object JsonValue =>
+            new JObject(
+                new JProperty("type", "udt"),
+                new JProperty("class", "org.apache.spark.ml.linalg.VectorUDT"),
+                new JProperty("pyClass", "pyspark.ml.linalg.VectorUDT"),
+                new JProperty("sqlType", new JObject(
+                    new JProperty("type", "struct"),
+                    new JProperty("fields", new JArray(
```
suhsteve, line 26: Will there be any issues not including the `"metadata": {}` property to each of these fields ?

src/csharp/Microsoft.Spark/ML/Linalg/SparseVectorUDT.cs
```diff
@@ -0,0 +1,75 @@
...
+            return obj;
+        }
+
+        internal override bool NeedConversion() => true;
+
+        /// <summary>
+        /// Returns a readable string that represents this type.
+        /// </summary>
+        public override string SimpleString =>
+            "struct<`type`:TINYINT,`size`:INT,`indices`:ARRAY<INT>,`values`:ARRAY<DOUBLE>>";
```
suhsteve, line 73: shouldn't this be `vector` ?

src/csharp/Microsoft.Spark/ML/Linalg/SparseVectorUDT.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            if (obj is RowConstructor rowConstructor)
+            {
+                return rowConstructor.GetRow();
+            }
+
+            return obj;
+        }
+
+        internal override bool NeedConversion() => true;
```
suhsteve, line 67: I don't see where the `SparseVectorUDT` class is being used, but why do you need this ?  Does this get unpickled as a `Row` ?

src/csharp/Microsoft.Spark/ML/Linalg/SparseVectorUDT.cs
GoEddie: I don't *think* so (I could be wrong!) , in PySpark they have a VectorUDT which has to be used and this is the schema that VectorUDT returns (https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/linalg/VectorUDT.scala) see https://stackoverflow.com/questions/51527066/vectortype-for-structtype-in-pyspark-schema

src/csharp/Microsoft.Spark/ML/Linalg/SparseVectorUDT.cs
GoEddie: The class is for users so they don't have to define the schema anytime they want to use it - the alternative is that they would have to know to send a vector from their app to spark would have to know the type of vector (sparse or dense) and the schema they need to send it across.

src/csharp/Microsoft.Spark/ML/Linalg/SparseVector.cs
```diff
@@ -0,0 +1,159 @@
...
+            int length = row.GetAs<int>(1);
+            int[] indices = (int[])row.GetAs<ArrayList>(2).ToArray(typeof(int));
+            double[] values = (double[])row.GetAs<ArrayList>(3).ToArray(typeof(double));
+
+            return new SparseVector(length, indices, values);
+        }
+    }
+
+
+
```
Niharikadutta, line 158: extra whitespace

src/csharp/Microsoft.Spark/ML/Linalg/SparseVector.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// </summary>
+        /// <returns>Array of every element in the <see cref="SparseVector"/></returns>
+        public double[] ToArray() => (double[])_jvmObject.Invoke("toArray");
+
+        /// <summary>
+        /// Returns the JVM toString value rather than the .NET ToString default
+        /// </summary>
+        /// <returns>JVM toString() value</returns>
+        public override string ToString() => (string)_jvmObject.Invoke("toString");
+
```
Niharikadutta, line 87: I see there a few methods we haven't exposed here such as `asML()`, `copy()`, `size()`, `toJson()`.. Is there a reason for this?

GoEddie: A few reasons, the asMl - I didn’t implement the mlreader/mlwriter on any of the classes yet, was going to do a pr later on to back fill them all.

the copy, toJson, count etc I wasn’t sure if the whole approach to this with the udt was going to be acceptable so put in the minimum to make it useable - if this is all good then I’ve got DenseVector, SparseMatrix and DenseMatrix to add so will follow up with more pr’s.

GoEddie: If you think they have to be there let me know and I’ll add them.

src/csharp/Microsoft.Spark/ML/Linalg/SparseVectorUDT.cs
suhsteve: I'm a bit confused of how you're using this DataType.  Can you give some examples of its usage in the PR, maybe in the Tests ?

Also if we follow the pattern that's been used when creating `DataType`s, then shouldn't `SimpleString` should be the one defined [here] ?(https://github.com/apache/spark/blob/9db566a8821c02427434c551ee6e4d2501563dfa/python/pyspark/ml/linalg/__init__.py#L165)

src/csharp/Microsoft.Spark/ML/Linalg/SparseVectorUDT.cs
suhsteve: Can I see an example of how this is going to be used ?

src/csharp/Microsoft.Spark/ML/Linalg/SparseVectorUDT.cs
GoEddie: I've changed the simple string to match the value in pyspark now.

The data type is shown here in the test:

https://github.com/dotnet/spark/pull/810/files?file-filters%5B%5D=.cs#diff-129fac0a213898cb61e61adff29463de6deac7ab63ec58aa7784f02df20eecd9R41

The point of it is to stop users having to define the schema themselves - the classes are a bit of a pain to work with so having to define the schema for  Sparse/Dense Vector/Matrix is annoying - if you think that they are best left out of here then I can take it out again.

suhsteve: Okay I see the example now.   Can you modify the usage so users won't have to explicitly call `SparseVector.FromRow` ?

GoEddie: I’ll have another look at it, it was a while ago when I tried and there was something stopping it but I can’t exactly remember what!

GoEddie:
I believe this is coming in the ML project

PR #812, opened 2021-01-22
Remove unnecessary RestoreSources

mmitche:
General cleanup. These should not be necessary (legacy from older restore methods)


imback82 (approved):
LGTM, thanks @mmitche!

PR #815, opened 2021-01-29
Support for Hyperspace 0.4.0

AFFogarty:
This PR adds full support for [Hyperspace 0.4.0](https://github.com/microsoft/hyperspace/releases/tag/v0.4.0).  This includes new APIs `optimizeIndex()` and `index()` introduced in Hyperspace 0.3.0 and 0.4.0, respectively.

### DIFF
--- a/src/csharp/Extensions/Microsoft.Spark.Extensions.Hyperspace/Hyperspace.cs
+++ b/src/csharp/Extensions/Microsoft.Spark.Extensions.Hyperspace/Hyperspace.cs
@@ -74,7 +74,8 @@ namespace Microsoft.Spark.Extensions.Hyperspace
         /// modes as listed below.
         /// </summary>
         /// <param name="indexName">Name of the index to refresh.</param>
-        /// <param name="mode">Refresh mode. Currently supported modes are "incremental" and "full".</param>
+        /// <param name="mode">Refresh mode. Currently supported modes are <c>incremental</c> and
+        /// <c>full</c>.</param>
         [HyperspaceSince(HyperspaceVersions.V0_0_1)]
         public void RefreshIndex(string indexName, string mode = "full") =>
             _jvmObject.Invoke("refreshIndex", indexName, mode);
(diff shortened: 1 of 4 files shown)


src/csharp/Extensions/Microsoft.Spark.Extensions.Hyperspace/Hyperspace.cs
```diff
@@ -70,11 +70,37 @@ public Hyperspace(SparkSession spark)
...
-        /// Update indexes for the latest version of the data.
+        /// Update indexes for the latest version of the data. This API provides a few supported refresh
+        /// modes as listed below.
         /// </summary>
         /// <param name="indexName">Name of the index to refresh.</param>
+        /// <param name="mode">Refresh mode. Currently supported modes are <c>incremental</c> and
+        /// <c>full</c>.</param>
         [HyperspaceSince(HyperspaceVersions.V0_0_1)]
-        public void RefreshIndex(string indexName) => _jvmObject.Invoke("refreshIndex", indexName);
+        public void RefreshIndex(string indexName, string mode = "full") =>
```
AFFogarty, line 80: Do you think we should keep `mode` as a `string`, or introduce an `enum`?

imback82, line 80: Can we separate out the API that takes in the mode, which is from 0.3?

AFFogarty: Sure.  Do you think we should keep `mode` as a `string` in .Net, or make it an `enum`?

imback82: `string` is fine.

AFFogarty: Updated the PR with `RefreshIndex` split in 2.

imback82 (approved):
LGTM (pending tests), thanks @AFFogarty!

Issue #827, opened 2021-02-19
[FEATURE REQUEST]: Support for Spark 3.0.2


imback82:
Spark 3.0.2 has just been released and we need to update `microsoft-spark` JAR to enable the support.

Issue #811, opened 2021-01-20
[FEATURE REQUEST]: Support for Spark 3.1.1


imback82:
Spark 3.1.1 release is coming soon. I ran E2E tests against the RC version of Spark 3.1.1 and the following two tests fail due to the internal changes, which we should address to support Spark 3.1.1:
```
TestDataFrameGroupedMapUdf
TestGroupedMapUdf
```

PR #830, opened 2021-03-04
Support Spark 3.0.2, 3.1.1

suhsteve:
Add support for Spark `3.0.2` and `3.1.1`

Addresses #827 and #811

### DIFF
--- a/src/csharp/Extensions/Microsoft.Spark.Extensions.Delta.E2ETest/DeltaTableTests.cs
+++ b/src/csharp/Extensions/Microsoft.Spark.Extensions.Delta.E2ETest/DeltaTableTests.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Spark.Extensions.Delta.E2ETest
         /// Run the end-to-end scenario from the Delta Quickstart tutorial.
         /// </summary>
         /// <see cref="https://docs.delta.io/latest/quick-start.html"/>
-        [SkipIfSparkVersionIsLessThan(Versions.V2_4_2)]
+        [SkipIfSparkVersionIsNotInRange(Versions.V2_4_2, Versions.V3_1_1)]
         public void TestTutorialScenario()
         {
             using var tempDirectory = new TemporaryDirectory();
@@ -223,7 +223,7 @@ namespace Microsoft.Spark.Extensions.Delta.E2ETest
         /// <summary>
         /// Test that methods return the expected signature.
         /// </summary>
-        [SkipIfSparkVersionIsLessThan(Versions.V2_4_2)]
+        [SkipIfSparkVersionIsNotInRange(Versions.V2_4_2, Versions.V3_1_1)]
         public void TestSignaturesV2_4_X()
         {
             using var tempDirectory = new TemporaryDirectory();
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameWriterTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameWriterTests.cs
@@ -69,12 +69,7 @@ namespace Microsoft.Spark.E2ETest.IpcTests
 
                 // TODO: Test dfw.Jdbc without running a local db.
 
-                dfw.Option("path", tempDir.Path).SaveAsTable("TestTable");
-
-                dfw.InsertInto("TestTable");
-
-                dfw.Option("path", $"{tempDir.Path}TestSavePath1").Save();
-                dfw.Save($"{tempDir.Path}TestSavePath2");
+                dfw.Save($"{tempDir.Path}TestSavePath1");
 
                 dfw.Json($"{tempDir.Path}TestJsonPath");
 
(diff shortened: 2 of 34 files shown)


src/csharp/Extensions/Microsoft.Spark.Extensions.Delta.E2ETest/DeltaTableTests.cs
```diff
@@ -30,7 +30,7 @@ public DeltaTableTests(DeltaFixture fixture)
         /// Run the end-to-end scenario from the Delta Quickstart tutorial.
         /// </summary>
         /// <see cref="https://docs.delta.io/latest/quick-start.html"/>
-        [SkipIfSparkVersionIsLessThan(Versions.V2_4_2)]
+        [SkipIfSparkVersionIsNotInRange(Versions.V2_4_2, Versions.V3_1_1)]
```
suhsteve, line 33: Disable Delta tests that have code path that creates a `org.apache.spark.sql.catalyst.expressions.Alias` object.   `Delta 0.8.0` is not compatible with `Spark 3.1.1`  https://github.com/delta-io/delta/issues/594

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameWriterTests.cs
```diff
@@ -85,6 +80,12 @@ public void TestSignaturesV2_3_X()
                 dfw.Text($"{tempDir.Path}TestTextPath");
 
                 dfw.Csv($"{tempDir.Path}TestCsvPath");
+
+                dfw.Option("path", tempDir.Path).SaveAsTable("TestTable");
```
suhsteve, line 84: Setting `path` option and then calling save is not supported by default unless  `spark.sql.legacy.pathOptionBehavior.enabled` conf is set.  Reverse order in test.

https://github.com/apache/spark/blob/bb9abb098a1e26c220546616d655380c479b0e42/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala#L289

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamReaderTests.cs
```diff
@@ -63,6 +62,7 @@ public void TestSignaturesV2_3_X()
                 dsr.Parquet(Path.Combine(TestEnvironment.ResourceDirectory, "users.parquet")));
             Assert.IsType<DataFrame>
                 (dsr.Text(Path.Combine(TestEnvironment.ResourceDirectory, "people.txt")));
+            Assert.IsType<DataFrame>(dsr.Format("json").Option("path", jsonFilePath).Load());
```
suhsteve, line 65: Setting `path` option and then calling `load` is not supported by default unless  `spark.sql.legacy.pathOptionBehavior.enabled` conf is set.  Reverse order in test.

https://github.com/apache/spark/blob/bb9abb098a1e26c220546616d655380c479b0e42/sql/core/src/main/scala/org/apache/spark/sql/DataFrameReader.scala#L259

src/scala/microsoft-spark-3-0/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
```diff
@@ -34,7 +34,7 @@ import scala.util.Try
  */
 object DotnetRunner extends Logging {
   private val DEBUG_PORT = 5567
-  private val supportedSparkVersions = Set[String]("3.0.0", "3.0.1")
+  private val supportedSparkVersions = Set[String]("3.0.0", "3.0.1", "3.0.2")
```
Niharikadutta, line 37: Shouldn't we add `3.1.1` here as well?

Niharikadutta:
@suhsteve Could you add a description of the changes made as a part of this PR?

src/scala/microsoft-spark-3-1/src/main/scala/org/apache/spark/api/dotnet/DotnetBackend.scala
```diff
@@ -0,0 +1,113 @@
...
+          .addLast("handler", new DotnetBackendHandler(self, objectTracker))
+      }
+    })
+
+    channelFuture = bootstrap.bind(new InetSocketAddress("localhost", portNumber))
+    channelFuture.syncUninterruptibly()
+    channelFuture.channel().localAddress().asInstanceOf[InetSocketAddress].getPort
+  }
+
+  private[dotnet] def setCallbackClient(address: String, port: Int): Unit = synchronized {
```
Niharikadutta, line 70: Can we add a small summary for all these functions?

suhsteve:
> @suhsteve Could you add a description of the changes made as a part of this PR?

Done.

src/scala/microsoft-spark-3-1/src/main/scala/org/apache/spark/api/dotnet/DotnetBackend.scala
suhsteve: No, these are copied from 3-0.

imback82 (commented):
A couple of things:

Can you spit these into following PRs?
  * Support 3.0.2
  * PR for just creating a new JAR microsoft-spark-3.1 JAR, but no need to build it, but just copying files.
  * PR for supporting Spark 3.1. (.NET changes + build)

BTW, is this backward compatible? I see lots of filtered tests. We should make this backward compatible if possible, o.w., we should release 2.0.

imback82:
OK, I will check again with new PRs. So you are saying we are good to go with 1.1, right?

PR #833, opened 2021-03-09
Support Spark 3.0.2

suhsteve:
Add support for Spark 3.0.2  #827

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,12 +33,16 @@ variables:
 
   # Filter DataFrameTests.TestDataFrameGroupedMapUdf and DataFrameTests.TestGroupedMapUdf backwardCompatible
   # tests due to https://github.com/dotnet/spark/pull/711
-  backwardCompatibleTestOptions_Windows_3: "--filter \
+  backwardCompatibleTestOptions_Windows_3_0: "--filter \
   (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestDataFrameGroupedMapUdf)&\
   (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestGroupedMapUdf)"
-  forwardCompatibleTestOptions_Windows_3: ""
-  backwardCompatibleTestOptions_Linux_3: $(backwardCompatibleTestOptions_Windows_3)
-  forwardCompatibleTestOptions_Linux_3: $(forwardCompatibleTestOptions_Linux_2_4)
+  forwardCompatibleTestOptions_Windows_3_0: ""
+  backwardCompatibleTestOptions_Linux_3_0: $(backwardCompatibleTestOptions_Windows_3_0)
+  forwardCompatibleTestOptions_Linux_3_0: $(forwardCompatibleTestOptions_Linux_2_4)
+
+  # Skip all forwardCompatible tests, Spark 3.0.2 is not supported in microsoft-spark-3-0_2.12-1.0.0.jar
+  forwardCompatibleTestOptions_Windows_3_0_2: "--filter FullyQualifiedName=NONE"
+  forwardCompatibleTestOptions_Linux_3_0_2: $(forwardCompatibleTestOptions_Windows_3_0_2)
 
   # Azure DevOps variables are transformed into environment variables, with these variables we
   # avoid the first time experience and telemetry to speed up the build.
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -33,12 +33,16 @@ variables:
...
   (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestDataFrameGroupedMapUdf)&\
   (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestGroupedMapUdf)"
-  forwardCompatibleTestOptions_Windows_3: ""
-  backwardCompatibleTestOptions_Linux_3: $(backwardCompatibleTestOptions_Windows_3)
-  forwardCompatibleTestOptions_Linux_3: $(forwardCompatibleTestOptions_Linux_2_4)
+  forwardCompatibleTestOptions_Windows_3_0: ""
+  backwardCompatibleTestOptions_Linux_3_0: $(backwardCompatibleTestOptions_Windows_3_0)
+  forwardCompatibleTestOptions_Linux_3_0: $(forwardCompatibleTestOptions_Linux_2_4)
+
+  # Skip all forwardCompatible tests, Spark 3.0.2 is not supported in microsoft-spark-3-0_2.12-1.0.0.jar
```
imback82, line 43:
```suggestion
  # Skip all forward compatible tests since Spark 3.0.2 is not supported in microsoft-spark-3-0_2.12-1.0.0.jar
```

azure-pipelines.yml
```diff
@@ -33,12 +33,16 @@ variables:
 
   # Filter DataFrameTests.TestDataFrameGroupedMapUdf and DataFrameTests.TestGroupedMapUdf backwardCompatible
   # tests due to https://github.com/dotnet/spark/pull/711
-  backwardCompatibleTestOptions_Windows_3: "--filter \
+  backwardCompatibleTestOptions_Windows_3_0: "--filter \
   (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestDataFrameGroupedMapUdf)&\
   (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestGroupedMapUdf)"
```
imback82, line 38: Not related to this PR, but this will require 2.0 release? This never worked in 1.0: https://github.com/dotnet/spark/blob/master/docs/release-notes/1.0.0/release-1.0.0.md#known-issues, but safer to release 2.0? cc @rapoth

imback82 (commented):
LGTM (one nit), thanks @suhsteve!

azure-pipelines.yml
suhsteve: I think we should revisit the criteria for major version releases.  It'll may be difficult if the criteria, ensuring the Worker is always forward compatible with all versions sharing the same major version.  With the current design, this will break any time a new Spark feature (minor) version is released, as well as any time any internal contracts change that requires changes to the Worker.

imback82: https://semver.org/

imback82: I am more concerned with backward compatibility.

> this will break any time a new Spark feature (minor) version is released

not necessarily true? Also, feature release is 9 months to a year, so it should be ok even if we have to bump the major version.

suhsteve: GroupedMapUdf wasn't working in 1.0.0 release, but having it working in 1.x releases isn't breaking backward compatibility is it?  Seems more like a bugfix or adding a new feature?

imback82: yea, that's what I was asking before. so it seems OK to go with 1.1?

suhsteve: > > this will break any time a new Spark feature (minor) version is released
> 
> not necessarily true?

At the moment forward compat will break at every minor release, unless we want to be more lenient on the pattern matching in the processors.  ie, match 2.4+ and spark 3.x

https://github.com/dotnet/spark/blob/550835feeb667edd10e80886c61641b1dd392a4c/src/csharp/Microsoft.Spark.Worker/Processor/TaskContextProcessor.cs#L26

https://github.com/dotnet/spark/blob/550835feeb667edd10e80886c61641b1dd392a4c/src/csharp/Microsoft.Spark.Worker/Processor/CommandProcessor.cs#L102

suhsteve: > yea, that's what I was asking before. so it seems OK to go with 1.1?

Yeah I'm okay with that.  @rapoth

imback82: > At the moment forward compat will break at every minor release,

Yes. As mentioned above, I am more concerned with backward compatibility for versioning (as semantic versioning does).

imback82:
Can you update the PR description on what's being done so we can use this as a reference PR for a patch release?

PR #836, opened 2021-03-09
Support Spark 3.1.1

suhsteve:
Add support for Spark 3.1.1 #811

- Update pom.xml and add 3-1 module
- Skip [incompatible delta tests](https://github.com/delta-io/delta/issues/594)
- update azure-pipeline.xml and add 3.1.1 to pipeline
- update [DataFrameWriter](https://github.com/apache/spark/blob/bb9abb098a1e26c220546616d655380c479b0e42/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala#L289) and [DataStreamReader](https://github.com/apache/spark/blob/bb9abb098a1e26c220546616d655380c479b0e42/sql/core/src/main/scala/org/apache/spark/sql/DataFrameReader.scala#L259) test order due to change in settings with api behavior.
- Relaxed 3.x version check in `CommandProcessor`, `TaskContextProcessor`, `Broadcast`, `DataFrame`

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -44,6 +44,54 @@ variables:
   forwardCompatibleTestOptions_Windows_3_0_2: "--filter FullyQualifiedName=NONE"
   forwardCompatibleTestOptions_Linux_3_0_2: $(forwardCompatibleTestOptions_Windows_3_0_2)
 
+  # Skip backwardCompatible tests because Microsoft.Spark.Worker requires Spark 3.1 support in
+  # CommandProcessor.cs and TaskContextProcessor.cs. Support added in https://github.com/dotnet/spark/pull/835
+  backwardCompatibleTestOptions_Windows_3_1: "--filter \
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestDataFrameGroupedMapUdf)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.DataFrameTests.TestGroupedMapUdf&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfRegistrationWithReturnAsRowType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfWithArrayChain)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfWithSimpleArrayType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfWithMapType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfWithRowArrayType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfWithReturnAsMapType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfWithReturnAsArrayOfArrayType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfWithArrayOfArrayType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfWithMapOfMapType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfWithReturnAsSimpleArrayType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfWithRowType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfComplexTypesTests.TestUdfWithReturnAsRowType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfSerDeTests.TestExternalStaticMethodCall)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfSerDeTests.TestInitExternalClassInUdf)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfSerDeTests.TestUdfClosure)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfSimpleTypesTests.TestUdfWithReturnAsDateType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfSimpleTypesTests.TestUdfWithReturnAsTimestampType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfSimpleTypesTests.TestUdfWithDateType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.UdfTests.UdfSimpleTypesTests.TestUdfWithTimestampType)&\
+  (FullyQualifiedName!=Microsoft.Spark.E2ETest.IpcTests.BroadcastTests.TestMultipleBroadcast)&\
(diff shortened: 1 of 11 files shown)


src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameWriterTests.cs
```diff
@@ -85,6 +80,12 @@ public void TestSignaturesV2_3_X()
                 dfw.Text($"{tempDir.Path}TestTextPath");
 
                 dfw.Csv($"{tempDir.Path}TestCsvPath");
+
```
suhsteve, line 83: Setting `path` option and then calling save is not supported by default unless  `spark.sql.legacy.pathOptionBehavior.enabled` conf is set.  Reverse order in test.

https://github.com/apache/spark/blob/bb9abb098a1e26c220546616d655380c479b0e42/sql/core/src/main/scala/org/apache/spark/sql/DataFrameWriter.scala#L289

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamReaderTests.cs
```diff
@@ -63,6 +62,7 @@ public void TestSignaturesV2_3_X()
                 dsr.Parquet(Path.Combine(TestEnvironment.ResourceDirectory, "users.parquet")));
             Assert.IsType<DataFrame>
                 (dsr.Text(Path.Combine(TestEnvironment.ResourceDirectory, "people.txt")));
+            Assert.IsType<DataFrame>(dsr.Format("json").Option("path", jsonFilePath).Load());
```
suhsteve, line 65: Setting `path` option and then calling `load` is not supported by default unless  `spark.sql.legacy.pathOptionBehavior.enabled` conf is set.  Reverse order in test.

https://github.com/apache/spark/blob/bb9abb098a1e26c220546616d655380c479b0e42/sql/core/src/main/scala/org/apache/spark/sql/DataFrameReader.scala#L259

src/csharp/Extensions/Microsoft.Spark.Extensions.Delta.E2ETest/DeltaTableTests.cs
```diff
@@ -30,7 +30,7 @@ public DeltaTableTests(DeltaFixture fixture)
         /// Run the end-to-end scenario from the Delta Quickstart tutorial.
         /// </summary>
         /// <see cref="https://docs.delta.io/latest/quick-start.html"/>
-        [SkipIfSparkVersionIsLessThan(Versions.V2_4_2)]
+        [SkipIfSparkVersionIsNotInRange(Versions.V2_4_2, Versions.V3_1_1)]
```
suhsteve, line 33: Disable Delta tests that have code path that creates a `org.apache.spark.sql.catalyst.expressions.Alias` object.   `Delta 0.8.0` is not compatible with `Spark 3.1.1`  https://github.com/delta-io/delta/issues/594

azure-pipelines.yml
```diff
@@ -44,6 +44,54 @@ variables:
   forwardCompatibleTestOptions_Windows_3_0_2: "--filter FullyQualifiedName=NONE"
   forwardCompatibleTestOptions_Linux_3_0_2: $(forwardCompatibleTestOptions_Windows_3_0_2)
 
+  # Skip backwardCompatible tests because Microsoft.Spark.Worker requires Spark 3.1 support in
+  # CommandProcessor.cs and TaskContextProcessor.cs. Support added in https://github.com/dotnet/spark/pull/835
```
suhsteve, line 48:
```suggestion
  # CommandProcessor.cs and TaskContextProcessor.cs. Support added in https://github.com/dotnet/spark/pull/836
```

imback82:
Thanks for splitting. This seems like a more manageable PR now. I will take a look tomorrow.

azure-pipelines.yml
```diff
@@ -44,6 +44,54 @@ variables:
   forwardCompatibleTestOptions_Windows_3_0_2: "--filter FullyQualifiedName=NONE"
   forwardCompatibleTestOptions_Linux_3_0_2: $(forwardCompatibleTestOptions_Windows_3_0_2)
 
+  # Skip backwardCompatible tests because Microsoft.Spark.Worker requires Spark 3.1 support in
```
imback82, line 47: So it seems that we need 2.0?

Assuming we release 1.1, if the client is using 1.1 for Spark 3.1, and the worker installed is 1.0, things will break. Right?

suhsteve: Yes, the 1.0.0 Worker will break any time we introduce a version that supports a new minor version of Spark.

src/csharp/Microsoft.Spark.Worker/Processor/CommandProcessor.cs
```diff
@@ -100,6 +100,7 @@ private static RDDCommand ReadRDDCommand(Stream stream)
                 (2, 3) => SqlCommandProcessorV2_3_X.Process(evalType, stream),
                 (2, 4) => SqlCommandProcessorV2_4_X.Process(evalType, stream),
                 (3, 0) => SqlCommandProcessorV2_4_X.Process(evalType, stream),
+                (3, 1) => SqlCommandProcessorV2_4_X.Process(evalType, stream),
```
imback82, line 103: Can we do `(3, _)` so that this doesn't force us to do a major version bump for a Spark minor version bump?

azure-pipelines.yml
Niharikadutta: Quick question, why is the above ^ a must? If the Spark minor version doesn't involve changes to how stream is sent, can't we just do what Terry suggests [here](https://github.com/dotnet/spark/pull/836/files/91be8f4ceaadd06f1e2c403ee46cb307732f5bd5#r593972415)?

suhsteve: No one said it's a must, it's just the current pattern we have it set up with.  The suggestion that Terry brought up was already discussed [here](https://github.com/dotnet/spark/pull/833#discussion_r589985702)

src/csharp/Microsoft.Spark.Worker/Processor/CommandProcessor.cs
suhsteve: We definitely can.  If we want to be more lenient and remove these artificial version checking gates, should we also revisit the version check we've been adding to the `jar`s ?  In the jar, instead of validating against supported versions, do the reverse and check against unsupported versions, this way users won't have to update the jar every time a patch version is released and they want to use the same jar.

imback82: JARs seem fine to me since they don't affect the backward compatibility. Since patch releases are >6 months apart anyway, I am fine with releasing a new version after the verification is done. But, I am open to this idea if there are strong reasons.

src/csharp/Extensions/Microsoft.Spark.Extensions.Delta.E2ETest/DeltaTableTests.cs
imback82: I think you should add this comment to the code?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameWriterTests.cs
imback82: same here. i would add this to the code.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/Streaming/DataStreamReaderTests.cs
imback82: ditto.

imback82 (commented):
LGTM (pending minor comments). We need to rework on azure-pipelines.yml before releasing 2.0, but this PR itself is good. Thanks @suhsteve!

imback82 (approved):

src/csharp/Microsoft.Spark.Worker/Processor/CommandProcessor.cs
imback82: https://github.com/dotnet/spark/discussions/846

I think we can introduce a config to bypass the check? The user can use the config on his/her own risk. WDYT?

suhsteve: I'm okay with an opt-in approach to bypassing the checks.  Is this only in regards to the jar or are there other places you had in mind (like in checks modified in this PR).

imback82: Just JARs should be enough for now.

imback82: BTW, this will only bypass checks on the Spark's patch version. We still need to check the minor version.

PR #861, opened 2021-03-22
Add DOTNET_WORKER_<ver>_DIR environment variable

suhsteve:
Update how the `ConfigurationService.GetWorkerExePath` is constructed.

Add a new environment variable, `DOTNET_WORKER_<ver>_DIR` where `ver` can be any of the following (built using the version of the `Microsoft.Spark` assembly):
- major_minor_build
- major_minor
- major

The above listing will take precedence when constructing the worker path (higher item has more precedence).  If it is not set, then fall back to original behavior (use `DOTNET_WORKER_DIR` if it is set, otherwise return the worker process name and expect it to exist in `PATH`).

### DIFF
--- a/src/csharp/Microsoft.Spark.UnitTest/ConfigurationServiceTests.cs
+++ b/src/csharp/Microsoft.Spark.UnitTest/ConfigurationServiceTests.cs
@@ -48,7 +48,9 @@ namespace Microsoft.Spark.UnitTest
             Environment.SetEnvironmentVariable(ConfigurationService.WorkerDirEnvVarName, workerDir);
 
             // Only ConfigurationService.WorkerDirEnvVarName is set, WorkerExePath will be built using it.
-            Assert.Equal(Path.Combine(workerDir, ConfigurationService.ProcFileName), configService.GetWorkerExePath());
+            Assert.Equal(
+                Path.Combine(workerDir, ConfigurationService.ProcFileName),
+                configService.GetWorkerExePath());
         }
 
         [Fact]
@@ -63,7 +65,9 @@ namespace Microsoft.Spark.UnitTest
             // ConfigurationService.WorkerDirEnvVarName and ConfigurationService.WorkerVerDirEnvVarName
             // environment variables are set. ConfigurationService.WorkerVerDirEnvVarName will take
             // precedence.
-            Assert.Equal(Path.Combine(workerVerDir, ConfigurationService.ProcFileName), configService.GetWorkerExePath());
+            Assert.Equal(
+                Path.Combine(workerVerDir, ConfigurationService.ProcFileName),
+                configService.GetWorkerExePath());
         }
 
         public void Dispose()
--- a/src/csharp/Microsoft.Spark.UnitTest/ConfigurationServiceTests.cs
+++ b/src/csharp/Microsoft.Spark.UnitTest/ConfigurationServiceTests.cs
@@ -76,6 +76,9 @@ namespace Microsoft.Spark.UnitTest
                 Assert.True(
                     string.IsNullOrWhiteSpace(
                         Environment.GetEnvironmentVariable(_workerDirEnvVars.WorkerMajorDir.Name)));
+                Assert.True(
+                    string.IsNullOrWhiteSpace(
+                        Environment.GetEnvironmentVariable(_workerDirEnvVars.WorkerDir.Name)));
                 Assert.Equal(
                     ConfigurationService.DefaultWorkerDirEnvVarName, configService.WorkerDirEnvVarName);
             }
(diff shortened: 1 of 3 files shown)


src/csharp/Microsoft.Spark/Services/ConfigurationService.cs
```diff
@@ -21,31 +21,45 @@ internal sealed class ConfigurationService : IConfigurationService
...
 
         private string _workerPath;
 
+        /// <summary>
+        /// The Microsoft.Spark major assembly version is used to construct the
+        /// WorkerVerDirEnvVarNameFormat environment variable.
+        /// </summary>
+        internal static string WorkerVerDirEnvVarName { get; } =
+            string.Format(
+                WorkerVerDirEnvVarNameFormat,
```
imback82, line 45: Just to be future proof, can we support all three levels? So we will try to go from patch version, minor version, then major version, then fall back to the defaut.

suhsteve: Sure.   Which do you prefer?

```
DOTNET_WORKER_V1_0_0_DIR
DOTNET_WORKER_V1_0_DIR
DOTNET_WORKER_V1_DIR

DOTNET_WORKER_1_0_DIR
DOTNET_WORKER_1_0_0_DIR
DOTNET_WORKER_1_DIR
```

suhsteve: Using the naming convention used in the lower half.

imback82: Sgtm

src/csharp/Microsoft.Spark.UnitTest/ConfigurationServiceTests.cs
```diff
@@ -0,0 +1,184 @@
...
+                Path.Combine(workerMajorDir, ConfigurationService.ProcFileName),
+                configService.GetWorkerExePath());
+        }
+
+        public void Dispose()
+        {
+            Environment.SetEnvironmentVariable(
+                _workerDirEnvVars.WorkerDir.Name,
+                _workerDirEnvVars.WorkerDir.Value);
+            Environment.SetEnvironmentVariable(
```
imback82, line 151: Don't we need to unset others and just leave the default one (we should save the previous value in the constructor and use it?), if I understand this correctly?

src/csharp/Microsoft.Spark.UnitTest/ConfigurationServiceTests.cs
```diff
@@ -0,0 +1,184 @@
...
+            Assert.Equal(
+                Path.Combine(workerDir, ConfigurationService.ProcFileName),
+                configService.GetWorkerExePath());
+        }
+
+        [Fact]
+        public void TestWorkerExePathWithEnvVarPrecedence()
+        {
+            var configService = new ConfigurationService();
+            var workerDir = "workerDir";
```
imback82, line 128: super it: I think this should be `string`. same in other places.

src/csharp/Microsoft.Spark.UnitTest/ConfigurationServiceTests.cs
```diff
@@ -0,0 +1,184 @@
...
+            Assert.True(_workerDirEnvVars.WorkerDir.IsSet());
+
+            // Only WorkerDir is set, WorkerExePath will be built using it.
+            Assert.Equal(
+                Path.Combine(workerDir, ConfigurationService.ProcFileName),
+                configService.GetWorkerExePath());
+        }
+
+        [Fact]
+        public void TestWorkerExePathWithEnvVarPrecedence()
```
imback82, line 125: This could have been just tested with `TestWorkerEnvName`, no? Any particular reason this is tied to `GetWorkerExePath`?

imback82 (commented):
LGTM (minor comments/questions), thanks @suhsteve!

src/csharp/Microsoft.Spark.UnitTest/ConfigurationServiceTests.cs
suhsteve: I'm saving all previously set env var names and values in the constructor, and then unsetting all the values.  In the tests I'm setting the env vars, doing validation, and then in dispose the original env var values will be restored.

src/csharp/Microsoft.Spark.UnitTest/ConfigurationServiceTests.cs
suhsteve: Sure I can combine the tests and just test `GetWorkerExePath` instead.

imback82 (approved):

PR #862, opened 2021-03-22
Add spark.dotnet.ignoreSparkPatchVersionCheck conf to ignore patch version in DotnetRunner

suhsteve:
Add a new spark conf, `spark.dotnet.ignoreSparkPatchVersionCheck`.  If set to true, then only check the major.minor `SPARK_VERSION` and ignore patch version.  If set to false, then check the `SPARK_VERSION` against the set of `supportedSparkVersions`

### DIFF
--- a/src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
+++ b/src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
@@ -19,7 +19,6 @@ import org.apache.spark
 import org.apache.spark.api.dotnet.DotnetBackend
 import org.apache.spark.deploy.{PythonRunner, SparkHadoopUtil}
 import org.apache.spark.internal.Logging
-import org.apache.spark.internal.config.dotnet.Dotnet.DOTNET_IGNORE_SPARK_PATCH_VERSION_CHECK
 import org.apache.spark.util.dotnet.{Utils => DotnetUtils}
 import org.apache.spark.util.{RedirectThread, Utils}
 import org.apache.spark.{SecurityManager, SparkConf, SparkEnv, SparkUserAppException}
@@ -35,18 +34,19 @@ import scala.util.Try
  */
 object DotnetRunner extends Logging {
   private val DEBUG_PORT = 5567
-  private val supportedSparkMajorMinorVersion = List("2", "3")
+  private val sparkMajorMinorVersionPrefix = "2.3."
   private val supportedSparkVersions = Set[String]("2.3.0", "2.3.1", "2.3.2", "2.3.3", "2.3.4")
-
-  val SPARK_VERSION = DotnetUtils.normalizeSparkVersion(spark.SPARK_VERSION)
 
   def main(args: Array[String]): Unit = {
     if (args.length == 0) {
       throw new IllegalArgumentException("At least one argument is expected.")
     }
 
-    val conf = Option(SparkEnv.get).map(_.conf).getOrElse(new SparkConf())
-    validateSparkVersions(conf)
+    DotnetUtils.validateSparkVersions(
+      Option(SparkEnv.get).map(_.conf).getOrElse(new SparkConf()),
+      spark.SPARK_VERSION,
+      sparkMajorMinorVersionPrefix,
+      supportedSparkVersions)
 
     val settings = initializeSettings(args)
 
@@ -164,21 +164,6 @@ object DotnetRunner extends Logging {
     } else {
       logError(s"DotnetBackend did not initialize in $backendTimeout seconds")
       DotnetUtils.exit(-1)
-    }
-  }
-
-  private def validateSparkVersions(conf: SparkConf): Unit = {
-    val ignorePatchVersion = conf.get(DOTNET_IGNORE_SPARK_PATCH_VERSION_CHECK)
-    val majorMinorVersion = supportedSparkMajorMinorVersion.mkString("", ".", ".")
-    if (!SPARK_VERSION.startsWith(majorMinorVersion)) {
-      throw new IllegalArgumentException(
-        s"Unsupported spark version used: ${spark.SPARK_VERSION}. Normalized spark version used: $SPARK_VERSION." +
-          s" Supported spark major.minor version: $majorMinorVersion")
-    } else if (!ignorePatchVersion && !supportedSparkVersions(SPARK_VERSION)) {
-      val supportedVersions = supportedSparkVersions.toSeq.sorted.mkString(", ")
-      throw new IllegalArgumentException(
-        s"Unsupported spark version used: ${spark.SPARK_VERSION}. Normalized spark version used: $SPARK_VERSION." +
-          s" Supported versions: $supportedVersions")
     }
   }
 
--- a/src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
+++ b/src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
@@ -34,7 +34,7 @@ import scala.util.Try
  */
 object DotnetRunner extends Logging {
   private val DEBUG_PORT = 5567
-  private val sparkMajorMinorVersionPrefix = "2.3."
+  private val supportedSparkMajorMinorVersionPrefix = "2.3"
   private val supportedSparkVersions = Set[String]("2.3.0", "2.3.1", "2.3.2", "2.3.3", "2.3.4")
 
   val SPARK_VERSION = DotnetUtils.normalizeSparkVersion(spark.SPARK_VERSION)
@@ -48,7 +48,7 @@ object DotnetRunner extends Logging {
       Option(SparkEnv.get).map(_.conf).getOrElse(new SparkConf()),
       spark.SPARK_VERSION,
       SPARK_VERSION,
-      sparkMajorMinorVersionPrefix,
+      supportedSparkMajorMinorVersionPrefix,
       supportedSparkVersions)
 
     val settings = initializeSettings(args)
--- a/src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
+++ b/src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
@@ -19,9 +19,10 @@ import org.apache.spark
 import org.apache.spark.api.dotnet.DotnetBackend
 import org.apache.spark.deploy.{PythonRunner, SparkHadoopUtil}
 import org.apache.spark.internal.Logging
+import org.apache.spark.internal.config.dotnet.Dotnet.DOTNET_IGNORE_SPARK_PATCH_VERSION_CHECK
 import org.apache.spark.util.dotnet.{Utils => DotnetUtils}
 import org.apache.spark.util.{RedirectThread, Utils}
-import org.apache.spark.{SecurityManager, SparkConf, SparkUserAppException}
+import org.apache.spark.{SecurityManager, SparkConf, SparkEnv, SparkUserAppException}
 
 import scala.collection.JavaConverters._
 import scala.io.StdIn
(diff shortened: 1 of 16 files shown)


src/scala/microsoft-spark-3-0/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
```diff
@@ -164,12 +167,18 @@ object DotnetRunner extends Logging {
     }
   }
 
-  private def validateSparkVersions: Unit = {
-    if (!supportedSparkVersions(SPARK_VERSION)) {
+  private def validateSparkVersions(conf: SparkConf): Unit = {
+    val ignorePatchVersion = conf.get(DOTNET_IGNORE_SPARK_PATCH_VERSION_CHECK)
+    val majorMinorVersion = supportedSparkMajorMinorVersion.mkString("", ".", ".")
```
imback82, line 172: just store it as a string to avoid `mkString`?

src/scala/microsoft-spark-3-0/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
```diff
@@ -164,12 +167,18 @@ object DotnetRunner extends Logging {
...
+        s"Unsupported spark version used: ${spark.SPARK_VERSION}. Normalized spark version used: $SPARK_VERSION." +
+          s" Supported spark major.minor version: $majorMinorVersion")
+    } else if (!ignorePatchVersion && !supportedSparkVersions(SPARK_VERSION)) {
       val supportedVersions = supportedSparkVersions.toSeq.sorted.mkString(", ")
       throw new IllegalArgumentException(
-        s"Unsupported spark version used: ${spark.SPARK_VERSION}. Normalized spark version used: ${SPARK_VERSION}." +
-          s" Supported versions: ${supportedVersions}")
+        s"Unsupported spark version used: ${spark.SPARK_VERSION}. Normalized spark version used: $SPARK_VERSION." +
+          s" Supported versions: $supportedVersions")
     }
```
imback82, line 182: Can we add warning message that it's being bypassed?

imback82 (commented):
I assume you tested this manually? How about just put this in Utils, so that we can unit-test this?

src/scala/microsoft-spark-3-0/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
suhsteve: Sure.  Do you want me to hardcode it in this method or set it as a class field at the top of the class definition ?

suhsteve:
> I assume you tested this manually? How about just put this in Utils, so that we can unit-test this?

Yup, was manually tested.  I can move to Utils.

src/scala/microsoft-spark-3-0/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
imback82: Class field seems fine to me.

suhsteve:
> > I assume you tested this manually? How about just put this in Utils, so that we can unit-test this?
> 
> Yup, was manually tested. I can move to Utils.

Moved to Utils and added tests.

imback82:
Can you check the test failures?

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/util/dotnet/UtilsTest.scala
```diff
@@ -0,0 +1,80 @@
...
+  def shouldThrowForUnsupportedMajorMinorVersion(): Unit = {
+    val conf = new SparkConf()
+    conf.set(DOTNET_IGNORE_SPARK_PATCH_VERSION_CHECK, false)
+
+    val sparkVersion = "2.4.4"
+    val sparkMajorMinorVersionPrefix = "2.3."
+    val supportedSparkVersions = Set[String]("2.3.0", "2.3.1", "2.3.2", "2.3.3", "2.3.4")
+
+    assertThrows(
+      classOf[IllegalArgumentException],
```
imback82, line 68: Can you use `intercept` and check the message as well (since we have two different places where we throw exceptions)? https://github.com/microsoft/hyperspace/blob/053b6b1e00f2b6dd62889e40d417dff1939079fc/src/test/scala/com/microsoft/hyperspace/index/CreateIndexTest.scala#L76

imback82: Or we don't have `scalatest` set up?

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
```diff
@@ -43,7 +44,12 @@ object DotnetRunner extends Logging {
       throw new IllegalArgumentException("At least one argument is expected.")
     }
 
-    validateSparkVersions
+    DotnetUtils.validateSparkVersions(
+      Option(SparkEnv.get).map(_.conf).getOrElse(new SparkConf()),
```
imback82, line 48: question. this SparkEnv will be set for `spark-submit` and livy submission right?

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/util/dotnet/UtilsTest.scala
imback82: If `scalatest` is not possible, can you grep the exception message and make sure it's thrown at a different place?

imback82 (commented):
LGTM (few minor comments), thanks @suhsteve!

src/scala/microsoft-spark-2-3/src/test/scala/org/apache/spark/util/dotnet/UtilsTest.scala
suhsteve: We don't have scalatest set up, only junit.  But we can still check exception message with it.

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/util/dotnet/Utils.scala
```diff
@@ -191,6 +193,45 @@ object Utils extends Logging {
...
+  /**
+   * Validates the normalized spark version by verifying:
+   *   - Spark version starts with sparkMajorMinorVersionPrefix.
+   *   - If ignoreSparkPatchVersionCheck spark conf is
+   *     - true: valid
+   *     - false: check if the spark version is in supportedSparkVersions.
+   * @param conf Spark conf.
+   * @param sparkVersion The spark version.
+   * @param normalizedSparkVersion: The normalized spark version.
+   * @param sparkMajorMinorVersionPrefix The spark major and minor version to validate against.
```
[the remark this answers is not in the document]
suhsteve: updated.

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/util/dotnet/Utils.scala
```diff
@@ -191,6 +193,45 @@ object Utils extends Logging {
...
+    if (!normalizedSparkVersion.startsWith(sparkMajorMinorVersionPrefix)) {
+      throw new IllegalArgumentException(
+        s"Unsupported spark version used: $sparkVersion. " +
+          s"Normalized spark version used: $normalizedSparkVersion. " +
+          s"Supported spark major.minor version: $sparkMajorMinorVersionPrefix")
+    } else if (ignorePatchVersion) {
+      logWarning(
+        s"Ignoring spark patch version. Spark version used: $sparkVersion. " +
+          s"Normalized spark version used: $normalizedSparkVersion. " +
+          s"Spark major.minor prefix used: $sparkMajorMinorVersionPrefix")
```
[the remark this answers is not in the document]
suhsteve: Removed trailing `.` from var definition.

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
suhsteve: I know it works for `spark-submit`, not sure about livy submission.

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/util/dotnet/Utils.scala
```diff
@@ -191,6 +193,45 @@ object Utils extends Logging {
...
+      normalizedSparkVersion: String,
+      supportedSparkMajorMinorVersionPrefix: String,
+      supportedSparkVersions: Set[String]): Unit = {
+    val ignorePatchVersion = conf.get(DOTNET_IGNORE_SPARK_PATCH_VERSION_CHECK)
+
+    if (!normalizedSparkVersion.startsWith(s"$supportedSparkMajorMinorVersionPrefix.")) {
+      throw new IllegalArgumentException(
+        s"Unsupported spark version used: $sparkVersion. " +
+          s"Normalized spark version used: $normalizedSparkVersion. " +
+          s"Supported spark major.minor version: $supportedSparkMajorMinorVersionPrefix")
```
imback82, line 220: Let's quote the version string so that there is no confusion with dots in the version.
```suggestion
        s"Unsupported spark version used: '$sparkVersion'. " +
          s"Normalized spark version used: '$normalizedSparkVersion'. " +
          s"Supported spark major.minor version: '$supportedSparkMajorMinorVersionPrefix'")
```

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
imback82: Hmm, to me `SparkEnv.get` should always return None since it's set in during SparkContext initialization? I think the reason this works is because of
https://github.com/apache/spark/blob/06d40696dc27c39824666a70e7c673880959a681/core/src/main/scala/org/apache/spark/SparkConf.scala#L70

And
https://github.com/apache/spark/blob/06d40696dc27c39824666a70e7c673880959a681/core/src/main/scala/org/apache/spark/deploy/SparkApplication.scala#L47

Were you getting `Some` from `SparkEnv.get`?

suhsteve: `SparkEnv.get` is `None`.  We can pass in a `new SparkConf` directly or just get the values from `sys.props` instead if you prefer.

imback82: If it's always None, I would remove it since it just adds confusion and document clearly. For example:
```scala
// Load defaults set in the system properties when this application is launched:
// https://github.com/apache/spark/blob/06d40696dc27c39824666a70e7c673880959a681/core/src/main/scala/org/apache/spark/deploy/SparkApplication.scala#L47
val sparkConf = new SparkConf(loadDefaults = true)
```

suhsteve: I opted to go with the `sys.props` since [RRunner.scala](https://github.com/apache/spark/blob/0494dc90af48ce7da0625485a4dc6917a244d580/core/src/main/scala/org/apache/spark/deploy/RRunner.scala#L48) is also getting the conf this way.

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/util/dotnet/Utils.scala
suhsteve: Added quotes.

src/scala/microsoft-spark-2-3/src/main/scala/org/apache/spark/deploy/dotnet/DotnetRunner.scala
imback82: OK, either one is fine as long as `SparkEnv.get` is removed.

imback82 (approved):

PR #868, opened 2021-03-26
Bug Fix for Spark 3.x - Avoid converting converted Row values

### LINKED ISSUE #760: [BUG]: Applying UDFs to TimestampType causes occasional exception

suhsteve:
There has been no modification to the `RowPickler` code between [EvaluatePython.scala (Spark 2.4.7)](https://github.com/apache/spark/blob/14211a19f53bd0f413396582c8970e3e0a74281d/sql/core/src/main/scala/org/apache/spark/sql/execution/python/EvaluatePython.scala#L253) and [EvaluatePython.scala (Spark 3.0.0)](https://github.com/apache/spark/blob/3fdfce3120f307147244e5eaf46d61419a723d50/sql/core/src/main/scala/org/apache/spark/sql/execution/python/EvaluatePython.scala#L253).

`Spark 2.4.7` used pyrolite `4.13` and starting with  `Spark 3.0.0` pyrolite was updated to `4.30`.  In `RowPickler`, Spark pickles the row values using:
```scala
        while (i < row.values.length) {
          pickler.save(row.values(i))
          i += 1
        }
```

In `pickler.save(Object)` Pyrolite checks whether the object has been `memoized`, and if it hasn't it will process the object and pickle it.  There was a [PR](https://github.com/irmen/Pyrolite/pull/56) in Nov 2017 (between the 4.13 and 4.30 releases) that updated the logic with how the `memoize` check was done.  [Pyrolite `4.13` checked the `System.identityHashCode(obj)`](https://github.com/irmen/Pyrolite/blob/55941dbf5d8e03381a5393a190062ca4447e21d0/java/src/main/java/net/razorvine/pickle/Pickler.java#L182), however [pyrolite `4.30` only checks the `obj.hashCode()` ](https://github.com/irmen/Pyrolite/blob/pyrolite-4.30/java/src/main/java/net/razorvine/pickle/Pickler.java), which is the default behavior, unless the [`valueCompare` flag](https://github.com/irmen/Pyrolite/blob/d4fafc3119f5e75387b0eff84104b47a9e7aea39/java/src/main/java/net/razorvine/pickle/Pickler.java#L89) has been toggled.  Toggling this flag would go back to the `4.13` behavior.  Spark 3.x, however, does not use the [Pickler constructor](https://github.com/irmen/Pyrolite/blob/d4fafc3119f5e75387b0eff84104b47a9e7aea39/java/src/main/java/net/razorvine/pickle/Pickler.java#L108) to set this.

I have refactored the `RowConstructor` class a bit to make it easier to understand as well as fixed the issue with converting already converted row values.

Fixes #760

### DIFF
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
@@ -24,59 +24,89 @@ namespace Microsoft.Spark.E2ETest.IpcTests
         [Fact]
         public void TestWithDuplicatedRows()
         {
+            var timestamp = new Timestamp(2020, 1, 1, 0, 0, 0, 0);
             var schema = new StructType(new StructField[]
             {
                 new StructField("ts", new TimestampType())
             });
-
             var data = new GenericRow[]
             {
-                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) })
+                new GenericRow(new object[] { timestamp })
             };
 
             DataFrame df = _spark.CreateDataFrame(data, schema);
+            Row[] rows = df
+                .WithColumn("tsRow", Struct("ts"))
+                .WithColumn("tsRowRow", Struct("tsRow"))
+                .Collect()
+                .ToArray();
 
-            df.WithColumn("tsRow", Struct("ts")).WithColumn("tsRowRow", Struct("tsRow")).Collect().ToArray();
+            Assert.Single(rows);
+
+            Row row = rows[0];
+            Assert.Equal(3, row.Values.Length);
+            Assert.Equal(timestamp, row.Values[0]);
+
+            Row tsRow = row.Values[1] as Row;
+            Assert.Single(tsRow.Values);
+            Assert.Equal(timestamp, tsRow.Values[0]);
+
+            Row tsRowRow = row.Values[2] as Row;
+            Assert.Single(tsRowRow.Values);
+            Assert.Equal(tsRowRow.Values[0], tsRow);
         }
 
         [Fact]
         public void TestWithDuplicateTimestamps()
         {
+            var timestamp = new Timestamp(2020, 1, 1, 0, 0, 0, 0);
             var schema = new StructType(new StructField[]
             {
                 new StructField("ts", new TimestampType())
             });
-
             var data = new GenericRow[]
             {
-                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) }),
-                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) }),
-                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) })
+                new GenericRow(new object[] { timestamp }),
+                new GenericRow(new object[] { timestamp }),
+                new GenericRow(new object[] { timestamp })
             };
 
             DataFrame df = _spark.CreateDataFrame(data, schema);
+            Row[] rows = df.Collect().ToArray();
 
-            df.Collect().ToArray();
+            Assert.Equal(3, rows.Length);
+            foreach (Row row in rows)
+            {
+                Assert.Single(row.Values);
+                Assert.Equal(timestamp, row.GetAs<Timestamp>(0));
+            }
         }
 
         [Fact]
         public void TestWithDuplicateDates()
         {
+            var date = new Date(2020, 1, 1);
             var schema = new StructType(new StructField[]
             {
                 new StructField("date", new DateType())
             });
-
             var data = new GenericRow[]
             {
-                new GenericRow(new object[] { new Date(2020, 1, 1) }),
-                new GenericRow(new object[] { new Date(2020, 1, 1) }),
-                new GenericRow(new object[] { new Date(2020, 1, 1) })
+                new GenericRow(new object[] { date }),
+                new GenericRow(new object[] { date }),
+                new GenericRow(new object[] { date })
             };
 
             DataFrame df = _spark.CreateDataFrame(data, schema);
 
-            df.Collect().ToArray();
+            Row[] rows = df.Collect().ToArray();
+
+            Assert.Equal(3, rows.Length);
+            foreach (Row row in rows)
+            {
+                Assert.Single(row.Values);
+                Assert.Equal(date, row.GetAs<Date>(0));
+            }
         }
     }
 }
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
@@ -1,0 +1,82 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Linq;
+using Microsoft.Spark.Sql;
+using Microsoft.Spark.Sql.Types;
+using Xunit;
+using static Microsoft.Spark.Sql.Functions;
+
+namespace Microsoft.Spark.E2ETest.IpcTests
+
+{
+    [Collection("Spark E2E Tests")]
+    public class RowTests
+    {
+        private readonly SparkSession _spark;
+
+        public RowTests(SparkFixture fixture)
+        {
+            _spark = fixture.Spark;
+        }
+
+        [Fact]
+        public void TestWithDuplicatedRows()
+        {
+            var schema = new StructType(new StructField[]
+            {
+                new StructField("ts", new TimestampType())
+            });
+
+            var data = new GenericRow[]
+            {
+                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) })
+            };
+
+            DataFrame df = _spark.CreateDataFrame(data, schema);
+
+            df.WithColumn("tsRow", Struct("ts")).WithColumn("tsRowRow", Struct("tsRow")).Collect().ToArray();
+        }
+
+        [Fact]
+        public void TestWithDuplicateTimestamps()
+        {
+            var schema = new StructType(new StructField[]
+            {
+                new StructField("ts", new TimestampType())
+            });
+
+            var data = new GenericRow[]
+            {
+                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) }),
+                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) }),
+                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) })
+            };
+
+            DataFrame df = _spark.CreateDataFrame(data, schema);
+
+            df.Collect().ToArray();
+        }
+
+        [Fact]
+        public void TestWithDuplicateDates()
+        {
+            var schema = new StructType(new StructField[]
+            {
+                new StructField("date", new DateType())
+            });
+
+            var data = new GenericRow[]
+            {
+                new GenericRow(new object[] { new Date(2020, 1, 1) }),
+                new GenericRow(new object[] { new Date(2020, 1, 1) }),
+                new GenericRow(new object[] { new Date(2020, 1, 1) })
+            };
+
+            DataFrame df = _spark.CreateDataFrame(data, schema);
+
+            df.Collect().ToArray();
+        }
+    }
+}
--- a/src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
+++ b/src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
@@ -143,9 +143,9 @@ namespace Microsoft.Spark.Worker.Command
                         // The following can happen if an UDF takes Row object(s).
                         // The JVM Spark side sends a Row object that wraps all the columns used
                         // in the UDF, thus, it is normalized below (the extra layer is removed).
-                        if (row is RowConstructor rowConstructor)
+                        if (row is Row r)
                         {
-                            row = rowConstructor.GetRow().Values;
+                            row = r.Values;
                         }
 
                         // Split id is not used for SQL UDFs, so 0 is passed.
--- a/src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
+++ b/src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
@@ -28,7 +28,11 @@ namespace Microsoft.Spark.Sql
         private static IDictionary<string, StructType> s_schemaCache;
 
         /// <summary>
-        /// Used by Unpickler to pass unpickled schema for handling.
+        /// Used by Unpickler to pass unpickled schema for handling. The Unpickler
+        /// will reuse the <see cref="RowConstructorConstructor"/> object when
+        /// it needs to start constructing a <see cref="Row"/>. The schema is passed
+        /// to <see cref="construct(object[])"/> and the returned
+        /// <see cref="IObjectConstructor"/> is used to build the rest of the <see cref="Row"/>.
         /// </summary>
         /// <param name="args">Unpickled schema</param>
         /// <returns>New RowConstructor object capturing the schema.</returns>
--- a/src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
+++ b/src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using Microsoft.Spark.Sql.Types;
 using Razorvine.Pickle;
 
@@ -38,6 +39,7 @@ namespace Microsoft.Spark.Sql
                 s_schemaCache = new Dictionary<string, StructType>();
             }
 
+            Debug.Assert((args != null) && (args.Length == 1) && (args[0] is string));
             return new RowConstructor(GetSchema((string)args[0]));
         }
 
@@ -58,6 +60,7 @@ namespace Microsoft.Spark.Sql
 
         private static StructType GetSchema(string schemaString)
         {
+            Debug.Assert(s_schemaCache != null);
             if (!s_schemaCache.TryGetValue(schemaString, out StructType schema))
             {
                 schema = (StructType)DataType.ParseDataType(schemaString);
@@ -68,6 +71,10 @@ namespace Microsoft.Spark.Sql
         }
     }
 
+    /// <summary>
+    /// Created from <see cref="RowConstructorConstructor"/> and subsequently used
+    /// by the Unpickler to construct a <see cref="Row"/>.
+    /// </summary>
     internal sealed class RowConstructor : IObjectConstructor
     {
         private readonly StructType _schema;
--- a/src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
+++ b/src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
@@ -11,16 +11,16 @@ using Razorvine.Pickle;
 namespace Microsoft.Spark.Sql
 {
     /// <summary>
-    /// RowConstructorConstructor is a custom unpickler for GenericRowWithSchema in Spark.
+    /// RowConstructor is a custom unpickler for GenericRowWithSchema in Spark.
     /// Refer to spark/sql/core/src/main/scala/org/apache/spark/sql/execution/python/
     /// EvaluatePython.scala how GenericRowWithSchema is being pickled.
     /// </summary>
-    internal sealed class RowConstructorConstructor : IObjectConstructor
+    internal sealed class RowConstructor : IObjectConstructor
     {
         /// <summary>
         /// Cache the schemas of the rows being received. Multiple schemas may be
         /// sent per batch if there are nested rows contained in the row. Note that
-        /// this is thread local variable because one RowConstructorConstructor object is
+        /// this is thread local variable because one RowConstructor object is
         /// registered to the Unpickler and there could be multiple threads unpickling
         /// the data using the same registered object.
         /// </summary>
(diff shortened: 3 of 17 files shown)


imback82:
@suhsteve Can you check the test failures if they are related?

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -143,9 +143,9 @@ internal class PicklingSqlCommandExecutor : SqlCommandExecutor
                         // The following can happen if an UDF takes Row object(s).
                         // The JVM Spark side sends a Row object that wraps all the columns used
                         // in the UDF, thus, it is normalized below (the extra layer is removed).
```
imback82, line 145: Is this comment still relevant?

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
```diff
@@ -119,16 +56,8 @@ internal void Reset()
...
 
-        /// <summary>
-        /// Get or cache the schema string contained in args. Calling this
-        /// is only valid if the child args contain the row values.
-        /// </summary>
-        /// <returns></returns>
-        private StructType GetSchema()
+        private static StructType GetSchema(string schemaString)
         {
-            Debug.Assert(s_schemaCache != null);
```
imback82, line 129: We can still have this check (this tells you the prerequisite clearly)

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
```diff
@@ -119,16 +56,8 @@ internal void Reset()
...
-            Debug.Assert((_args != null) && (_args.Length == 1) && (_args[0] is string));
```
imback82, line 130: Shall we move this assert to above `construct`?

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
```diff
@@ -4,112 +4,49 @@
...
+        /// <param name="args">Unpickled schema</param>
+        /// <returns>New RowConstructor object capturing the schema.</returns>
         public object construct(object[] args)
         {
-            // Every first call to construct() contains the schema data. When
-            // a new RowConstructor object is returned from this function,
-            // construct() is called on the returned object with the actual
-            // row data. The original RowConstructor object may be reused by the
-            // Unpickler and each subsequent construct() call can contain the
-            // schema data or a RowConstructor object that contains row data.
```
imback82, line 62: this comment is not relevant any more? (If not, can you update instead of removing?)

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
```diff
@@ -4,112 +4,49 @@
...
             if (s_schemaCache is null)
             {
                 s_schemaCache = new Dictionary<string, StructType>();
             }
 
-            // Return a new RowConstructor where the args either represent the
```
imback82, line 68: ditto on the comment. Let's at least explain what's happening since it's not straightforward.

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
```diff
@@ -138,4 +67,21 @@ private StructType GetSchema()
             return schema;
         }
     }
+
+    internal sealed class RowConstructor : IObjectConstructor
```
imback82, line 71: Can you add some high-level comment for this class. The relationship to `RowConstructorConstructor`, etc.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            });
+
+            var data = new GenericRow[]
+            {
+                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) })
+            };
+
+            DataFrame df = _spark.CreateDataFrame(data, schema);
+
+            df.WithColumn("tsRow", Struct("ts")).WithColumn("tsRowRow", Struct("tsRow")).Collect().ToArray();
```
imback82, line 39: Can you check the actual values?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            var data = new GenericRow[]
+            {
+                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) }),
+                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) }),
+                new GenericRow(new object[] { new Timestamp(2020, 1, 1, 0, 0, 0, 0) })
+            };
+
+            DataFrame df = _spark.CreateDataFrame(data, schema);
+
+            df.Collect().ToArray();
```
imback82, line 59: ditto

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            var data = new GenericRow[]
+            {
+                new GenericRow(new object[] { new Date(2020, 1, 1) }),
+                new GenericRow(new object[] { new Date(2020, 1, 1) }),
+                new GenericRow(new object[] { new Date(2020, 1, 1) })
+            };
+
+            DataFrame df = _spark.CreateDataFrame(data, schema);
+
+            df.Collect().ToArray();
```
imback82, line 79: ditto

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
```diff
@@ -0,0 +1,82 @@
...
+using Microsoft.Spark.Sql;
+using Microsoft.Spark.Sql.Types;
+using Xunit;
+using static Microsoft.Spark.Sql.Functions;
+
+namespace Microsoft.Spark.E2ETest.IpcTests
+
+{
+    [Collection("Spark E2E Tests")]
+    public class RowTests
```
imback82, line 15: Can we also have a test with UDF? https://github.com/dotnet/spark/issues/760

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
suhsteve: Worker will crash without this, so I believe it is ?

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
suhsteve: sure.  readded.

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
suhsteve: sure.  readded.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
suhsteve: added same udf

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
suhsteve: added comment.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
suhsteve: validating values.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
suhsteve: validating values.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/RowTests.cs
suhsteve: validating values.

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
suhsteve: updated class summary.

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
suhsteve: updated class summary.

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
imback82: oh I meant respect to the code. I think "extra layer is removed" is regarding the `RowConstructor`, but now that it's gone, is the comment up to date?

suhsteve: Extra layer can refer to `Row`, so we take out `Values` from it ?

suhsteve: I'm okay with removing the ( )'s though if things sound unclear.

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
```diff
@@ -94,7 +94,7 @@ public Timestamp(DateTime dateTime)
         /// <summary>
         /// Readable string representation for this type.
         /// </summary>
-        public override string ToString() => _dateTime.ToString("yyyy-MM-dd HH:mm:ss.ffffff");
+        public override string ToString() => _dateTime.ToString("yyyy-MM-dd HH:mm:ss.ffffffZ");
```
suhsteve, line 97: I assume this is a bug ?  Converting to string and casting back to Timestamp in Spark caused the time to shift 8 hours.
@elvaliuliuliu @imback82

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
suhsteve: @elvaliuliuliu do we need to update the description or does it still apply ?

src/csharp/Microsoft.Spark/Sql/Types/Timestamp.cs
suhsteve: Adding a separate PR instead to address this.  https://github.com/dotnet/spark/pull/871

imback82: thanks. let's discuss this in #871 and remove from this PR.

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
```diff
@@ -11,105 +11,48 @@
 namespace Microsoft.Spark.Sql
 {
     /// <summary>
-    /// RowConstructor is a custom unpickler for GenericRowWithSchema in Spark.
+    /// RowConstructorConstructor is a custom unpickler for GenericRowWithSchema in Spark.
     /// Refer to spark/sql/core/src/main/scala/org/apache/spark/sql/execution/python/
     /// EvaluatePython.scala how GenericRowWithSchema is being pickled.
     /// </summary>
-    internal sealed class RowConstructor : IObjectConstructor
+    internal sealed class RowConstructorConstructor : IObjectConstructor
```
imback82, line 18: Is there a better name than `RowConstructorConstructor` or is this a convention?

suhsteve: I'm okay with any name suggestion.  `RowConstructorFactory`?  Rename back to `RowConstructor` and change the class at the bottom to `RowWithSchemaConstructor` ?

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
```diff
@@ -11,105 +11,48 @@
...
-            {
-                if (_args[i] is RowConstructor rowConstructor)
-                {
-                    _args[i] = rowConstructor.GetRow();
-                }
-            }
-
-            return new Row(_args, _parent.GetSchema());
+            Debug.Assert((args != null) && (args.Length == 1) && (args[0] is string));
+            return new RowConstructor(GetSchema((string)args[0]));
```
imback82, line 47: Can we just pass `s_schemaCache` to `GetSchema`?

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
imback82: the latter sounds fine to me?

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
suhsteve: I don't follow.  You want `s_schemaCache` to be passed as an argument to `GetSchema` ?

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
```diff
@@ -11,105 +11,48 @@
...
-        /// Construct a Row object from unpickled data. This is only to be called
-        /// on a RowConstructor that contains the row data.
-        /// </summary>
-        /// <returns>A row object with unpickled data</returns>
-        public Row GetRow()
-        {
-            Debug.Assert(_parent != null);
-
-            // It is possible that an entry of a Row (row1) may itself be a Row (row2).
-            // If the entry is a RowConstructor then it will be a RowConstructor
```
imback82, line 92: I guess we already have test case handling this right?

suhsteve: yeah there are a few that have rows as column values.

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
imback82: yes, so that `GetSchema` doesn't have to depend on `s_schemaCache` directly.

imback82: and we can remove the assert from `GetSchema` since there is no prerequisite now.

suhsteve: got it.  modified.

src/csharp/Microsoft.Spark/Sql/RowConstructor.cs
suhsteve: renamed to latter.

imback82 (approved):
LGTM (if tests pass), thanks @suhsteve!

imback82:
@suhsteve BTW, did you also test against the repro in #760?

suhsteve:
> @suhsteve BTW, did you also test against the repro in #760?

Yeah I ran it against the repro.

suhsteve:
> @suhsteve BTW, did you also test against the repro in #760?

Hmm I'm surprised it's passing.  It was failing earlier for `TestUdfWithDuplicateTimestamps`

suhsteve:
> > @suhsteve BTW, did you also test against the repro in #760?
> 
> Hmm I'm surprised it's passing. It was failing earlier for `TestUdfWithDuplicateTimestamps`

The target machines must be using UTC time as their timezone.  It fails locally on my machine.

```
  Message: 
    Assert.Equal() Failure
    Expected: 1970-01-02 00:00:00.000000
    Actual:   1970-01-02 08:00:00.000000
```

imback82:
Can you push an empty commit?

imback82 (approved):

PR #871, opened 2021-03-26
Add 'Z' to the string format in Timestamp.ToString() to indicate UTC time.

suhsteve:
Currently the following evaluates to `false`
```csharp
var dateTimeObj = new DateTime(2021, 01, 01);
new Timestamp(DateTime.Parse(new Timestamp(dateTimeObj).ToString())).Equals(new Timestamp(dateTimeObj))
```

This PR addresses this by updating the `Timestamp.ToString` method to include the UTC timezone information; by adding `Z` to the string format.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -16,20 +16,27 @@ variables:
   backwardCompatibleRelease: '1.0.0'
   forwardCompatibleRelease: '1.0.0'
 
-  backwardCompatibleTestOptions_Windows_2_3: ""
-  forwardCompatibleTestOptions_Windows_2_3: ""
-  backwardCompatibleTestOptions_Linux_2_3: ""
-  forwardCompatibleTestOptions_Linux_2_3: ""
+  # Filter UdfSimpleTypesTests.TestUdfWithReturnAsTimestampType and UdfSimpleTypesTests.TestUdfWithTimestampType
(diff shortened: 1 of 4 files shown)


imback82:
Can you paste the expected output from either Scala or Python in the PR description?

imback82:
Thanks for the update. Could you also paste the output from .NET to see the difference?

azure-pipelines.yml
```diff
@@ -16,14 +16,17 @@ variables:
   backwardCompatibleRelease: '1.0.0'
   forwardCompatibleRelease: '1.0.0'
 
-  backwardCompatibleTestOptions_Windows_2_3: ""
+  # Filter Timestamp tests due to Timestamp.ToString() changes. Please see https://github.com/dotnet/spark/pull/871
```
imback82, line 19: Can you also mention that this is a bug fix, not necessarily a breaking change?

imback82 (commented):
Can you check the test failure please?

imback82 (approved):
LGTM, thanks @suhsteve!

Niharikadutta:
@suhsteve Could you check the conflicting `azure-pipelines.yml` ?

imback82 (approved):

imback82:
Since this breaks forward compatibility, I am reverting this change, and come back to this for 2.0 release.

PR #886, opened 2021-04-07
Spark 3.1.0 APIs - DataStreamReader and DataStreamWriter 

suhsteve:

**DataStreamReader**
```scala
def table(tableName: String): DataFrame
```

**DataStreamWriter**
```scala
def toTable(tableName: String): StreamingQuery
```


imback82 (approved):
LGTM, thanks @suhsteve!

PR #887, opened 2021-04-07
Spark 3.1.0 APIs - Column

suhsteve:

**Column**
```scala
def withField(fieldName: String, col: Column): Column
def dropFields(fieldNames: String*): Column
```


imback82 (approved):
LGTM, thanks @suhsteve!

src/csharp/Microsoft.Spark/Sql/Column.cs
```diff
@@ -450,6 +450,31 @@ public Column GetItem(object key)
             return ApplyMethod("getItem", key);
         }
 
+        /// <summary>
+        /// An expression that adds/replaces field in <see cref="Types.StructType"/> by name.
+        /// </summary>
+        /// <param name="fieldName">The name of the field</param>
+        /// <param name="column"></param>
```
suhsteve, line 457: missing description

PR #888, opened 2021-04-07
Spark 3.1.0 APIs - DataFrame

suhsteve:

**DataFrame**
```scala
def unionByName(other: Dataset[T], allowMissingColumns: Boolean): Dataset[T]
def sameSemantics(other: Dataset[T]): Boolean
def semanticHash(): Int
```

Added into scala in Spark 2.0.0, but added into pySpark in Spark 3.1.0
```scala
def inputFiles: Array[String]
```

### DIFF
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
@@ -696,7 +696,8 @@ namespace Microsoft.Spark.E2ETest.IpcTests
         }
 
         /// <summary>
-        /// Test signatures for APIs introduced in Spark 3.0.*
+        /// Test signatures for APIs introduced in Spark 3.0.*.
+
         /// </summary>
         [SkipIfSparkVersionIsLessThan(Versions.V3_0_0)]
         public void TestSignaturesV3_0_X()
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
@@ -668,7 +668,7 @@ namespace Microsoft.Spark.E2ETest.IpcTests
             _df.CreateGlobalTempView("global_view");
             _df.CreateOrReplaceGlobalTempView("global_view");
 
-            Assert.IsType<string[]>(_df.InputFiles());
+            Assert.IsType<string[]>(_df.InputFiles().ToArray());
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -696,10 +696,10 @@ public void TestSignaturesV2_4_X()
         }
 
         /// <summary>
-        /// Test signatures for APIs introduced in Spark 3.*
+        /// Test signatures for APIs introduced in Spark 3.0.*
```
Niharikadutta, line 699:
```suggestion
        /// Test signatures for APIs introduced in Spark 3.0.*.

```

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -729,5 +729,18 @@ public void TestSignaturesV3_X_X()
             _df.Explain("cost");
             _df.Explain("formatted");
         }
+
+        /// <summary>
+        /// Test signatures for APIs introduced in Spark 3.1.*
```
Niharikadutta, line 734:
```suggestion
        /// Test signatures for APIs introduced in Spark 3.1.*.

```

Niharikadutta:
LGTM @suhsteve (few minor nits). Thanks!

Niharikadutta:
@suhsteve One quick question, should we include the `inputFiles` API as introduced in [Python from 3.1.0](https://github.com/apache/spark/blob/0494dc90af48ce7da0625485a4dc6917a244d580/python/pyspark/sql/dataframe.py#L2626)?

suhsteve:
> @suhsteve One quick question, should we include the `inputFiles` API as introduced in [Python from 3.1.0](https://github.com/apache/spark/blob/0494dc90af48ce7da0625485a4dc6917a244d580/python/pyspark/sql/dataframe.py#L2626)?

Hmm, that API was introduced on the scala side since 2.0.0, but just got into pyspark in 3.1.0.  Wouldn't hurt to expose it as well.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataFrameTests.cs
```diff
@@ -729,5 +732,19 @@ public void TestSignaturesV3_X_X()
             _df.Explain("cost");
             _df.Explain("formatted");
         }
+
+        /// <summary>
+        /// Test signatures for APIs introduced in Spark 3.1.*.
+
```
imback82, line 738:
```suggestion

```

imback82 (dismissed):
LGTM, thanks @suhsteve!

PR #889, opened 2021-04-07
Spark 3.1.0 APIs - SparkContext

suhsteve:

**SparkContext**
Added in `scala` in Spark 3.1.0
```scala
def addArchive(path: String): Unit
def listArchives(): Seq[String]
```

Added in `pySpark` in Spark 3.1.0
```scala
def getCheckpointDir: Option[String]
```

Added in `scala` in Spark 2.0.0
```scala
def listFiles(): Seq[String]
```


src/csharp/Microsoft.Spark/SparkContext.cs
```diff
@@ -289,6 +291,26 @@ public void AddFile(string path, bool recursive = false)
...
+        /// Archive path can be either a local file, a file in HDFS (or other Hadoop-supported
+        /// filesystems), or an HTTP, HTTPS or FTP URI. The given path should be one of .zip,
+        /// .tar, .tar.gz, .tgz and .jar.
+        /// </param>
+        [Since(Versions.V3_1_0)]
+        public void AddArchive(string path)
+        {
+            _jvmObject.Invoke("addArchive", path);
+        }
+
```
Niharikadutta, line 313: Is there a reason we haven't included `[listArchives]`(https://github.com/apache/spark/blob/0494dc90af48ce7da0625485a4dc6917a244d580/core/src/main/scala/org/apache/spark/SparkContext.scala#L1564)?

suhsteve: I assumed `listFiles` was originally left out on purpose, so following the same pattern I left this out as well.  Although I don't really see any harm in adding both.

suhsteve: Hmm, `listFiles` was probably never added because `pySpark` doesn't expose it.

src/csharp/Microsoft.Spark/SparkContext.cs
```diff
@@ -300,6 +338,21 @@ public void SetCheckpointDir(string directory)
...
+        /// <returns>
+        /// The directory where RDDs are checkpointed. Returns `null` if no checkpoint
+        /// directory has been set.
+        /// </returns>
+        public string GetCheckpointDir()
+        {
+            var optionalCheckPointDir =
+                new Option((JvmObjectReference)_jvmObject.Invoke("getCheckpointDir"));
+
+            return optionalCheckPointDir.IsDefined() ? (string)optionalCheckPointDir.Get() : null;
```
imback82, line 353: `OrNull()`?

Btw, is this how we have been handling if the API returns `Option`?

suhsteve: yup, we've been using the above pattern in other places.

imback82 (approved):
LGTM, thanks @suhsteve!

PR #890, opened 2021-04-07
Spark 3.1.0 APIs - Functions

suhsteve:

**Functions**
```scala
def percentile_approx(e: Column, percentage: Column, accuracy: Column): Column
def nth_value(e: Column, offset: Int, ignoreNulls: Boolean): Column
def nth_value(e: Column, offset: Int): Column
def acosh(e: Column): Column
def acosh(columnName: String): Column
def asinh(e: Column): Column
def asinh(columnName: String): Column
def atanh(e: Column): Column
def atanh(columnName: String): Column
def assert_true(c: Column): Column
def assert_true(c: Column, e: Column): Column
def raise_error(c: Column): Column
def timestamp_seconds(e: Column): Column
def slice(x: Column, start: Column, length: Column): Column
```


Niharikadutta (commented):
LGTM @suhsteve . Thanks!

imback82 (approved):
LGTM, thanks @suhsteve!

PR #893, opened 2021-04-07
Spark 3.1.0 APIs - Catalog

suhsteve:
PR adds support for the following APIs added in Spark 3.1.0 and adds missing `createTable` APIs:

**Catalog**
Introduced in Spark 3.1.0
```scala
def createTable(tableName: String, source: String, description: String, options: Map[String, String]): DataFrame
def createTable(tableName: String, source: String, schema: StructType, description: String, options: Map[String, String]): DataFrame
```

Missing createTable APIs
```scala
def createTable(tableName: String, source: String, options: Map[String, String]): DataFrame
def createTable(tableName: String, source: String, schema: StructType, options: Map[String, String]): DataFrame
```

### DIFF
--- a/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/CatalogTests.cs
+++ b/src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/CatalogTests.cs
@@ -1,0 +1,114 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Spark.E2ETest.Utils;
+using Microsoft.Spark.Sql;
+using Microsoft.Spark.Sql.Catalog;
+using Microsoft.Spark.Sql.Types;
+using Xunit;
+using static Microsoft.Spark.E2ETest.Utils.SQLUtils;
+
+namespace Microsoft.Spark.E2ETest.IpcTests
+{
+    [Collection("Spark E2E Tests")]
+    public class CatalogTests
+    {
+        private readonly SparkSession _spark;
+
+        public CatalogTests(SparkFixture fixture)
+        {
+            _spark = fixture.Spark;
+        }
+
+        /// <summary>
+        /// Test signatures for APIs up to Spark 2.3.*.
+        /// </summary>
+        [Fact]
+        public void TestSignaturesV2_3_X()
+        {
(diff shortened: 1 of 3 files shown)


imback82 (approved):
LGTM, thanks @suhsteve!

imback82:
Is the following a random failure we need to revisit?
```
 Failed Microsoft.Spark.E2ETest.IpcTests.StreamingQueryManagerTests.TestSignaturesV2_3_X [470 ms]
  Error Message:
   Assert.Equal() Failure
Expected: 2
Actual:   3
  Stack Trace:
     at Microsoft.Spark.E2ETest.IpcTests.StreamingQueryManagerTests.TestSignaturesV2_3_X() in D:\a\1\s\dotnet-spark\src\csharp\Microsoft.Spark.E2ETest\IpcTests\Sql\Streaming\StreamingQueryManagerTests.cs:line 46

```

suhsteve:
> Is the following a random failure we need to revisit?
> 
> ```
>  Failed Microsoft.Spark.E2ETest.IpcTests.StreamingQueryManagerTests.TestSignaturesV2_3_X [470 ms]
>   Error Message:
>    Assert.Equal() Failure
> Expected: 2
> Actual:   3
>   Stack Trace:
>      at Microsoft.Spark.E2ETest.IpcTests.StreamingQueryManagerTests.TestSignaturesV2_3_X() in D:\a\1\s\dotnet-spark\src\csharp\Microsoft.Spark.E2ETest\IpcTests\Sql\Streaming\StreamingQueryManagerTests.cs:line 46
> ```

May be something to look into.  Could be from other streaming queries from other tests that haven't been stopped.

suhsteve:
> > Is the following a random failure we need to revisit?
> > ```
> >  Failed Microsoft.Spark.E2ETest.IpcTests.StreamingQueryManagerTests.TestSignaturesV2_3_X [470 ms]
> >   Error Message:
> >    Assert.Equal() Failure
> > Expected: 2
> > Actual:   3
> >   Stack Trace:
> >      at Microsoft.Spark.E2ETest.IpcTests.StreamingQueryManagerTests.TestSignaturesV2_3_X() in D:\a\1\s\dotnet-spark\src\csharp\Microsoft.Spark.E2ETest\IpcTests\Sql\Streaming\StreamingQueryManagerTests.cs:line 46
> > ```
> 
> May be something to look into. Could be from other streaming queries from other tests that haven't been stopped.

I think we can just modify it and not validate the length since we are only interested in checking that JVM calls are successful.

imback82:
> I think we can just modify it and not validate the length since we are only interested in checking that JVM calls are successful.

We should first investigate the root cause. I will create an issue to follow up.

PR #952, opened 2021-06-10
Extensions.DotNet.Interactive - Add ENV to control disposal of tmp dir

suhsteve:
The DotnetInteractive extension creates a temp directory to store dlls and nuget packages and registers this temp directory with the DotnetInteractive kernel for disposal.

This PR adds a new ENV variable to control whether this tmp directory should be registered for disposal.

### DIFF
--- a/src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive/AssemblyKernelExtension.cs
+++ b/src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive/AssemblyKernelExtension.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Spark.Extensions.DotNet.Interactive
     public class AssemblyKernelExtension : IKernelExtension
     {
         private const string TempDirEnvVar = "DOTNET_SPARK_EXTENSION_INTERACTIVE_TMPDIR";
-        private const string DisposeTempDirEnvVar = "DOTNET_SPARK_EXTENSION_INTERACTIVE_DISPOSE_TMPDIR";
+        private const string PreserveTempDirEnvVar = "DOTNET_SPARK_EXTENSION_INTERACTIVE_PRESERVE_TMPDIR";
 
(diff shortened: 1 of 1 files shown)


suhsteve:
@gaoyangxiaozhu can you also add context here ?

gaoyangxiaozhu:
thank you @suhsteve for the PR. The PR here can help fix one our customer issue where launch and kill the dotnetinterpreter will cause the spark crash due to DLLs file not found issue.

In details, in our prod environment, we will launch multiple interpreters (sparkdotnet, python) for customer to run the code from referenced notebook file . And we will kill the related interpreter once the run finished. For example, we will launch two interpreters: one is spark dotnet interpreter named A1 and another is python interpreter named A2 for run code blocks C1 (C#) and C2 (python) respectively. And  A1 will be direcrtly killed if the code blocks C1 has been run finished.

Will kill the spark dotnet interpereter will cause the temp folder will also be deleted due to this code from dotnet interpreter extension part 
https://github.com/dotnet/spark/blob/4ede264c2bfd40a4249a6f49065055df8c936502/src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive/AssemblyKernelExtension.cs#L44 

Since the extension already add this DLLs files to sparkContext via addfile API, so when trigger another spark action as spark.range(10).collect() after dotnet interpreter been killed, the executors will still try to fetch the missed files which added by user from driver node and then encounter file not found issue.

image

It can been seen that current implement doesn't cover the scenarios that dotnet interpreter can be created/killed multiple times before spark application run finished.
With @suhsteve 's fix to introduce this addition env variable to dotnet interactive extension to control if need to delete the temp file when interpreter been killed. We can keep the temp folder to avoid DLLs not found issue.

Niharikadutta:
LGTM, thanks @suhsteve

imback82 (approved):
LGTM, thanks @suhsteve (assuming this is tested E2E)!

Issue #818, opened 2021-02-03
[FEATURE REQUEST]: Support System.Decimal in CreateDataFrame API


jgavin:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to use the `System.Decimal` type when using the CreateDataFrame API. Currently we see the following:

`decimal? minValue = 4.0M;
  decimal? maxValue = 5.0M;
  var df = sparkSession.CreateDataFrame(
      new List<GenericRow>
      {
          new GenericRow(
              new object[]
              {
                  "1234",
                  minValue,
                  maxValue,
              }),
      },
      new StructType(
          new List<StructField>()
          {
              new StructField("companyId", new StringType()),
              new StructField("minValue", new DecimalType()),
              new StructField("maxValue", new DecimalType()),
          }));
  df.Show();`

`System.NotSupportedException: 'Type System.Decimal not supported yet'`

This presents issues in trying to unit test code that reads decimal types from a DB.  

**Describe the solution you'd like**
To be able to create Dataframes with Decimals without an error 😄 

**Describe alternatives you've considered**
Alternatives for us might be to change the type being read from the DB, but this is a significant challenge - for now we are unable to unit test this code. 

**Additional context**
N/A

Niharikadutta:
Thanks @jgavin for creating this issue, we are yet to add support for `System.Decimal` and will get back to you about the timeline of getting this in.

dbeavon:
@Niharikadutta 

Did you see that decimals were added to Apache Arrow?

- https://issues.apache.org/jira/browse/ARROW-11422
- https://github.com/apache/arrow/pull/9356

This happened recently.  Was that a prerequisite?  Decimal types are pretty critical for us as well.

GoEddie:
Hi @Niharikadutta + @imback82  - is there a reason for no decimal or is just waiting for development? is this a good issue to take?

imback82:
Hi @GoEddie, this is just waiting for development and a good issue to take.

PR #982, opened 2021-10-14
Feature/decimal support

GoEddie:

This implements https://github.com/dotnet/spark/issues/818

On the Apache Spark side decimal is implemented using a java.math.BigDecimal and the only way to construct that is using a int/long - there is no way to construct it with any value larger than max long unless you use the string constructor so I pass a string back and forth between .NET and the JVM, hope that is ok.

### DIFF
--- a/src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
+++ b/src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
@@ -75,12 +75,6 @@ namespace Microsoft.Spark.Interop.Ipc
 
         private ISocketWrapper GetConnection()
         {
-            // Limit the number of connections to the JVM backend. Netty is configured
-            // to use a set number of threads to process incoming connections. Each
-            // new connection is delegated to these threads in a round robin fashion.
-            // A deadlock can occur on the JVM if a new connection is scheduled on a
-            // blocked thread.
-            _socketSemaphore.Wait();
             if (!_sockets.TryDequeue(out ISocketWrapper socket))
             {
                 socket = SocketFactory.CreateSocket();
@@ -188,8 +182,16 @@ namespace Microsoft.Spark.Interop.Ipc
         {
             object returnValue = null;
             ISocketWrapper socket = null;
+
             try
-            {
+            {                
+                // Limit the number of connections to the JVM backend. Netty is configured
+                // to use a set number of threads to process incoming connections. Each
+                // new connection is delegated to these threads in a round robin fashion.
+                // A deadlock can occur on the JVM if a new connection is scheduled on a
+                // blocked thread.
+                _socketSemaphore.Wait();
+
                 // dotnet-interactive does not have a dedicated thread to process
                 // code submissions and each code submission can be processed in different
                 // threads. DotnetHandler uses the CLR thread id to ensure that the same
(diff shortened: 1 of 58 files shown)


Macromullet:
Can we get some traction on this one? It looks like this PR is ready and we have a lot of code that uses decimals that will be difficult to write tests around without this.

src/csharp/Microsoft.Spark/Interop/Ipc/SerDe.cs
```diff
@@ -322,6 +322,13 @@ public static void Write(Stream s, long value)
         public static void Write(Stream s, double value) =>
             Write(s, BitConverter.DoubleToInt64Bits(value));
 
+        /// <summary>
+        /// Writes a decimal to a stream as a string.
+        /// </summary>
+        /// <param name="s">The stream to write</param>
+        /// <param name="value">The decimal to write</param>
+        public static void Write(Stream s, decimal value) => Write(s, value.ToString());
```
Macromullet, line 330: Should use ToString(CultureInfo.InvariantCulture) if we are using a string on the wire.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -267,6 +267,9 @@ private ISocketWrapper GetConnection()
                     case 'd':
                         returnValue = SerDe.ReadDouble(inputStream);
                         break;
+                    case 'm':
+                        returnValue = decimal.Parse(SerDe.ReadString(inputStream));
```
Macromullet, line 271: Use decimal.Parse(SerDe.ReadString(inputStream), CultureInfo.InvariantCulture) to ensure we are using invariant culture on the wire.

Macromullet:
Any updates on this PR?

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/DataTypesTests.cs
```diff
@@ -0,0 +1,65 @@
...
+                        new StructField("max", new DecimalType(38, 0)),
+                        new StructField("zero", new DecimalType(38, 0)),
+                        new StructField("minusOne", new DecimalType(38, 0)),
+                        new StructField("array", new ArrayType(new DecimalType(38,0)))
+                    }));
+
+            Row row = df.Collect().First();
+            Assert.Equal(decimal.MinValue, row[0]);
+            Assert.Equal(decimal.MaxValue, row[1]);
+            Assert.Equal(decimal.Zero, row[2]);
```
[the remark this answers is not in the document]
cutecycle: > This is because internally, BigDecimal uses BigInteger, and BigInteger also only has a single concept of zero. A BigInteger behaves as a two's-complement integer, and two's-complement only has a single zero.

AFFogarty:
Hey @GoEddie , are you still working on this?

GoEddie:
Hi @AFFogarty,

I had given up really as no one seemed to be reviewing pr’s but am happy to get it up to date again.

ed

wudanzy:
Hi @GoEddie , thanks for the contribution! Are you still working on this? We recently get write permission of this repo and happy to move this forward. We just added support for 3.3 and 3.5.

PR #994, opened 2021-10-29
Support for Spark 2.4.8

suhsteve:
Add support for Spark 2.4.8 and add E2E tests for these new supported Spark versions


suhsteve:
Should we add 3.0.2 and 3.0.3 ?

PR #1021, opened 2022-02-02
Switch to 1ES hosted pools on main

jonfortescue:
We've officially decommissioned the old BuildPools (see partners mail) in favor of 1ES hosted pools. This change moves dotnet/spark appropriately.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -121,8 +121,8 @@ stages:
         - Build
       displayName: Sign Artifacts
       pool:
-        name: NetCoreInternal-Pool
-        queue: buildpool.windows.10.amd64.vs2017
+        name: NetCore1ESPool-Internal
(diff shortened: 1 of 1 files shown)


suhsteve:
@jonfortescue We are also using the `Hosted VS2017` pool in our pipeline.  Are there any plans to decommission this in the future or is it still fine to use ?  If the former, then what should we switch to instead ?

jonfortescue:
@suhsteve nope, no plans to decommission those pools. You're good to use them as long as they're only being used for public builds; for internal builds, the guidance is to only use the 1ES hosted pools.

suhsteve (approved):
LGTM, thanks @jonfortescue !

jonfortescue:
@suhsteve awesome! I can't merge bc I'm not a maintainer so if you could merge for me that would be greatly appreciated. :)

imback82 (approved):
LGTM, thanks @jonfortescue!

Issue #1060, opened 2022-04-29
[BUG]: Semaphore Release Bug when an error occurs before GetConnection in JvmBridge


Macromullet:
**Describe the bug**
In JvmBridge, a semaphore is used to restrict the number of active sockets. The semaphore is waited for in GetConnection() and released in the finally block. Unfortunately there are calls before GetConnection() where the semaphore is acquired that are not guaranteed to always run successfully. In situations where there is an error before GetConnection(), the semaphore is never attempted to Wait() and the subsequent finally Release() yields an error that the semaphore count went over the max allowed.

Steps to repro (easy):
The easiest way to make this happen is to add a throw new Exception("Test") block before the GetConnection() call.

Steps to reproduce the behavior (how I found it):
I tried to make a dataFrame using the GenericRow[] and passed DateTime.Now as one of the objects. PayloadHelper.BuildPayload didn't like that and errored. This caused the semaphore counts to get out of whack.

**Expected behavior**
Semaphores should always have perfect acquire/release guarantees.

I'll attempt a PR

PR #1061, opened 2022-04-29
JvmBridge semaphore fix

Macromullet:
This PR address issue #1060 by doing the following:

- moved field usage of _socketSemaphore to a single method, vs cross-method ensuring that one method has strict acquire/release semantics and that no other calls to GetConnection(), even inadvertently could ever mess with the semaphore
- Moved semaphore Wait() call to the first method of the try/finally so that it gets called no matter what
- I added an ifnull check around _sockets.Enqueue(socket), just to be extra safe as well in the finally block.

I have tested this using a fairly large project that was exhibiting the error (See the bug) but I'm unsure of how to write an automated test around the original failure scenario without adding some API abstractions around the semaphore lock itself and making something like PerformSemaphoreLockedOperation(SemaphoreSlim semaphore, Action action); I didn't wan't to go overboard.

### DIFF
--- a/src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
+++ b/src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
@@ -75,11 +75,6 @@ namespace Microsoft.Spark.Interop.Ipc
 
         private ISocketWrapper GetConnection()
         {
-            // Limit the number of connections to the JVM backend. Netty is configured
-            // to use a set number of threads to process incoming connections. Each
-            // new connection is delegated to these threads in a round robin fashion.
-            // A deadlock can occur on the JVM if a new connection is scheduled on a
-            // blocked thread.
             if (!_sockets.TryDequeue(out ISocketWrapper socket))
             {
                 socket = SocketFactory.CreateSocket();
@@ -187,10 +182,16 @@ namespace Microsoft.Spark.Interop.Ipc
         {
             object returnValue = null;
             ISocketWrapper socket = null;
-            var acquiredSemaphore = false;
 
             try
-            {
+            {                
+                // Limit the number of connections to the JVM backend. Netty is configured
+                // to use a set number of threads to process incoming connections. Each
+                // new connection is delegated to these threads in a round robin fashion.
+                // A deadlock can occur on the JVM if a new connection is scheduled on a
+                // blocked thread.
+                _socketSemaphore.Wait();
+
                 // dotnet-interactive does not have a dedicated thread to process
                 // code submissions and each code submission can be processed in different
                 // threads. DotnetHandler uses the CLR thread id to ensure that the same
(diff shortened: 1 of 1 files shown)


src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -218,6 +219,8 @@ private ISocketWrapper GetConnection()
                     methodName,
                     args);
 
+                _socketSemaphore.Wait();
```
suhsteve, line 222: Can this just be moved to the top of this `try` block ?  Then there would be no need to introduce the `bool` flag.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -80,7 +80,6 @@ private ISocketWrapper GetConnection()
             // new connection is delegated to these threads in a round robin fashion.
             // A deadlock can occur on the JVM if a new connection is scheduled on a
             // blocked thread.
```
suhsteve, line 82: This comment block should follow semaphore.Wait

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
```diff
@@ -290,7 +293,10 @@ private ISocketWrapper GetConnection()
                 {
                     // DotnetBackendHandler caught JVM exception and passed back to dotnet.
                     // We can reuse this connection.
-                    _sockets.Enqueue(socket);
+                    if (socket != null) // Safety check
```
suhsteve, line 296: Combine this with the outer `if` ?

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
Macromullet: It could. I was under the impression that there was a reason you wanted to grab the connection later. I agree that moving it up top would simplify the code and was my initial idea, but I assumed the connection was established later on purpose. I don't see any other fields mutated before GetConnection being used by GetConnection so it seems like a safe operation. If you agree let me know and I'll go ahead and change.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
Macromullet: Makes sense. Thanks.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
suhsteve: I think it should be safe to go ahead and move the wait to the top.

Macromullet: Pushed a fix. Please review.

src/csharp/Microsoft.Spark/Interop/Ipc/JvmBridge.cs
Macromullet: Pushed a fix. Please review.

suhsteve (approved):
LGTM thanks @Macromullet

PR #1132, opened 2023-01-11
Wait before running apt command

ericstj:
Trying out a fix for the CI issue that's surfacing as:

```
E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable) 
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it? 
```

When running `apt` commands.  As seen here: https://dev.azure.com/dnceng-public/public/_build/results?buildId=126859&view=logs&j=6275ae02-8f8b-5df6-73dc-1bd6a4ec92dd&t=bfb24330-1904-5583-de8c-be2e1157559c&l=29

### DIFF
--- a/azure-pipelines-e2e-tests-template.yml
+++ b/azure-pipelines-e2e-tests-template.yml
@@ -44,13 +44,13 @@ stages:
 
               while($numRetry -le $maxRetry)
               {
-                  sudo fuser /var/lib/dpkg/lock-frontend >/dev/null 2>&1
+                  sudo fuser /var/lib/dpkg/lock-frontend 2>&1
 
                   if ($LASTEXITCODE -ne 0) { Break }
                   
                   sleep 1
                   $numRetry++
-                  echo 'Waited $numRetry s for release of /var/lib/dpkg/lock-frontend'
+                  echo "Waited $numRetry s for release of /var/lib/dpkg/lock-frontend"
               }
               
               sudo apt update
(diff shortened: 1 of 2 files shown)


ericstj:
My attempts to move this into a script failed, because the leg that's running this doesn't clone before running.  I need to port the logic into powershell entirely.

ericstj:
Looks like my theory here didn't work out.  This now waits for some package operation to finish, but that doesn't happen after 5 minutes, so perhaps the theory that some updater service is running in the background is not correct.  🤷‍♂️

AFFogarty:
It seems that with a 5 minute timeout, it sometimes works and sometimes doesn't.  I made a commit increasing the max to 10 minutes, to see what happens.

AFFogarty (approved):
The build succeeds with this mitigation.  Let's ship this, then RCA the issue later.

ericstj:
RCA and fix is suggested here: https://github.com/dotnet/arcade/issues/12240 but this change is fine to take regardless as @AFFogarty suggested.

AFFogarty:
> RCA and fix is suggested here: [dotnet/arcade#12240](https://github.com/dotnet/arcade/issues/12240) but this change is fine to take regardless as @AFFogarty suggested.

Let's ship this PR for the sake of unblocking folks ASAP.  Then we can pull that Arcade fix once it's available.

suhsteve (approved):

PR #1157, opened 2023-08-04
Removing deprecated tasks

missymessa:
Hi dotnet/spark, 

The .NET Eng Services team is fully deprecating a couple of custom Azure DevOps tasks that haven't been used in many years. This PR is to clean up usage of those tasks. 

Thanks!
Missy

See: https://github.com/dotnet/dnceng/issues/507

### DIFF
--- a/eng/common/templates/job/job.yml
+++ b/eng/common/templates/job/job.yml
@@ -125,19 +125,6 @@ jobs:
     workspace: ${{ parameters.workspace }}
 
   steps:
-  - ${{ if eq(parameters.enableTelemetry, 'true') }}:
-    # Telemetry tasks are built from https://github.com/dotnet/arcade-extensions
-    - task: sendStartTelemetry@0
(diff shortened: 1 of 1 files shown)


wudanzy:
Hi @missymessa , is this PR still needed?

grazy27:
> Hi @missymessa , is this PR still needed?

Everything inside `eng\common` is defined in Arcade, and should be copied from there. 
For other dotnet repos, there's a bot that automatically creates a PR and overwrites changes for Arcade each release. 

Btw, it seems we don't use these yamls at all, since new pipelines are defined separately

missymessa:
Up to you all if you think this is still necessary. At the time this PR was created, this repo had the tasks that we were removing, so we wanted to do our due diligence to make sure we weren't breaking folks. Feel free to close it!

wudanzy:
Close those, since we don't need this any more.

PR #1166, opened 2024-04-05
Replace BinaryFormatter with MessagePack serialization

### LINKED ISSUE #1131: JVM IPC Deserialization uses BinaryFormatter, which is now Deprecated for OWASP CWE
**Describe the bug**
We'll be getting SYSLIB0011 errors for the way the Broadcast, Worker, and RDD are formatting streams. 

My current understanding:

* .NET Spark communicates with a spark workers and makes broadcast variables JVM compatible by using a system of converting CLR objects to JVM objects.
* This is all currently performed raw via BinaryFormatter.
* Untrusted binary deserialization of data with undefined schema can result in RCE.
* The guidance is "go to xml or json/bson", but we can't just do IPC with those formats?
* In theory, protobuf/Arrow could allow us whatever level of control over the individual bytes may be necessary to continue byte-level IPC with the JVM
* The missing piece is a definition of the stream format that would go over the socket?

Historical example: jni4net (also uses BinaryFormatter, but has a some level of struct definition:
*https://github.com/jni4net/jni4net/blob/ac2189c37253710e7b729797631419b0bf3b8559/jni4net.tested.n/src/generated/net/sf/jni4net/tested/JavaInstanceFields.generated.cs#L16

Notes: 
Can we write directly to MemoryStream?
https://github.com/dotnet/spark/pull/1112#discussion_r1094785106

"passing protobufs between Java and C using JNI": 
https://medium.com/@dhaval.durve/passing-protobufs-between-java-and-native-c-code-using-jni-9808b60f6d2c

An equivalent of this CVE, and the object filter used to resolve it
https://security.snyk.io/vuln/SNYK-PYTHON-PYSPARK-3021140

https://github.com/apache/spark/pull/18166/files#diff-6a1d1601920af68466d7c30dc02170468abbe408138734c00d50d2ba1b81ba35R179

BinaryFormatter Guidance: 
https://learn.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide

Arrow buffers:
https://arrow.apache.org/docs/python/ipc.html

BinaryFormatter Marshaller in ProtoBuf.net: 
https://github.com/protobuf-net/protobuf-net.Grpc/blob/main/tests/protobuf-net.Grpc.Test.Integration/CustomMarshaller.cs

Protobuf scalar `bytes` for arbitrary byte lengths:
https://developers.google.com/protocol-buffers/docs/proto3#scalar

Wind down plan in dotnet: 
https://github.com/dotnet/designs/pull/141/commits/bd0a0661f9d248ed31a354d27ad026efd6719690

"Is binary serialization inherently unsafe?" 
https://stackoverflow.com/a/66825699

pyspark's implementation of this is based on py4j; they were going to use protobuf but opted for strings
https://github.com/py4j/py4j/blob/b4514ecd40ea121a35f9cf50bbf2ccea95354245/py4j-python/src/py4j/protocol.py#L9
https://github.com/py4j/py4j/blob/1f8a0b6dc216f16092d9c1b2556897eec8653a62/py4j-python/src/py4j/java_gateway.py#L1737

Though I will say... things seem to be BinaryFormatter all the way down? 
https://github.com/protobuf-net/protobuf-net/search?q=binaryformatter

arsdragonfly:
Fixes #1131 . As explained in my comment #1131 the whole thing is very much pointless due to the nature of Spark being RCE so we must be able to pass arbitrary data/computation around (except for one place where we don't really need the flexibility of deserializing arbitrary types and I've replaced that with a strongly typed deserializer). Nevertheless, using MessagePack's typeless serialization will hopefully saves us the trouble of the ecosystem downstream.

### DIFF
--- a/src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
+++ b/src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
@@ -5,9 +5,9 @@
 using System;
 using System.IO;
 using System.Reflection;
-using System.Runtime.Serialization.Formatters.Binary;
 using Microsoft.Spark.Utils;
 using Xunit;
+using MessagePack;
 
 namespace Microsoft.Spark.UnitTest
 {
@@ -18,6 +18,11 @@ namespace Microsoft.Spark.UnitTest
         private class TestClass
         {
             private readonly string _str;
+
+            // TODO: find out why MessagePack is requiring a parameterless constructor.
+            public TestClass()
+            {
+            }
 
             public TestClass(string s)
             {
@@ -149,16 +154,13 @@ namespace Microsoft.Spark.UnitTest
             return Deserialize(Serialize(udf));
         }
 
-#pragma warning disable SYSLIB0011 // Type or member is obsolete
-        // TODO: Replace BinaryFormatter with a new, secure serializer.
         private byte[] Serialize(Delegate udf)
         {
             UdfSerDe.UdfData udfData = UdfSerDe.Serialize(udf);
 
             using (var ms = new MemoryStream())
             {
-                var bf = new BinaryFormatter();
-                bf.Serialize(ms, udfData);
+                MessagePackSerializer.Typeless.Serialize(ms, udfData);
                 return ms.ToArray();
             }
         }
@@ -167,11 +169,9 @@ namespace Microsoft.Spark.UnitTest
         {
             using (var ms = new MemoryStream(serializedUdf, false))
             {
-                var bf = new BinaryFormatter();
-                UdfSerDe.UdfData udfData = (UdfSerDe.UdfData)bf.Deserialize(ms);
+                UdfSerDe.UdfData udfData = (UdfSerDe.UdfData)MessagePackSerializer.Typeless.Deserialize(ms);
                 return UdfSerDe.Deserialize(udfData);
             }
         }
-#pragma warning restore
     }
 }
--- a/src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
+++ b/src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
@@ -14,7 +14,6 @@ namespace Microsoft.Spark.UnitTest
     [Collection("Spark Unit Tests")]
     public class UdfSerDeTests
     {
-        [Serializable]
         private class TestClass
         {
             private readonly string _str;
--- a/src/csharp/Microsoft.Spark.Worker/Microsoft.Spark.Worker.csproj
+++ b/src/csharp/Microsoft.Spark.Worker/Microsoft.Spark.Worker.csproj
@@ -11,7 +11,8 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.DotNet.DependencyManager" Version="10.10.0-beta.20254.4" />
-    <PackageReference Include="System.Memory" Version="4.5.3" />
+    <PackageReference Include="System.Memory" Version="4.5.5" />
+    <PackageReference Include="MessagePack" Version="2.5.140" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.Spark\Microsoft.Spark.csproj" />
--- a/src/csharp/Microsoft.Spark/Broadcast.cs
+++ b/src/csharp/Microsoft.Spark/Broadcast.cs
@@ -20,7 +20,6 @@ namespace Microsoft.Spark
     /// also attempts to distribute broadcast variables using efficient broadcast algorithms to
     /// reduce communication cost.
     /// </summary>
-    [Serializable]
     public sealed class Broadcast<T> : IJvmObjectReferenceProvider
     {
         [NonSerialized]
--- a/src/csharp/Microsoft.Spark/Broadcast.cs
+++ b/src/csharp/Microsoft.Spark/Broadcast.cs
@@ -4,12 +4,12 @@ using System.Collections.Generic;
 using System.IO;
 using System.Net;
 using System.Runtime.Serialization;
-using System.Runtime.Serialization.Formatters.Binary;
 using System.Threading;
 using Microsoft.Spark.Interop;
 using Microsoft.Spark.Interop.Ipc;
 using Microsoft.Spark.Network;
 using Microsoft.Spark.Services;
+using MessagePack;
 
 namespace Microsoft.Spark
 {
@@ -223,8 +223,7 @@ namespace Microsoft.Spark
         /// <param name="stream">Stream to which the object is serialized</param>
         private void Dump(object value, Stream stream)
         {
-            var formatter = new BinaryFormatter();
-            formatter.Serialize(stream, value);
+            MessagePackSerializer.Typeless.Serialize(stream, value);
         }
     }
 
--- a/src/csharp/Microsoft.Spark/RDD/Collector.cs
+++ b/src/csharp/Microsoft.Spark/RDD/Collector.cs
@@ -6,11 +6,11 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Runtime.Serialization.Formatters.Binary;
 using Microsoft.Spark.Interop.Ipc;
 using Microsoft.Spark.Sql;
 using Microsoft.Spark.Utils;
 using static Microsoft.Spark.Utils.CommandSerDe;
+using MessagePack;
 
 namespace Microsoft.Spark.RDD
 {
@@ -66,15 +66,13 @@ namespace Microsoft.Spark.RDD
         }
 
         /// <summary>
-        /// Deserializer using the BinaryFormatter.
+        /// Deserializer using MessagePack.
         /// </summary>
         private sealed class BinaryDeserializer : IDeserializer
         {
-            private readonly BinaryFormatter _formater = new BinaryFormatter();
-
             public object Deserialize(Stream stream, int length)
             {
-                return _formater.Deserialize(stream);
+                return MessagePackSerializer.Typeless.Deserialize(stream);
             }
         }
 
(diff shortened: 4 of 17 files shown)


dbeavon (commented):
Starting review ... 04/16/2024

dbeavon (approved):
Can we merge this?  

Would be great to have a new version of Microsoft.Spark with these MessagePack changes, and with an upgrade to .net 8.

dbeavon:
@elvaliuliuliu @AFFogarty 

Can someone please help with this merge? Or assign to reviewers who have write access?

This introduces dependencies on :

- "MessagePack"  
- "MessagePack.Annotations" 
- "Microsoft.NET.StringTools" (MSBuild utilities)

@arsdragonfly Have you already found someone to help you merge this, by any chance?

dbeavon:
@luisquintanilla @GoEddie 
Do you have rights to approve of PR's?  Can you help with this one?
It was a primary concern that needed to be addressed, according to  [AFFogarty](https://github.com/AFFogarty)

src/csharp/Microsoft.Spark.Worker/Microsoft.Spark.Worker.csproj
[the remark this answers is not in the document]
arsdragonfly: not sure why the netstandard2.1 tf was there before, looks redundant to me

GeorgeS2019:
@luisquintanilla

You are one of the pioneering advocates. Could you help identify who has the right to trigger the merge?

I will write you en email.

GeorgeS2019:
@dbeavon 

Can this merge also work for .NET 8.0?
https://github.com/dotnet/spark/issues/1170

GeorgeS2019:
@dbeavon 

Could you put all the necessary merges and upgrade e.g. to .NET8 to your [fork](https://github.com/dbeavon/sparkdotnet)?

I could direct people to your fork.

We state clearly which latest version of Spark this will support.

We work on this fork until the dotnet team responds.

dbeavon:
@GeorgeS2019 

> Could you put all the necessary merges and upgrade e.g. to .NET8 to your [fork](https://github.com/dbeavon/sparkdotnet)?

Yes, but I think we need to know what cloud platforms we want to support.  I think there are some minor tweaks needed for HDInsight.

I'm guessing we should limit ourselves to HDInsight for now (and local OSS on windows).

Then we need to decide if we use a new version number or stick with 2.1.x.
Then we need to decide if we publish a nuget for our Microsoft.Spark changes (to make it easier when new developers start working on a .Net driver) 

The amount of work to be done is non-trivial, maybe about 40 hours to get up and running with .net 8 on HDI.   Most of the work is non-technical and NOT an area where I add value (ie. management and decision making and communicating).    I don't want to host this in my own fork forever.  Have you tried to contact Microsoft?  Do you have a .net foundation membership?

Frankly, I'd prefer if you owned a fork, and I'll send you PR's to merge.  Your help would be invaluable, whether technical or non-technical.    I can send you PR's right away for .net 8 and messagepack and HDI compatibility.

GeorgeS2019:
Start small..we recruit more to join. Local Windows 11 OS is key.

GeorgeS2019:
Updated

dbeavon:
> @arsdragonfly @dbeavon
> 
> With this merge, what version of Spark will this support?
> 

The versions of Spark are probably the same as they were in Microsoft.Spark 2.1.1.

I am only testing Spark 3.1.2.  (Version supported for HDI 5) That has been sitting around for a long time.  



I am pretty sure we could get to other versions of spark if needed.  It seems like the least of our concerns at this point.

What is the concern insofar as Spark versioning?   Please remember that .net and spark are pretty independent of each other.  You can support a wide range of versions without too much difficulty.

GeorgeS2019:
Just start very small..others will join with renewed interest

arsdragonfly:
still discussing internally on how to structure project governance, stay tuned

GrabYourPitchforks:
Note to @AFFogarty and other MSFT reviewers: MessagePack's "typeless" functionality is banned for MSFT internal use under the same regulation that bans BinaryFormatter for MSFT internal use. If you accept this PR, you will still have to jump through the same "ask upper management for an exception to use this" hoops that you would've had to go through with BF.

As OP notes, you could make an argument that since no logical trust boundary is being crossed, it's all a moot point anyway and the exception should be granted. I imagine your security team would sign off on that statement. But the main point I'm trying to make is that this PR does not impact any internal signoff + release process. This PR should be viewed wholly through the lens of broader ecosystem impact, nothing more.

arsdragonfly:
> Note to @AFFogarty and other MSFT reviewers: MessagePack's "typeless" functionality is banned for MSFT internal use under the same regulation that bans BinaryFormatter for MSFT internal use. If you accept this PR, you will still have to jump through the same "ask upper management for an exception to use this" hoops that you would've had to go through with BF.
> 
> As OP notes, you could make an argument that since no logical trust boundary is being crossed, it's all a moot point anyway and the exception should be granted. I imagine your security team would sign off on that statement. But the main point I'm trying to make is that this PR does not simplify any internal signoff + release process. This PR should be viewed wholly through the lens of broader ecosystem impact, nothing more. I am only posting this comment because I am the compliance owner for serialization related technologies and I wanted to clarify how this PR would / would not impact any signoff.

Thanks for the reminder. This is the unfortunate workaround that we'd probably have to settle around since BF is on its way out and [the strictly statically typed alternative that Orleans folks chose to use since Orleans 7 will IMO make .NET for Spark completely useless for anything but the most trivial toy programs.](https://github.com/dotnet/orleans/issues/8312) @luisquintanilla could you help drive the security review/signoff process for this particular PR and for our future community-driven development model in general?

src/csharp/Microsoft.Spark/Broadcast.cs
```diff
@@ -20,7 +20,6 @@ namespace Microsoft.Spark
     /// also attempts to distribute broadcast variables using efficient broadcast algorithms to
     /// reduce communication cost.
     /// </summary>
-    [Serializable]
     public sealed class Broadcast<T> : IJvmObjectReferenceProvider
     {
         [NonSerialized]
```
grazy27, line 25: Let's replace `[NonSerialized]` wtih `[IgnoreMember]` attribute from MessagePack to ensure non-breaking changes

src/csharp/Microsoft.Spark/Broadcast.cs
```diff
@@ -20,7 +20,6 @@ namespace Microsoft.Spark
     /// also attempts to distribute broadcast variables using efficient broadcast algorithms to
     /// reduce communication cost.
     /// </summary>
-    [Serializable]
     public sealed class Broadcast<T> : IJvmObjectReferenceProvider
```
grazy27, line 23: In order to register broadcast var in the registry on it being serialized, we can implement IMessagePackSerializationCallbackReceiver, and rename existing `OnSerialized` to `OnBeforeSerialize()`. Otherwise it's impossible to access var from worker context

src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
```diff
@@ -5,20 +5,24 @@
...
     {
-        [Serializable]
         private class TestClass
         {
             private readonly string _str;
 
+            // TODO: find out why MessagePack is requiring a parameterless constructor.
+            public TestClass()
+            {
+            }
```
grazy27, line 24: There are several more classes that require similar empty ctors, such as ExternalClass or Broadcast

src/csharp/Microsoft.Spark/RDD/Collector.cs
```diff
@@ -66,15 +66,13 @@ internal interface IDeserializer
...
         /// </summary>
         private sealed class BinaryDeserializer : IDeserializer
         {
-            private readonly BinaryFormatter _formater = new BinaryFormatter();
-
             public object Deserialize(Stream stream, int length)
             {
-                return _formater.Deserialize(stream);
+                return MessagePackSerializer.Typeless.Deserialize(stream);
             }
```
grazy27, line 76: There's a bug in MessagePack I found while tried to reuse these changes, when called from Collector::Collect: Deserializer erases entire stream after the first read. 

Workaround for it is to either read stream in small chunks and pass to MessagePack, or load entire stream in MemoryStream and then use it.


https://github.com/MessagePack-CSharp/MessagePack-CSharp/issues/1881

grazy27 (commented):
Thanks for your PR!

src/csharp/Microsoft.Spark.UnitTest/UdfSerDeTests.cs
arsdragonfly: Now that [the standalone nuget package for BF](https://www.nuget.org/packages/System.Runtime.Serialization.Formatters/9.0.0-preview.6.24327.7) is now available, I'm leaning towards scraping the whole PR and moving to that instead, because this is IMO a major limitation that could break any code that uses user/3rd party classes that reference a class without an empty ctor. What do you think?

grazy27: Since microsoft doesn't add [Serializable] attributes to [new types since .NET 6](https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md),  sticking to NuGet will still not work long-term - As soon as there's a new commonly used type -  integration will fall apart.

Imho the best option long-term would be either to investigate MsgPack further (Custom resolver for classes without default ctor and workaround for the bug with empty stream), or, if there's no way to work-around ctor behavior - make serializer configurable and include both options

I created a PR with version bump up for .NET and dependencies, for .NET8 it's still possible to use BinaryFormatter from library, for .NET9 we'll have to migrate to NuGet

bolcman (approved):

GeorgeS2019:
@bolcman who else is/are needed to merge this PR?

GeorgeS2019 (approved):

arsdragonfly:
do not merge this PR (see above mentioned problems)

wudanzy:
This was merged in https://github.com/dotnet/spark/pull/1202, thank you @arsdragonfly and all other involved.

Issue #1170, opened 2024-04-17
[FEATURE REQUEST]: .Net 8.0 support


dbeavon:
**Is your feature request related to a problem? Please describe.**

We need .Net 8 support.  It is the current LTS.

**Describe the solution you'd like**

I'm able to successfully test my solutions on .Net 8.  There is no reason not to upgrade.

**Describe alternatives you've considered**

Prior runtimes are no longer needed.

wudanzy:
This is done by #1197.

Issue #1048, opened 2022-04-11
[BUG]: downloadDriverFile failing on Utils.fetchFile using microsoft-spark-3-2_2.12-2.1.0 on a 10.4 LTS (includes Apache Spark 3.2.1, Scala 2.12)


strtdusty:
**Describe the bug**
I am trying to upgrade my spark-submit job in azure databricks to use the newest dotnet spark version.  I have upgraded the version number in db-init.sh, upgraded my nuget package and updated the spark-submit params.  I see there was a recent change in downloadDriverFile because the signature of fetchFile changed.   I'm guessing I just missed something in the upgrade.

**To Reproduce**

1.  Create an app that uses Microsoft.Spark v 2.1
Setup the db-init.sh to pull the 2.1 version:
`https://github.com/dotnet/spark/releases/download/v2.0.0/Microsoft.Spark.Worker.netcoreapp3.1.linux-x64-2.1.0.tar.gz`
2. Upload the artifacts to dbfs
3.  Create a databricks job using spark-submit params similar to 
`["--class","org.apache.spark.deploy.dotnet.DotnetRunner","dbfs:/FileStore/OO/OO-TestApp/microsoft-spark-3-2_2.12-2.1.0.jar","dbfs:/FileStore/OO/OO-TestApp/SparkTestApp.zip","SparkTestApp"]`
Use a 10.4 LTS  cluster (includes Apache Spark 3.2.1, Scala 2.12)

Standard Error:
`Warning: Ignoring non-Spark config property: libraryDownload.sleepIntervalSeconds
Warning: Ignoring non-Spark config property: libraryDownload.timeoutSeconds
Warning: Ignoring non-Spark config property: eventLog.rolloverIntervalSeconds
Exception in thread "main" java.lang.NoSuchMethodError: org.apache.spark.util.Utils$.fetchFile$default$7()Z
	at org.apache.spark.deploy.dotnet.DotnetRunner$.downloadDriverFile(DotnetRunner.scala:222)
	at org.apache.spark.deploy.dotnet.DotnetRunner$.main(DotnetRunner.scala:77)
	at org.apache.spark.deploy.dotnet.DotnetRunner.main(DotnetRunner.scala)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

Issue #1043, opened 2022-03-14
[BUG]: Spark Worker throwing System.IO.FileNotFoundException exception


DeepikaPrabha:
**Describe the bug**
I am reading a csv file and a adding a column using Udf.   This csv file has around 17M rows. I am using 8 executors (8 cores * 56GB)

```
using System;
Func<Column, Column> GetCanonUrlUdf = Udf<string, string>( url => { 
    if(!String.IsNullOrEmpty(url)) return "Dummy";
    return "";
});


var path = "<CSV file path>"

var df = spark.Read().Option("header", false).Option("inferShchema", true).Option("delimiter", "\t").Csv(path);
//Display(df);
df = df.WithColumn("newCol", UDFfunction(df["_c0"]));
Display(df);
```
When I do display, just after reading CSV, at line no (2), I don't get any error but if I  do display at line no 4, it throws me below error.

```
[2022-03-14T11:35:06.9462243Z] [vm-e3201287] [Error] [TaskRunner] [1] ProcessStream() failed with exception: System.IO.FileNotFoundException: Assembly 'ℛ*2c8e9aaa-736f-41af-87d1-d96f674af627#1-26, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' file not found ''
   at Microsoft.Spark.Utils.UdfSerDe.<>c.b__10_0(TypeData td) in /_/src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs:line 270
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Microsoft.Spark.Utils.UdfSerDe.DeserializeType(TypeData typeData) in /_/src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs:line 261
   at Microsoft.Spark.Utils.UdfSerDe.Deserialize(UdfData udfData) in /_/src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs:line 160
   at Microsoft.Spark.Utils.CommandSerDe.DeserializeUdfs[T](UdfWrapperData data, Int32& nodeIndex, Int32& udfIndex) in /_/src/csharp/Microsoft.Spark/Utils/CommandSerDe.cs:line 333
[...]

Issue #796, opened 2020-12-14
[BUG]: UDFs in .NET Interactive: example fails (on .NET 5 / Jupyter / macOS)


oefe:
**Describe the bug**
The UDFs in .NET Interactive](https://docs.microsoft.com/en-us/dotnet/spark/how-to-guides/dotnet-interactive-udf-issue#define-a-udf-in-net-interactive) fails on macOS Catalina using .NET 5 and Jupyter.

**To Reproduce**

1. Launch Spark using `spark-submit --class org.apache.spark.deploy.dotnet.DotnetRunner --master local microsoft-spark-3-0_2.12-1.0.0.jar debug`
2. Launch Jupyter using `jupyter notebook`
3. Create a new notebook
4. Follow the steps in [UDFs in .NET Interactive](https://docs.microsoft.com/en-us/dotnet/spark/how-to-guides/dotnet-interactive-udf-issue#define-a-udf-in-net-interactive)
5. Executing cell #5 (particularly `df2.Show();`) fails with: 

```
[2020-12-14T13:27:09.8868410Z] [Triton-273] [Error] [JvmBridge] JVM method execution failed: Nonstatic method 'showString' failed for class '142' when called with 3 arguments ([Index=1, Type=Int32, Value=20], [Index=2, Type=Int32, Value=20], [Index=3, Type=Boolean, Value=False], )
[2020-12-14T13:27:09.8874210Z] [Triton-273] [Error] [JvmBridge] org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 28.0 failed 1 times, most recent failure: Lost task 0.0 in stage 28.0 (TID 26, triton-273.fritz.box, executor driver): org.apache.spark.api.python.PythonException: System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Spark.Utils.UdfSerDe.<>c.<DeserializeType>b__10_0(TypeData td) in /_/src/csharp/Microsoft.Spark/Utils/UdfSerDe.cs:line 265
```
(rest of traceback omitted for brevity)

**Expected behavior**

- Should be able to use UDFs
- The example in [UDFs in .NET Interactive](https://docs.microsoft.com/en-us/dotnet/spark/how-to-guides/dotnet-interactive-udf-issue#define-a-udf-in-net-interactive) should work

**Screenshots**
NA

**Desktop (please complete the following information):**
 - OS: macOS 10.15.7
 - .NET 5 (5.0.100)
[...]

PR #1178, opened 2024-07-08
Spark 3.5.x support

grazy27:
## Changes:
- Implemented compatibility with **Spark 3.5.3** (fixes included in a separate commit).  
- Updated project dependencies.  
- Upgraded **.NET 6 → .NET 8** and **.NET 4.6.1 → .NET 4.8**.  
- Fixed several small bugs, including:  
  - Null reference exceptions.  
  - Handling Windows paths with spaces.  
  - Exceptions when after job completion.
  - A few more issues when running locally and on Databricks.  

---

### Spark:
- Spark 3.5.0 on Databricks 14.3: Works, see comment bellow.  
- Spark 3.5.1 on Windows
- Spark 3.5.2 on Windows

### Databricks:  
- **Fails on 15.4**:  
  The following error occurs:  
  ```log
  [Error] [JvmBridge] JVM method execution failed: Static method 'createPythonFunction' failed for class 'org.apache.spark.sql.api.dotnet.SQLUtils' when called with 7 arguments ([Index=1, Type=Byte[], Value=System.Byte[]], [Index=2, Type=Hashtable, Value=Microsoft.Spark.Interop.Internal.Java.Util.Hashtable], [Index=3, Type=ArrayList, Value=Microsoft.Spark.Interop.Internal.Java.Util.ArrayList], [Index=4, Type=String, Value=Microsoft.Spark.Worker], [Index=5, Type=String, Value=2.1.1.0], [Index=6, Type=ArrayList, Value=Microsoft.Spark.Interop.Internal.Java.Util.ArrayList], [Index=7, Type=null, Value=null])
  [2024-09-13T10:47:53.1569404Z] [machine] [Error] [JvmBridge] java.lang.NoSuchMethodError: org.apache.spark.api.python.SimplePythonFunction.<init>(Lscala/collection/Seq;Ljava/util/Map;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Lorg/apache/spark/api/python/PythonAccumulatorV2;)V
  	at org.apache.spark.sql.api.dotnet.SQLUtils$.createPythonFunction(SQLUtils.scala:35)
  	at org.apache.spark.sql.api.dotnet.SQLUtils.createPythonFunction(SQLUtils.scala)
  ```
  
- **Works on 14.3**:  
  Tested on Databricks 14.3, and it works. However, there is a missing functionality for Vector UDFs.  
  Since `UseArrow` is always set to `true` on Databricks, Vector UDFs do not function properly and can crash the entire job. This occurs because Spark splits a single expected `RecordBatch` into a collection of smaller batches, while the code assumes a single batch.  
  Relevant Spark settings: `useArrow, maxRecordsPerBatch`.  

---

## Affected Tickets:
- #1170  
- #1048


grazy27:
@dotnet-policy-service agree

GeorgeS2019:
@grazy27 

Can you share how many of the unit tests pass

grazy27:
> @grazy27
> 
> Can you share how many of the unit tests pass
> 
> The UDF unit tests have not been updated. Are you able to get all of them to pass?
> 
> image

Hello @GeorgeS2019 ,  they do.
image

Saw your issue, probably my env uses UTF8 by default. 
Several tests fail from time to time with `executor driver): java.nio.file.NoSuchFileException: C:\Users\grazy27\AppData\Local\Temp\spark-cc2cf7bc-3c8c-4fdf-a496-266424de943d\userFiles-92d122bb-af9a-40ea-a430-131454afc705\archive.zip `
But they pass if run second time, so I didn't dive deeper

travis-leith:
What is the status of this PR?

grazy27:
> What is the status of this PR?

It works, the tests pass, and performance-wise, it's the best solution I've found for integrating .NET with Spark. The next steps are on Microsoft's side.

I'm also working on implementing [CoGrouped UDFs](https://github.com/apache/spark/blob/30152d0ee938d64cadf8d4acd72115750f424e3b/python/docs/source/getting_started/quickstart_df.ipynb#L876), and I plan to push those updates here as well

GeorgeS2019:
@grazy27 

Can you investigate if your solution work in polyglot .NET notebook interactive?

Previously we all had problem with the UDF.

travis-leith:
> The next steps are on Microsoft's side.

Any idea who is "in charge" of this repo?

grazy27:
> @grazy27
> 
> Can you investigate if your solution work in polyglot .NET notebook interactive?
> 
> Previously we all had problem with the UDF.
> 
> #796

I can take a look, but only if a lonely evening with bad weather rolls around :) No promises, as this isn’t my primary focus.

There are two suggestions from developers that might help. The first is for a [separate code cell](https://learn.microsoft.com/en-us/previous-versions/dotnet/spark/how-to-guides/dotnet-interactive-udf-issue#call-a-udf-on-a-dataframe), and the second is for a [separate environment variable](https://github.com/dotnet/spark/issues/796#issuecomment-744764920). Have you tried both approaches, and does the issue still persist?

wudanzy:
Hi Ihor (@grazy27), thanks for the contribution! I recently get the write permission of this repo and happy to move this forward. Due to limited bandwidth in our team and other priorities, we don't have concrete work items on this project. But we can review your code and let's work together to move this forward!

grazy27:
> Hi Ihor (@grazy27), thanks for the contribution! I recently get the write permission of this repo and happy to move this forward. Due to limited bandwidth in our team and other priorities, we don't have concrete work items on this project. But we can review your code and let's work together to move this forward!

Hello Dan <@wudanzy>,

That's fantastic news—great to hear!

I'd be happy to help with a few more issues to get this project back on track. In my opinion, the most important ones are: 
* Support for UDFs when `UseArrow = true`
* Migrating to a standalone NuGet package for BinarySerializer and upgrading the solution to .NET 9
* Addressing the bug with Databricks 15.4

wudanzy:
Thanks for sharing that!

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
```diff
@@ -839,4 +838,143 @@ private CommandExecutorStat ExecuteDataFrameGroupedMapCommand(
             return stat;
         }
     }
+
+    internal class ArrowOrDataFrameCoGroupedMapCommandExecutor : ArrowOrDataFrameSqlCommandExecutor
```
wudanzy, line 842: Hi Igor, is it possible to split this PR a little bit? This PR is huge, better to only contain 3.5 support and Net 8 support. We can leave other changes to future PRs.

grazy27: Sure, Dan. I'll create a separate PR for CoGrouped UDFs and binary serializer.  Vast majority of other fixes are related to each other though, so PR will still be relatively large.
At one point, I was unsure if this would ever get merged, so I ended up including all the improvements I needed to properly test whether the library meets my requirements in one place, so that if someone wants to build a version it's relatively simple to accomplish.

GeorgeS2019: @wudanzy 

https://github.com/dotnet/spark/pull/1178#issuecomment-2309533060

It will help this project if the support of polyglot notebook is included
https://github.com/dotnet/spark/pull/1178#issuecomment-2309508160

GeorgeS2019: Possible to setup CI/CD so that each PR usability can be tracked?

SparkSnail: > Possible to setup CI/CD so that each PR usability can be tracked? @SparkSnail https://github.com/SparkSnail/spark/actions

Yes, current test pipeline for the repo is broken, we are working to recover pipeline for PRs.

src/csharp/Microsoft.Spark.Worker/Processor/TaskContextProcessor.cs
```diff
@@ -22,29 +22,40 @@ internal TaskContext Process(Stream stream)
...
+            Secret = SerDe.ReadString(stream),
+
+            StageId = SerDe.ReadInt32(stream),
+            PartitionId = SerDe.ReadInt32(stream),
+            AttemptNumber = SerDe.ReadInt32(stream),
+            AttemptId = SerDe.ReadInt64(stream),
+        };
 
-        private static void ReadBarrierInfo(Stream stream)
+        // Needed for 3.3.4+, 3.4.x, 3.5.x
```
SparkSnail, line 44: Could you please double check the version? From the jira: https://issues.apache.org/jira/browse/SPARK-36173, the cpus field is added from spark 3.3.0.

src/csharp/Microsoft.Spark.E2ETest/IpcTests/SparkContextTests.cs
```diff
@@ -64,9 +64,21 @@ public void TestSignaturesV3_1_X()
...
-            sc.AddArchive(archivePath);
 
-            Assert.IsType<string[]>(sc.ListArchives().ToArray());
+            try
+            {
+                sc.AddArchive(archivePath);
+            }
+            catch (Exception)
+            {
+                // It tries to delete non-existent file, but other from that its ok
```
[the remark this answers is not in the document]
grazy27: My logic here is that since nothing related to this API changed inside Dotnet.Spark, and it just calls `AddArchive` on JVM sparkContext, and archive is added successfully - it must be an internal bug in spark itself. I plan to test it with Scala API directly and report later

image

src/csharp/Microsoft.Spark.Worker/Processor/TaskContextProcessor.cs
grazy27: Fixed, thanks for attentiveness

src/csharp/Microsoft.Spark.E2ETest/IpcTests/Sql/CatalogTests.cs
```diff
@@ -88,10 +88,10 @@ public void TestSignaturesV2_4_X()
         }
 
         /// <summary>
-        /// Test signatures for APIs introduced in Spark 3.1.*.
+        /// Test signatures for APIs introduced in Spark 3.1.* - 3.2.*
```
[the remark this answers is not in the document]
grazy27: Removed these changes, there's one method which doesn't support dbName = `null` anymore. I updated comments and test case. 
Thanks for the review!

src/csharp/Microsoft.Spark.E2ETest/IpcTests/SparkContextTests.cs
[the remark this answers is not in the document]
grazy27: Actually it is the exact reason why other tests fail when run together with it: 
image

I restricted this test to (3.1..3.2), versions on which it had already been tested on

src/csharp/Microsoft.Spark.E2ETest/SparkFixture.cs
```diff
@@ -76,7 +76,7 @@ public SparkFixture()
             {
                 // Scala-side driver for .NET emits the following message after it is
                 // launched and ready to accept connections.
-                if (!isSparkReady &&
+                if (!isSparkReady && arguments.Data != null &&
```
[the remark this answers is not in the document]
grazy27: Fixed

src/csharp/Microsoft.Spark.Worker/Command/SqlCommandExecutor.cs
grazy27: Removed unnecessary refactoring and new features from the PR, preserved only .NET, Spark 3.5 and a few fixes

src/csharp/Microsoft.Spark/Sql/Catalog/Catalog.cs
```diff
@@ -248,23 +248,24 @@ public Database GetDatabase(string dbName) =>
             new Database((JvmObjectReference)Reference.Invoke("getDatabase", dbName));
 
         /// <summary>
-        /// Get the function with the specified name. If you are trying to get an in-built
-        /// function then use the unqualified name.
+        ///  Get the function with the specified name. This function can be a temporary function
```
wudanzy, line 251: remove the whitespace here.

src/scala/microsoft-spark-3-5/pom.xml
```diff
@@ -0,0 +1,91 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
wudanzy, line 1: Is this file copied from src/scala/microsoft-spark-3-2/pom.xml? Can you try to move it and then modify? Similar to this: https://github.com/dotnet/spark/commit/80c745bd4099ee4fe28899c85b14ffa2b26e4756

wudanzy: It highlights what is changed.

wudanzy (commented):
Hi @grazy27, I got a basic idea of what is changed, overall, it looks good to me. One thing I found that the scala files content are not changed too much Could you please see if you can move the files instead of adding new ones, which helps highlight what is changed.

src/csharp/Microsoft.Spark.E2ETest/Microsoft.Spark.E2ETest.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
+    <EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>
```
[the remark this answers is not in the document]
grazy27: Yes, we have to, it's a [new requirement for .NET7+](https://learn.microsoft.com/en-us/dotnet/core/compatibility/serialization/7.0/binaryformatter-apis-produce-errors), since binary serialization is now forbidden by default. 

image


For .NET9, we'll have to migrate to the standalone NuGet package, since .NET8 is the last supported version.  

We can arrange a quick call where I can shed some light on background, what do we use serialization for, and alternatives to consider. 

[PR about replacing Serializer](https://github.com/dotnet/spark/pull/1166), inside there are comments on why it doesn't work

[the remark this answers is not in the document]
wudanzy: Got it!

src/scala/microsoft-spark-3-5/pom.xml
grazy27: Sure, it's already done, https://github.com/dotnet/spark/pull/1178/commits/e7eccdfdbc5be41ef66e8507107aba2a3153b2b4

grazy27:
> @grazy27
> 
> Can you investigate if your solution work in polyglot .NET notebook interactive?
> 
> Previously we all had problem with the UDF after making adjustment to migrate to .net6.
> 
> #796
> 
> image https://github.com/Apress/introducing-.net-for-apache-spark/tree/main/ch04/Chapter4

Followed out in https://github.com/dotnet/spark/discussions/1179, and found a related bug: https://github.com/dotnet/spark/issues/1043

src/scala/microsoft-spark-3-5/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+      <artifactId>spark-sql_${scala.binary.version}</artifactId>
+      <version>${spark.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <!-- https://mvnrepository.com/artifact/org.apache.spark/spark-sql -->
+    <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-sql_${scala.binary.version}</artifactId>
+      <version>${spark.version}</version>
+      <scope>provided</scope>
```
wudanzy, line 41: Are those two the same?

wudanzy (dismissed):
LGTM

src/scala/microsoft-spark-3-5/pom.xml
wudanzy: Looks good.

wudanzy:
Hi @grazy27, it looks to me, but we have to wait for the check and another review. @SparkSnail is fixing the broken check and will do another review.

src/scala/microsoft-spark-3-5/pom.xml
grazy27: Slightly different, scala lang version is more specific, it has to match package name
image

grazy27:
> Hi @grazy27, it looks to me, but we have to wait for the check and another review. @SparkSnail is fixing the broken check and will do another review.

Wonderful, thanks @wudanzy. 
I'll create a few more PRs with another improvements after this one is merged

wudanzy:
Hi @grazy27 , good news is that we (thanks @SparkSnail ) have fixed the workflow and now it works for external contributors, we have just upgraded to Spark 3.3 to verify that process. Please resolve the conflicts and we will trigger the test.

wudanzy:
Please note that the test must be triggered by a committer, it is the best outcome after discussing with internal admins. Please feel  free to ping the thread when you want a new test.

wudanzy:
You can reuse this PR, will you use another one?

grazy27:
Oops, thats automatic github action, I'll rebase on the weekend

wudanzy:
Seems that you need to change global.json.

You may verify if it works by `build.cmd -pack -c Release /p:PublishSparkWorker=true /p:SparkWorkerPublishDir=D:\a\path\to\Microsoft.Spark.Worker`

wudanzy:
Are you able to reproduce the above errors locally? I didn't see the error before. Got some links from the web: https://lightrun.com/answers/dotnet-sourcelink-building-a-net-core-31-project-results-in-msb4062
https://github.com/dotnet/sourcelink/issues/386

And it is weird why it is dotnet 9.0 instead of 8.0.

grazy27:
> Are you able to reproduce the above errors locally? I didn't see the error before. Got some links from the web: https://lightrun.com/answers/dotnet-sourcelink-building-a-net-core-31-project-results-in-msb4062 [dotnet/sourcelink#386](https://github.com/dotnet/sourcelink/issues/386)
> 
> And it is weird why it is dotnet 9.0 instead of 8.0.

That's weird, there's no mentions of .NET9 in the PR, I'll have a look tomorrow or in a two days, hope that's OK

wudanzy:
Sure, take your time. Merry Christmas!

grazy27:
@wudanzy thanks you, Merry Christmas!
 
It uses .NET9, [ as it's a latest installed SDK](https://github.com/actions/runner-images/blob/0a970d8e119aebd34846187d0f2c576b2f5f7683/images/windows/Windows2022-Readme.md?plain=1#L498), which is fine - locally I can build .net8 with .net9 sdk without any problems. 

If i uninstall .net9 sdk and install 8- it builds using .net8 without any problems as well. 

The error we can see in the pipeline output is related to some vm configurational issue, as it seems. 
I couldn't reproduce it locally, and can't access pipelines output -  logs from failing stage could help, would you mind sharing them? 

<br><br>


I tried to define SDK version manually:
In order to force a specific version, in global.json we add 
``` json
  "sdk": {
    "version": "8.0.404"
  }
```
But if added, then there's a crash from [arcade project](https://github.com/dotnet/arcade/tree/main). Dotnet.Spark has a very outdated arcade version `1.0`, whereas there's `10` already available.
Updating arcade should help, but from the first glance it affects a lot, changeset is massive and deserves a separate PR. 
Error when specified version manually:
```
C:\Program Files\dotnet\sdk\8.0.404\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(709,5): error MSB4018: The "CreateAppHost" task f
ailed unexpectedly. [W:\Projects\1\dotnet-spark\src\csharp\Microsoft.Spark.Worker\Microsoft.Spark.Worker.csproj::TargetFra
mework=net8.0] [C:\Users\ihor.zhuravlov\.nuget\packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.20230.5\tools\AfterSolutionBuild.proj]
C:\Program Files\dotnet\sdk\8.0.404\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(709,5): error MSB4018: System.MissingMethodExcept
ion: Method not found: 'Boolean Microsoft.NET.HostModel.ResourceUpdater.IsSupportedOS()'. [W:\Projects\1\dotnet-spark\src\
csharp\Microsoft.Spark.Worker\Microsoft.Spark.Worker.csproj::TargetFramework=net8.0] [C:\Users\ihor.zhuravlov\.nuget\packages\microsoft.dotnet
.arcade.sdk\1.0.0-beta.20230.5\tools\AfterSolutionBuild.proj]
```

wudanzy:
Hi @grazy27 , thanks for investigation! This is [the log of a failed build](https://github.com/user-attachments/files/18256210/error_log.txt), and as a comparison, this is [the log of a successful build](https://github.com/user-attachments/files/18256209/normal_log.txt).

wudanzy:
I found that error happened for netstandard2.1 targets, so maybe a quick check is to see if error goes away if remove 2.1 targets. We could then upgrade Arcade version in following PRs.

grazy27:
> I found that error happened for netstandard2.1 targets, so maybe a quick check is to see if error goes away if remove 2.1 targets. We could then upgrade Arcade version in following PRs.

It could help, but if i understand correctly, 2.1 allows us to use newer API's like Span<>, Memory<> and so on, that natively available in 2.1, and results in better performance than through referenced nuget packages.  Let's try one last thing before:

From the log i can see that arcade will download dotnet if it's not available, thus i have an idea - let's try referencing 8.0.403 (On VM, there's only 8.0.404 available). This way it'll use locally downloaded ,net and should work. 
I also fixed error with an empty var `_OfficialBuildIdArgs`

If it doesn't help, I'll try removing 2.1 target

wudanzy:
Sounds good, let's try that, thanks for the change! /AzurePipelines run

grazy27:
It used a correct .net SDK version, but still failed with the same error. 

I added a commit to remove .net standard 2.1, let's see if it helps

wudanzy:
It still fails. [error_log.txt](https://github.com/user-attachments/files/18266265/error_log.txt)

grazy27:
Thanks for pipeline logs!

That's an arcade's impact. Task that fails embeds commit info from git into package info using sourcelink. 
And starting from .net8 SDK, it's forbidden to reference SourceLink packages explicitly, as this info is embedded to the SDK itself.  
https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8/sdk#source-link
https://github.com/dotnet/sourcelink/issues/1142

In newer arcade this is already fixed:
image


Since this behavior is defined in the Arcade nuget package itself, we can't modify it in Dotnet.Spark src. 

I see two options here: 
* Move .NET8 and arcade to a separate PR, keep only spark 3.5.x here
* Update Arcade in this PR

What's your opinion?

grazy27:
@wudanzy done, let's see if it passes e2e tests with 2.4.
 I'll create a separate pr for .net and arcade after this one is merged

wudanzy:
Build has passed

wudanzy:
Failed for 3.5.x tests, spark distributions are not downloaded. 
[error.log](https://github.com/user-attachments/files/18271465/error.log)

wudanzy:
We have to change https://github.com/dotnet/spark/blob/main/azure-pipelines-e2e-tests-template.yml. For 3.5.x, it used hadoop 3 instead of hadoop 2.7.

wudanzy:
Hi @grazy27, I also found that we need winutils for 3.5.x. I will take a look to see how to compile one today.

wudanzy:
Oh, I mistake the above versions for spark versions, it is actually hadoop versions, so we already have that. So it would work if you change https://github.com/dotnet/spark/blob/main/azure-pipelines-e2e-tests-template.yml

grazy27:
On it

grazy27:
@SparkSnail Hello! You may have more background in this: 

I'm troubleshooting failing build pipeline, and can see that only for 3.3.3 there's a hadoop3 bin, but for 3.3.4 there's 2. 

Is it possible to switch to using hadoop 3 for 3.3.3+? 
I tested, and both links work: 
* https://archive.apache.org/dist/spark/spark-3.3.3/spark-3.3.3-bin-hadoop3.tgz
* https://archive.apache.org/dist/spark/spark-3.3.4/spark-3.3.4-bin-hadoop3.tgz

If so, I'd like to try keeping only one if: `if ver >= 3.3.3`, and then we can use powershell with version type comparison, 
image

wudanzy:
@SparkSnail is on vocation, I think we can first have a try.

grazy27:
Wise decision, thanks for the update :)

Committed usage of hadoop3 for 3.3.0+

wudanzy:
The tests passed! Thanks for the investigation and fix! It is too late today, I will review tomorrow.

src/csharp/Microsoft.Spark.Worker/Processor/TaskContextProcessor.cs
```diff
@@ -31,31 +31,32 @@ internal TaskContext Process(Stream stream)
...
+
             StageId = SerDe.ReadInt32(stream),
             PartitionId = SerDe.ReadInt32(stream),
             AttemptNumber = SerDe.ReadInt32(stream),
             AttemptId = SerDe.ReadInt64(stream),
         };
 
+        // Needed for 3.3.0+
+        // https://issues.apache.org/jira/browse/SPARK-36173
         private static TaskContext ReadTaskContext_3_3(Stream stream)
```
wudanzy, line 46: I am wondering if ReadTaskContext_3_3 can rely on ReadTaskContext_2_x.

wudanzy (approved):
LGTM! And let's wait for @SparkSnail for another review.

grazy27:
> LGTM! And let's wait for @SparkSnail for another review.

Thanks, Dan, and happy New Year!

wudanzy:
Thanks, happy new year!

SparkSnail:
> @SparkSnail Hello! You may have more background in this, what's your opinion? Is there any limitations I'm missing, or its worth giving it a try?
> 
> I'm troubleshooting failing build pipeline, and can see that only for 3.3.3 there's a hadoop3 bin, but for 3.3.4 there's 2.
> 
> Reference code: image
> 
> Is it possible to switch to using hadoop 3 for 3.3.0+? I tested, and all these links work:
> 
> * https://archive.apache.org/dist/spark/spark-3.3.0/spark-3.3.0-bin-hadoop3.tgz
> * https://archive.apache.org/dist/spark/spark-3.3.1/spark-3.3.1-bin-hadoop3.tgz
> * https://archive.apache.org/dist/spark/spark-3.3.2/spark-3.3.2-bin-hadoop3.tgz
> * https://archive.apache.org/dist/spark/spark-3.3.3/spark-3.3.3-bin-hadoop3.tgz
> * https://archive.apache.org/dist/spark/spark-3.3.4/spark-3.3.4-bin-hadoop3.tgz
> * https://archive.apache.org/dist/spark/spark-3.5.1/spark-3.5.1-bin-hadoop3.tgz
> 
> If so, I'd like to try keeping only one if: `if ver >= 3.3.0`, and then we can use powershell with version type comparison, like image

The change to use Hadoop3 in Spark 3.3.3 is because they doesn't provide hadoop3 binary downloading in official website, and there is Hadoop2 binary from Spark 3.3.4, hadoop3 is only a workaround for Spark3.3.3 and still use Hadoop2 binary to keep align with previous versions. I'm fine to upgrade to Hadoop3 if Spark 3.5 also need Hadoop3.

SparkSnail (approved):

wudanzy:
Hi @grazy27, I remembered that you preferred rebase and merge, but that was disabled, and I didn't find a way to enable it. 
image

wudanzy:
Merged, thanks for your contribution! @grazy27

src/csharp/Microsoft.Spark.Worker/Processor/TaskContextProcessor.cs
grazy27: Yep, it would be nice to reuse common logic, let's have a look at it after removing support for the obsolete spark versions

GeorgeS2019:
@wudanzy

It is unclear to me if [**UDF** problems with **.NET interactive**](https://github.com/dotnet/spark/issues/1199)  listed here are fully resolved.

PR #1197, opened 2025-01-03
Updated Dotnet and Arcade

grazy27:
## Changes:
- Updated .Net6 -> .Net8
- Updated .Net Framework 4.6.1 -> 4.8
- Updated Nuget dependencies
- Updated Arcade 1.0 -> 8. Used [release 8.0](https://github.com/dotnet/arcade/tree/release/8.0) branch
- Added a few git ignores and fixed a warning in the build pipeline

---

## Affected Tickets:
- #1170  
- #1198


.vscode/settings.json
wudanzy, line 1: Do we need this? What is this for?

LICENSE
wudanzy, line 1: Can we split this into another PR? Since we are not familiar with this, we may need to involve some people from legal department to review. Splitting them would be clearer.

azure-pipelines-pr.yml
```diff
@@ -151,7 +151,6 @@ extends:
         - script: build.cmd -pack
                     -c $(buildConfiguration)
                     -ci
-                    $(_OfficialBuildIdArgs)
```
wudanzy, line 154: Why removing it?

eng/AfterSolutionBuild.targets
wudanzy, line 1: Are all files under eng generated by a script?

.vscode/settings.json
grazy27: We don't, thanks for spotting. I'll add it to git.ignore

LICENSE
grazy27: That's part of Arcade validation pipeline, the same license is used all dotnet MIT projects

[Arcade PR for license validation](https://github.com/dotnet/arcade/commit/4a43fdd015308fd283e4c380fdc201e433fd0088)

azure-pipelines-pr.yml
grazy27: For pr pipeline - we don't define it
image

eng/AfterSolutionBuild.targets
grazy27: No, only the ones in `\eng\common`. 
I divided changes by commits, copypaste and modifications are separate ones

grazy27:
Removed vscode config as a separate commit

Btw, there's .NET9 already available, but it's not LTS, so I still chose .NET8. Does my logic make sense to you?

grazy27:
@wudanzy build timed out, could you please share the logs where exactly did it stall?

wudanzy:
Maybe flaky, retried.

wudanzy:
Hi @grazy27, you have more experience than me on .NET. Do you know if the upgraded .NET 8 (or net48) runtime can run spark jobs compiled by .NET 6 (or net47)? Or do we need an integration test?

wudanzy (approved):
LGTM from my sides, let @SparkSnail to have another look

grazy27:
> Hi @grazy27, you have more experience than me on .NET. Do you know if the upgraded .NET 8 (or net48) runtime can run spark jobs compiled by .NET 6 (or net47)? Or do we need an integration test?

In order to update Dotnet.Spark users will have to pull a new nuget, rebuild their project, and test that it's upgraded successfully. And if spark.worker is used - copypaste dlls to accessible location for it

Overall CLRs of 6 & 8, and 4.6 & 4.8 are compatible,
For framework, users might have to tinker with binding redirects in app.config. 
For .NET8 - no changes needed, the only caveat is that there might be APIs that were deprecated or removed in .net8, so if .net6 lib references these APIs - there will be an error. 

This situation can theoretically occur with spark.worker and UDF dlls, if users decide not to upgrade their project to 8: 
image

SparkSnail (approved):

docs/building/ubuntu-instructions.md
```diff
@@ -110,65 +112,66 @@ Let us now build the Spark .NET Scala extension layer. This is easy to do:
...
 ```
 cd src/scala
-mvn clean package 
+mvn clean package
 ```
 You should see JARs created for the supported Spark versions:
 * `microsoft-spark-2-3/target/microsoft-spark-2-3_2.11-<version>.jar`
 * `microsoft-spark-2-4/target/microsoft-spark-2-4_2.11-<version>.jar`
 * `microsoft-spark-3-0/target/microsoft-spark-3-0_2.12-<version>.jar`
+* `microsoft-spark-3-0/target/microsoft-spark-3-5_2.12-<version>.jar`
```
SparkSnail, line 121: 3-5

grazy27: Yup, I'll fin in the next PR

GeorgeS2019:
@SparkSnail  @grazy27 @wudanzy @radical @shanselman what are next to do list?

Issue #1200, opened 2025-01-17
[BUG]: Unable to build from source due to shutdown of scala-tools.org repository


laneser:
**Describe the bug**

**To Reproduce**

1. Go to src/scala, run mvn clean package
shows

```
[...]

PR #1201, opened 2025-01-17
fix: remove scala-tools.org from pom.xml due to shutdown of scala-tools.org repository

laneser:

ref https://stackoverflow.com/questions/41754260/issues-with-scala-tools-org-and-goodstuff-im

scala-maven-plugin is available on maven central and jcenter.

related issue https://github.com/dotnet/spark/issues/1200

### DIFF
--- a/benchmark/scala/pom.xml
+++ b/benchmark/scala/pom.xml
@@ -12,20 +12,6 @@
     <spark.version>2.4.4</spark.version>
   </properties>
 
-  <pluginRepositories>
-    <pluginRepository>
-      <id>scala</id>
-      <name>Scala Tools</name>
-      <url>http://scala-tools.org/repo-releases/</url>
-      <releases>
-        <enabled>true</enabled>
-      </releases>
-      <snapshots>
-        <enabled>false</enabled>
-      </snapshots>
-    </pluginRepository>
-  </pluginRepositories>
-
   <dependencies>
     <dependency>
       <groupId>org.scala-lang</groupId>
--- a/benchmark/scala/pom.xml
+++ b/benchmark/scala/pom.xml
@@ -7,7 +7,7 @@
   <inceptionYear>2019</inceptionYear>
   <properties>
     <encoding>UTF-8</encoding>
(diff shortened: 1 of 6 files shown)


laneser:
I cannot see the detail build error log, but I tried build again and found mismatch version error,
(but I was build successed before ...)
So fixed the version, and passed at my environment.
Hope this could passed the CICD.

wudanzy:
This is the log. [log.txt](https://github.com/user-attachments/files/18483958/log.txt)

wudanzy:
##[warning].NET 5 has some compatibility issues with older Nuget versions(<=5.7), so if you are using an older Nuget version(and not dotnet cli) to restore, then the dotnet cli commands (e.g. dotnet build) which rely on such restored packages might fail. To mitigate such error, you can either: (1) - Use dotnet cli to restore, (2) - Use Nuget version 5.8 to restore, (3) - Use global.json using an older sdk version(<=3) to build
Info: Azure Pipelines hosted agents have been updated and now contain .Net 5.x SDK/Runtime along with the older .Net Core version which are currently lts. Unless you have locked down a SDK version for your project(s), 5.x SDK might be picked up which might have breaking behavior as compared to previous versions. You can learn more about the breaking changes here: https://docs.microsoft.com/en-us/dotnet/core/tools/ and https://docs.microsoft.com/en-us/dotnet/core/compatibility/ . To learn about more such changes and troubleshoot, refer here: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/dotnet-core-cli?view=azure-devops#troubleshooting
##[error]Dotnet command failed with non-zero exit code on the following projects : [
  'D:\\a\\_work\\1\\s\\src\\csharp\\Microsoft.Spark.Worker.UnitTest\\Microsoft.Spark.Worker.UnitTest.csproj'

wudanzy:
Seems you are only modifying in the Spark 2.4, how about just remove it? Spark 2.4 is too outdated.

wudanzy (approved):

Issue #1199, opened 2025-01-05
[FEATURE REQUEST]: Tracking better support for UDF when using Dotnet Interativ


GeorgeS2019:
https://github.com/dotnet/interactive/issues/3797

When using UDF or User-Defined Functions with Spark.NET, there are problems
[4. User-Defined Functions](https://learning.oreilly.com/library/view/introducing-net-for/9781484269923/html/502681_1_En_4_Chapter.xhtml)

**Describe the solution you'd like**
Now, there is a **new initiative to continue support for Spark.NET**, we need more compatibility with Dotnet **Interativ** and Dotnet **Spark.NET**, especially when working with **UDF**. 

**Describe alternatives you've considered**
Coordinate with SPARK.NET user community to identify ways to make UDF workable using Interativ. 
https://github.com/dotnet/spark/discussions/1179#discussioncomment-11427251

PR #1217, opened 2025-07-21
Fixed UDFs in dotnet.interactive environments

grazy27:
This PR fixes support for User-Defined Functions (UDFs) in .NET Interactive environments, including Jupyter and Polyglot Notebooks.

# Fixed issues:
###  DisposableDirectory API was removed: 
Switched to own simple implementation, all it does - clears the directory before the core shutdown. 

### Running any command failed after the dotnet.interactive core restart 
(If spark itself was not restarted). Fixed by:
     * Adding a random string to the generated metadata filename
     * Using `Spark.GetDefaultSession` as a fallback when `Spark.GetActiveSession` returns null. This implementation is identical to the [SparkSession.Active](https://github.com/apache/spark/blob/d1d29f9dd9ae8a9dcee3147e7276dd4cffdb541d/sql/api/src/main/scala/org/apache/spark/sql/SparkSession.scala#L1028C3-L1028C7), but doesn't throw exceptions when spark is not running
     
###   Current kernel context doesn't contain ResolvedPackageReferences anymore. 
Dotnet.Interactive [will consider](https://github.com/dotnet/spark/discussions/1179#discussioncomment-13422334) exposing missing API.  As of now, the only way to reach all referenced nuget packages is to extract it from dispose queue from the core. This approach is highly-coupled with the internal implementation of dotnet.interactive, and might break in future. We need to migrate to the dedicated API as soon as it's released. Considered alternative options: 
* Listen for PackageAdded, populate collection of referenced packages from there: Doesn't work when dotnet.interactive is not the first package installed
* Get the list of all referenced executables from `csharpkernel .ScriptState.Script.Options.MetadataReferences`, and extract package paths from there. Build probingPaths and NativeProbingPaths manually. This approach doesn't work, because this collection is of a base type, `JustReference`, but File path is only available in a derived class, `MetadataReference`, which is also internal and can be accessed only via reflection. And then we need to implement fragile logic to re-create a package structure dotnet.interactive has already built.
* Also, I tried to use `CommandSucceeded` events instead of `AddMiddleware`. However, I refrained from migrating to this approach as ocasionally it didn't receive events on the successfull command execution, resulting in missing DLL errors.

## Related issues and discussions
* #1179 
* https://github.com/dotnet/interactive/issues/3797 
* #1199

Links:
* [How to run UDF in the notebook](https://learn.microsoft.com/en-us/previous-versions/dotnet/spark/how-to-guides/dotnet-interactive-udf-issue?source=docs#call-a-udf-on-a-dataframe)

### DIFF
--- a/src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive.UnitTest/PackageResolverTests.cs
+++ b/src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive.UnitTest/PackageResolverTests.cs
@@ -60,6 +60,7 @@ namespace Microsoft.Spark.Extensions.DotNet.Interactive.UnitTest
             string actualMetadataPath = actualFiles[1];
             string actualMetadataFilename = Path.GetFileName(actualMetadataPath);
 
+            Assert.Equal(2, actualFiles.Length);
             Assert.Equal(nugetFile.FullName, actualNugetPath);
             Assert.StartsWith(tempDir.Path, actualMetadataPath);
             Assert.Matches("dependencyProviderMetadata_[a-f\\d]{8}00000000001", actualMetadataFilename);
--- a/src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive/AssemblyKernelExtension.cs
+++ b/src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive/AssemblyKernelExtension.cs
@@ -63,7 +63,7 @@ namespace Microsoft.Spark.Extensions.DotNet.Interactive
 
                     if ((context.HandlingKernel is CSharpKernel kernel) &&
                         (command is SubmitCode) &&
-                            TryGetSparkSession(out SparkSession sparkSession) &&
+                        TryGetSparkSession(out SparkSession sparkSession) &&
                         TryEmitAssembly(kernel, tempDir.FullName, out string assemblyPath))
                     {
                         sparkSession.SparkContext.AddFile(assemblyPath);
@@ -128,7 +128,7 @@ namespace Microsoft.Spark.Extensions.DotNet.Interactive
             try
             {
                 sparkSession = SparkSession.GetActiveSession() ?? SparkSession.GetDefaultSession();
-                return true;
+                return sparkSession != null;
             }
             catch (Exception ex) when (ex.InnerException is JvmException)
             {


jonsequitur:
> However, I refrained from migrating to this approach as ocasionally it didn't receive events on the successfull command execution 

Do you have a stable repro for this? This sounds like a bug.

src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive/AssemblyKernelExtension.cs
```diff
@@ -39,22 +39,31 @@ public Task OnLoadAsync(Kernel kernel)
...
+                });
 
                 compositeKernel.AddMiddleware(async (command, context, next) =>
                 {
                     await next(command, context);
 
                     if ((context.HandlingKernel is CSharpKernel kernel) &&
                         (command is SubmitCode) &&
-                        TryGetSparkSession(out SparkSession sparkSession) &&
+                            TryGetSparkSession(out SparkSession sparkSession) &&
```
wudanzy, line 66: Please format this line.

src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive/AssemblyKernelExtension.cs
```diff
@@ -39,22 +39,31 @@ public Task OnLoadAsync(Kernel kernel)
         {
```
wudanzy, line 39: Is it possible that OnLoadAsync is called more than once?

grazy27: No, it's only called once
image

src/csharp/Extensions/Microsoft.Spark.Extensions.DotNet.Interactive.UnitTest/PackageResolverTests.cs
```diff
@@ -54,20 +55,18 @@ public void TestPackageResolver()
                 });
 
             var packageResolver = new PackageResolver(mockSupportNugetWrapper.Object);
-            IEnumerable<string> actualFiles = packageResolver.GetFiles(tempDir.Path);
+            string[] actualFiles = packageResolver.GetFiles(tempDir.Path).ToArray();
```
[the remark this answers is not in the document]
grazy27: Added the check

wudanzy (approved):

PR #1255, opened 2026-09-08
Add Spark 4.0 broadcast support and fix encrypted broadcast handling

SparkSnail:
## Summary

Add `Broadcast<T>` support for Spark 4.0.x and fix encrypted broadcast transport handling shared with existing Spark versions.

## Changes

- Enable Spark 4.0.x broadcast creation through the existing adapter.
- Correct encrypted upload framing to use 32-bit chunk lengths.
- Read each broadcast ID and exactly one MessagePack value without waiting for EOF or consuming the next broadcast's data.
- Validate broadcast IDs, acknowledge successful setup, and close the decryption socket on both success and failure.
- Keep unbuffered socket behavior scoped to encrypted broadcasts; preserve existing serialization and plaintext handling.
- Add unit and E2E coverage for broadcast updates, removal/re-addition, Unpersist, Destroy, malformed input, and recovery after a failed task.
- Extend pipeline coverage with Broadcast tests for Spark 4.0.0–4.0.4, encrypted runs for Spark 4 and 3.5.3, and encrypted Broadcast tests in the Spark 4 no-reuse/simple Worker lanes.

## Validation

- Core unit tests: 206 passed.
- Worker unit tests: 157 passed; one Linux-only SIGTERM test skipped on Windows.
- Local Windows E2E: 74 passed.
- Local Linux/WSL E2E: 74 passed.
- E2E coverage includes Spark 4.0.4 and 3.5.3, plaintext/encrypted broadcasts, and Spark 4 Worker lifecycle modes.
- Build, whitespace, and YAML syntax checks passed.


src/csharp/Microsoft.Spark/Broadcast.cs
```diff
@@ -108,19 +108,21 @@ private string CreateTempFilePath(SparkConf conf)
         /// </summary>
         /// <param name="sc">SparkContext object of type <see cref="SparkContext"/></param>
         /// <param name="value">Broadcast value of type object</param>
+        /// <param name="version">Spark version; defaults to the running Spark environment</param>
         /// <returns>Returns broadcast variable of type <see cref="JvmObjectReference"/></returns>
-        private JvmObjectReference CreateBroadcast(SparkContext sc, T value)
+        internal JvmObjectReference CreateBroadcast(SparkContext sc, T value, Version version = null)
```
wudanzy, line 113: Not familiar with C#, can the default value added in the parameter?

src/csharp/Microsoft.Spark.Worker/Processor/EncryptedBroadcastReader.cs
```diff
@@ -0,0 +1,111 @@
...
+using Microsoft.Spark.Utils;
+
+namespace Microsoft.Spark.Worker.Processor
+{
+    /// <summary>
+    /// Reads interleaved broadcast IDs and MessagePack values from the decryption socket.
+    /// </summary>
+    internal sealed class EncryptedBroadcastReader
+    {
+        private const int ReadChunkSize = 8192;
```
wudanzy, line 18: Shall we make it configurable?

src/csharp/Microsoft.Spark.Worker/Processor/EncryptedBroadcastReader.cs
```diff
@@ -0,0 +1,111 @@
...
+        // The runtime's maximum byte[] length; Array.MaxLength is unavailable on .NET Framework.
+        private const int MaxByteArrayLength = 0x7FFFFFC7;
```
wudanzy, line 20: Shall we make it configurable since it impacts RAM usage.

wudanzy (approved):

src/csharp/Microsoft.Spark/Broadcast.cs
SparkSnail: C# supports optional parameters, but SparkEnvironment.SparkVersion is a runtime property and cannot be used directly as a parameter default. Using Version version = null and resolving it with version ??= SparkEnvironment.SparkVersion inside the method provides the intended behavior: callers can omit the version, and we use the running Spark version.

src/csharp/Microsoft.Spark.Worker/Processor/EncryptedBroadcastReader.cs
SparkSnail: This controls the initial buffer size and the maximum bytes requested per socket read, not the total broadcast size or memory limit. The buffer grows when needed, so larger broadcasts are still supported. Making this configurable would primarily be a performance-tuning option. I propose keeping it as an internal default for this PR and revisiting configurability if profiling demonstrates a need.

src/csharp/Microsoft.Spark.Worker/Processor/EncryptedBroadcastReader.cs
SparkSnail: The memory concern is valid. This constant represents the byte-array size ceiling, not a preallocated buffer or a memory budget, the buffer starts at 8 KiB and grows as needed. However, the reader does buffer a complete serialized value, so large broadcasts can still consume substantial memory below this ceiling.
A configurable lower buffer limit would be a separate fail-fast policy: it would reject oversized broadcasts, rather than reduce the memory needed to deserialize them, and would not cap total Worker memory. I propose retaining the hard ceiling for this protocol-correctness PR and considering configurable resource limits separately, with explicit defaults and oversized-payload behavior.

